Issue #4, opened 2017-05-29
Amazon Aurora - cannot set password when creating virtual server


versantus:
I've been experimenting with using Amazon Aurora as a remote MySQL database for Virtualmin. There seems to be an incompatibility, based around the way Virtualmin creates and updates MySQL users. 

If I set the Aurora details in Webmin > Servers > MySQL Database Server > Config, Virtualmin correctly tries to create the database and users when I create a new Virtual Server.

However, upon creation, and at modification time, I get an error
"set password for 'example'@'hostname.com'=password('greatpassword') failed: Can't find any matching row in the user table"

The server is created, and the MySQL user is added to the database on Aurora, but no password is set.  Attempts to modify the server result in the same error, as do the same efforts from the virtualmin CLI tool ("virtualmin modify-database-pass  --domain mydomain.com --type mysql --pass "greatpassword"")

Trying to run the same SQL ("set password for....") directly on the Aurora DB also fails with the same error, even though the user definitely exists ("select * from user where User='example' and host='hostname.com'" returns the result).

A simple fix is to change the way Virtualmin sets and updates the password in /usr/share/webmin/virtual-server/feature-mysql.pl so that it specifies the user and host params in a different way.

Once this change is made, the user is created on the database, and can be modified as normal.

```
diff --git a/feature-mysql.pl b/feature-mysql.pl
index 33c52dd..cf64c82 100755
--- a/feature-mysql.pl
+++ b/feature-mysql.pl
@@ -2637,7 +2637,7 @@ if (&compare_versions($remote_mysql_version, "5.7.6") >= 0) {
        return ("insert into user (host, user, ssl_type, ssl_cipher, x509_issuer, x509_subject, plugin, authentication_string) values ('$host', '$user', '', '', '', '', 'mysql_native_password', $encpass)");
        }
 elsif (&compare_versions($remote_mysql_version, 5) >= 0) {
-       return ("insert into user (host, user, ssl_type, ssl_cipher, x509_issuer, x509_subject) values ('$host', '$user', '', '', '', '')", "flush privileges", "set password for '$user'\@'$host' = $encpass");
+       return ("insert into user (host, user, ssl_type, ssl_cipher, x509_issuer, x509_subject) values ('$host', '$user', '', '', '', '')", "flush privileges", "update user set password=$encpass where User='$user' and Host='$host'");
        }
 else {
        return ("insert into user (host, user, password) values ('$host', '$user', $encpass)");
[...]

PR #5, opened 2017-05-29
support Aurora weirdness https://github.com/virtualmin/virtualmin-gpl/issues/4

versantus:
Change how MySQL password is set/updated to better support Aurora


jcameron:
As mentioned on https://virtualmin.com/node/52366 , we can't use this patch as it would break other systems on which `set password` is the only reliable way to set the password correctly.

swelljoe:
I guess we'd need this to be a separate code path that detects whether the db is Aurora or MySQL (I think MariaDB also works the same as MySQL).

jcameron:
Note that this applies to any MySQL server that has the `--skip-name-resolve` flag set.

I've committed https://github.com/virtualmin/virtualmin-gpl/commit/fad53fbfb639f58739aeb811b25882788c65fe21 in an attempt to handle this case.

versantus:
Thanks for your work so far on this. I'll try out that version and let you know how I get on.

PR #6, opened 2017-06-11
solved node 42087 MySQL password hashes fail

Malkaduhimi:
The error which is solved is the following:
Failed to migrate virtual server : SQL set password for 'dbuser'@'localhost' = '*B1DFB503E89D3BEDEAB2386570D000510B9E3162&references_priv' failed : Password hash should be a 41-digit hexadecimal number

The reason was, is that the algorithm was selecting everything after password= untill it found another '=', I have changed the regular expression to stop when finding a '&'


jcameron:
Thanks!

PR #32, opened 2017-12-07
Fix Apache CGI test

adbario:
Currently "Re-check configuration" hangs, if both mod_cgi and mod_cgid are not available for Apache, because of typo in error name. Also fixed a typo in the error message.


jcameron:
Thanks!

Issue #31, opened 2017-12-06
Quotas not working (v6.02)


adbario:
Hi, with the latest Webmin and Virtualmin, quotas are not saved anymore for virtual servers when users home directory is outside of the default "/home" on separate disk partition. Virtualmin sets unlimited quotas event though default account plan has 1024 MB quota.

Webmin 1.869
Virtualmin 6.02
CentOS 7.4
Home directory base: /var/www/vhosts

```
$ findmnt -r

/var/www/vhosts /dev/vdb1 ext4 rw,relatime,seclabel,quota,usrquota,grpquota,data=ordered
```

Also bandwidth monitoring doesn't show up on System Information anymore and the alert box "Information! Virtualmin's configuration has not been checked since..." doesn't go away even after re-checking connfiguration.

All these were still working fine with v6.01 few days ago.

Edit: bandwidth came up after a while (clean install), quotas are still missing.

adbario:
Downgrading back to v.6.01 fixes the issues: quotas are working normally again and are set properly, also Virtualmin config re-check alert disappears after running it.

adbario:
Found the issue finally...

There's a check to have either mod_cgi or mod_cgid module available for Apache, but that check ahs an error on the error message name where it's trying to point, and this makes "Re-check configuration" to hang if neither of these modules are available.

Also very strange behaviour elsewhere after this, as the config check doesn't finish properly and seems to skip some things. There was also a typo in the actual error message.

<img width="751" alt="screen shot 2017-12-07 at 2 01 21" src="https://user-images.githubusercontent.com/22136575/33696274-645ffd44-daf9-11e7-8536-321cde1444f4.png">

Fixed here #32 

What's the reason to make Apache to use cgi? Let's say that you're running a server only for websites and no other services, and if there's no use for cgi then it shouldn't be forced.

adbario:
Forcing cgi was added because of issue with email autoconfig scripts #28. I think there should be some kind of check to only require cgi if those email config scripts are in use?

Issue #28, opened 2017-11-15
Email autoconfig doesn't run perl it just prints the perl source code.


chris001:
Steps to reproduce.
1. Email autoconfig is enabled globally for all domains.
2. Ubuntu 16.04 Xenial LTS
3. Visit the Thunderbird autoconfig web url: `http://autoconfig.example.com/mail/config-v1.1.xml?emailaddress=info@example.com`
4. Get redirected to: `http://autoconfig.example.com/cgi-bin/autoconfig.cgi?emailaddress=info@example.com`
5. See printout of the perl source code.  It should run, not print out!
```
#!/usr/bin/perl
# This script gets copied to each domain's CGI directory to output the
# correct SMTP and IMAP server details.

# These variables get replaced when the script is copied
$OWNER = "";
$USER = 'example';
$SMTP_HOST = 'example.com';
$SMTP_PORT = '587';
$SMTP_TYPE = 'STARTTLS';
$SMTP_SSL = 'yes';
$SMTP_ENC = 'password-cleartext';
$IMAP_HOST = 'mail.example.com';
$IMAP_PORT = '993';
$IMAP_TYPE = 'SSL';
$IMAP_SSL = 'yes';
$IMAP_ENC = 'password-cleartext';
$PREFIX = 'example';
$STYLE = '0';

sub error_exit
{
print "Content-type: text/plain\n\n";
print @_,"\n";
exit(0);
}

# Get email address parameter
if ($ENV{'QUERY_STRING'} =~ /emailaddress=([^&]+)/) {
	# Thunderbird style
	$email = $1;
	$email =~ s/%(..)/pack("c",hex($1))/ge;
	($mailbox, $SMTP_DOMAIN) = split(/\@/, $email);
	$mailbox && $SMTP_DOMAIN ||
	    &error_exit("emailaddress parameter is not in user@domain format");
	}
elsif ($ENV{'REQUEST_METHOD'} eq 'POST') {
	# Outlook style
	read(STDIN, $buf, $ENV{'CONTENT_LENGTH'});
	$buf =~ /<EMailAddress>([^@<>]+)@([^<>]+)<\/EMailAddress>/i ||
		&error_exit("EMailAddress missing from input XML");
	($mailbox, $SMTP_DOMAIN) = ($1, $2);
	$email = $1."\@".$2;
	}
else {
	&error_exit("Missing emailaddress parameter");
	}

# Work out the full username
if ($mailbox eq $USER) {
[...]

PR #33, opened 2017-12-07
Require cgi only if mail is enabled

adbario:
Mail client auto-configuration requires active Apache module mod_cgi or mod_cgid. If mail is disabled on the server, cgi requirement shouldn't exist.

Issues #28 and #31

Also skip checking missing FTP shell if FTP in disabled.

### DIFF
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -13925,7 +13925,8 @@ if ($config{'web'}) {
 	    !$apache::httpd_modules{'mod_fcgid'}) {
 		return $text{'tmpl_ephpmode2'};
 		}
-	if ($apache::httpd_modules{'core'} >= 2.4 &&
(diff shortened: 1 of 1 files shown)


jcameron:
But CGI support is also needed for potentially running PHP scripts, webalizer, awstats and other functions.

adbario:
Oh yes you're right, I didn't think of all these options. Anyway would be good to have some way to disable it, ie. I do have one server which only runs WordPress sites with PHP-FPM, no mail, webalizer or awstats, and no need for CGI. Would be great to be able to turn it off without hacking virtual-server-lib-funcs.pl manually after every Virtualmin update.

jcameron:
In that case I'd just recommend enabled the cgi Apache module anyway - it will be cheap if it isn't used.

adbario:
Sure that works, I'll do that or just turn off the check manually. Thanks!

PR #35, opened 2017-12-11
Certificate install error - unwanted replace of space to new line

vmswtje:
- (bugfix) virtualmin api: install-cert api script removes spaces but this causes validation errors because of the required ------BEGIN CERTIFICATE----- syntax of the first and last name

(as far i can see, it isn't possible to install a string-based certificate from cli or api because of this line, because there should be a space in the first and last string)


jcameron:
Thanks!

PR #47, opened 2018-01-26
Proposal for an optimal configuration for FastCGI

herewithme:


swelljoe:
Those variables can't be set that way with environment variables. The environment variables that are available in mod_fcgid are ALL_CAPS. So, this won't do what you think it does. The docs are here: https://httpd.apache.org/mod_fcgid/mod/mod_fcgid.html (each section will list the related environment variables if any)

I do think we could tweak the max requests downward some. I'm not sure why it's set to 9999 (it can be set to unlimited with 0, but that's potentially dangerous for misbehaving scripts...and 9999 is pretty high from that perspective, as well).

herewithme:
You're right, these changes can occur in the configuration of Apache HTTPD.
I complicated my life;

I close :)

PR #63, opened 2018-09-04
Add MariaDB to Wizard Step 4

Francewhoa:
Hi Virtualmin enthusiasts :)

This is my first commit to Virtualmin project on GitHub. All are welcome to let me know if any question or suggestion :)

If applied, this suggested commit will update the description text of the "Post-Installation Wizard" at /virtual-server/wizard.cgi?step=4&xnavigation=1

Find attached mockup. The suggested updates are highlighted in red. Where the slash symbol "/" means "or". In other words, "MariaDB or MySQL". To remain objective they are listed in alphabetical order.

An increasing number of Linux distros have migrated their default database server from MySQL to MariaDB. Such as Debian 9 Stretch. For newcomers to have a more pleasant experience while installing Virtualmin "Post-Installation Wizard", and reduce risk of confusion, I suggest to include MariaDB into the description. Otherwise it's risky that newcomers to Virtualmin may falsely believe that the "Post-Installation Wizard" will install and replace their default MariaDB with MySQL. As you know it won't do that. The description applies to any MySQL compatible/forked database. Such as MariaDB.

mockup---add mariadb to wizard description---2018-09-03---21 17 48---francewhoa


jcameron:
Thanks for the patch!

jcameron:
Pity there's not a general name for the "mysql family" of databases..

rostovtsev:
Yes, it is indeed.

Should I make it displayed explicitly as MariaDB or MySQL across Webmin/Usermin/Virtualmin?

jcameron:
I think it would be messy to use MySQL / MariaDB everywhere.

rostovtsev:
Yes, this is why I offered to have it displayed selectively. Either MariaDB or MySQL.

Francewhoa:
My vote goes to display selectively. Either display MariaDB or MySQL. Based on what is the default database server.

I would be happy to contribute testing and documentation if needed

Another benefit of display selectively is that Virtualmin graphical user interface could scale big. For example, if in the future there is a need to display an additional item(s). Such as [Percona](https://en.wikipedia.org/wiki/Percona_Server_for_MySQL). Or any other MySQL family items. It could be added when needed.

An additional benefit of display selectively is that it would reduce the amount of false bug reports. I mean presently without display selectively or with "MariaDB / MySQL" if the end user know they have MariaDB installed as default, and Virtualmin reads "MySQL" or "MariaDB / MySQL". It is risky the end users might be confuse, or would falsely believe it's a bug. While it is not.

I'm assuming it is more work to display selectively though. I'm guessing it requires logic in the back-end. But the end users would have a more pleasant experience.

Francewhoa:
By the way thanks @rostovtsev, @jcameron, and @swelljoe for all the awesome work you and the Virtualmin community did with the new Authentic Theme. We heard lots of positive feedback. 
At Ubertus we invited many of our clients to [donate](https://github.com/authentic-theme/authentic-theme#donate) to show appreciation and financially contribute to future development.

swelljoe:
It's all @rostovtsev 's hard work! Me and Jamie just answered questions sometimes. Ilia did all the heavy lifting on Authentic, and we are forever grateful. Some day I hope we can sufficiently repay him!

rostovtsev:
Thank you, Joe for your kind words!

I have been working today to let Webmin modules to provide support to have various labels based on the running service version/type.

I will push the commit when it's ready to test. You should like it! :wink:

rostovtsev:
Here is the first attempt to make it work. I don't like it to be in `%modules` hash and thought to have it exported to `module_ambiguous.info` to gather all files on the load and parse them. Practically it would be only one file.

It replaces almost all places to MariaDB in case you have it installed.

Please have a look at the commit and the screenshots.

screenshot from 2018-09-09 19-00-43

screenshot from 2018-09-09 18-57-10

In case you don't like it or think it's inefficient let me know, I will unroll it.

https://github.com/webmin/webmin/commit/0864f12bc13fdcf5fc539c657ebf1bdb877c12e2

jcameron:
So what triggers the replacement for this patch? Is it a user preference, or would it depend on whether MySQL or MariaDB is installed?

rostovtsev:
Yes, it would depend on the [command output](https://github.com/webmin/webmin/commit/0864f12bc13fdcf5fc539c657ebf1bdb877c12e2#diff-386f498865bcbec736fad4fb80306abaR5143
) (whether it's installed or not), and it potentially would work for any module, not only MySQL/MariaDB pair.

I want to factor out `%modules` and put this into the file but there are some obstacles for doing it, like triggering the function not from inside the module (from other [pages](https://github.com/webmin/webmin/commit/0864f12bc13fdcf5fc539c657ebf1bdb877c12e2#diff-386f498865bcbec736fad4fb80306abaR5101)).

Maybe using some global file, in Webmin root directory?

jcameron:
I'm worried this would make loading language strings really slow, because it would mean running a command on each load. Also, it would prevent them from being cached which is something webmin does at startup time.

rostovtsev:
At the moment no files are read and things work nearly at the same speed, I though of that. My patch only uses `%text` and `%rv` hashes, without loading anything. It's just a modifier.

If we put `%modules` logic into the file, like into `/usr/libexec/webmin/modules.quirks` we could also cache it, right?

rostovtsev:
The replacement happens only if [the test 1](https://github.com/webmin/webmin/commit/0864f12bc13fdcf5fc539c657ebf1bdb877c12e2#diff-386f498865bcbec736fad4fb80306abaR5136) is passed, otherwise it never reaches step 2 and 3.

Francewhoa:
Good morning @rostovtsev :) About those [screenshots above](/virtualmin/virtualmin-gpl/pull/63#issuecomment-419725880). Looks good. I like that both the left side menu and the wizard page read "**MariaDB Database Server**". Because it's clear it's a database server. Some new comer users might not be familiar with MariaDB or the other MySQL flavors. In this case they would likely using Virtualmin interface to search for "Database". Then "MariaDB Database Server" would show in the results. From there they would be able to find "MariaDB Database Server" or do their own research and learn about MariaDB.

As for the code & logic, I trust both you and jcameron's recommendation. I'm not a developer. My strengths are documentation, testing, and Agile Product Manager services on demand.

Francewhoa:
@All :) I updated this ticket title to match the present goal. Your suggestion are always welcome.

Francewhoa:
@All :) I am interested to contribute documentation on the website at https://www.virtualmin.com/documentation
About a draft for this new feature with display selectively. I tried but there is no edit button on this [page](https://www.virtualmin.com/documentation) or its sub-pages. Are you interested in documentation volunteer contributions? If yes, how would you like to proceed? And who is the contact person for this? I created a new forum ticket about this at https://www.virtualmin.com/node/58701
I suggest to continue the discussion there about documentation contribution.
I'm referring to a draft documentation such as updating the present documentation, F.A.Q. about MariaDB or MySQL display selectively. Which in turn could be review for approval before publication.

swelljoe:
I'm the contact for any Virtualmin.com website stuff (though Jamie, and Eric, and Ilia all have admin privileges, Drupal, at least the way we use it, is so crotchety and messy that I don't like to inflict it on anyone else). I've add you to the doc admin role. I *think* that'll let you do what you want. Let me know if you're missing privileges you need and I'll sort it out.

swelljoe:
Seems like maybe we should just add a Virtualmin config variable that gets automatically set to whatever database we find when we run the install wizard, right? Instead of detecting the database regularly...it doesn't change, and if it does change, a bunch of stuff has to change, and presumably there would be a config re-check that could re-set the variable.

The idea of checking for a command seems very hacky and error-prone. Like, what if the user has both, but only one is being used for Virtualmin...and we detect the wrong one? This is pretty common, actually; users have mariadb installed from CentOS, but also the MySQL Community package (or commercial one) installed in another location, and they've configured the MySQL module to use it. That would, I think, lead to the wrong label.

The more I think about it, the more uncomfortable I feel about using command detection to choose the label. Being fuzzily wrong and calling both databases "MySQL" is bad, but it's kind of a hard wrong to call MySQL MariaDB, which could absolutely happen if we're just detecting the existence of a mariadb command or path.

rostovtsev:
The way detection works in committed code, is based on actual module configuration, the executable for `mysql command` can only be one value.. It's all about the user module configuration, so we wouldn't end up messing around, even if user has two databases installed, I think. (the command itself is the way we detect database version out of `mysql` across the module).

The current code does lots of automatic replacements (a lot of it) - everywhere on the module itself, all config pages, all cluster modules and a lot of other places. If you give the idea a green light, I will fix it in few other places in Virtualmin, and possibly Cloudmin as well, along with taking out `%modules` to a cached file, under `module/module.quircks`, that would let any module developer to use it, if needed. (I doubt though it would be ever needed - so much work out of damn labels).

Having a variable set is a fine idea but after checking number of needed fixes, I dropped this thought almost instantly.

Francewhoa:
@swelljoe :) About your [comment](https://github.com/virtualmin/virtualmin-gpl/pull/63#issuecomment-419741035) above. This is to confirm it works. I'm able to edit the doc. Thanks.

rostovtsev:
We decided to undo changes above and find better solution.

What about having such labels to be wrapped in `%`, like `%MySQL%`, and later Webmin could upon initial file load process it? We could also move it to the theme level, where themes could in no time replace those conditional blocks.

PR #69, opened 2019-01-16
Set domain envs & run pre- and post scripts.

Harm133:
Added the function call to set domain envs at the start of the request. After these are set the 'make_changes' function is called to run the pre-edit (custom) script(s).
After the routine has run, the 'made_changes' function is called to run the post-edit (custom) script(s).

After this the ENV variables get cleared.

Regarding : https://www.virtualmin.com/node/58810


jcameron:
Thanks for the patch!

Issue #70, opened 2019-01-18
CentOS Linux 7 PHP-FPM support was not detected : No FPM configuration directory found


agemarks:
CentOS Linux 7.5.1804
Webmin 1.894
Virtualmin 6.04
package php-fpm removed, installed rh-php72-php-fpm

but when Re-Check Configuration:
PHP-FPM support was not detected : No FPM configuration directory found
and FPM is not available in Website Options PHP script execution mode.

jcameron:
Thanks for the fix for this issue!

PR #72, opened 2019-01-19
added support for rh-php*-php-fpm packages from centos-sclo-rh

agemarks:
#70 php-fpm packages from centos-sclo-rh can be detected and configured now.


jcameron:
Thanks for the patch!

PR #73, opened 2019-01-24
More updates to env setting position

Harm133:
More swapping around of lines so the env variables get set at the correct time to allow custom scripts to run.
No extra logic has been added.

### DIFF
--- a/newchain.cgi
+++ b/newchain.cgi
@@ -1,4 +1,4 @@
-#!/usr/local/bin/perl
+#!/usr/bin/perl
 # Save CA certificate
 
 require './virtual-server-lib.pl';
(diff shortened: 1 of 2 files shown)


jcameron:
That shouldn't be necessary - when Webmin is installed, it will update all .cgi and .pl scripts to use the correct path.

Harm133:
Without these changes this:

$oldchain = &get_website_ssl_file($d, 'ca');

Wont run properly.

jcameron:
I'm confused ... how does that `$oldchain` line fail exactly?

Harm133:
It fails on my setup at least, it doesn't properly open the file. If I make those changes it works correctly. Same goes for newkey.cgi btw.

PR #78, opened 2019-02-02
Require 'mysqli' PHP module for phpMyAdmin

garas:
PHP7 has removed 'mysql' module, so I can't install/upgrade phpMyAdmin on PHP7.

Use 'mysqli' instead

### DIFF
--- a/scripts/phpmyadmin.pl
+++ b/scripts/phpmyadmin.pl
@@ -62,7 +62,7 @@ return ( 5 );
 
 sub script_phpmyadmin_php_modules
 {
-return ("mysql");
+return ("mysqli");
 }
 
 sub script_phpmyadmin_php_optional_modules


jcameron:
This should already be handled for all scripts by this code : https://github.com/virtualmin/virtualmin-gpl/blob/master/scripts-lib.pl#L961

garas:
I've tried upgrading existing phpMyAdmin, but Virtualmin tried to install `php-mysql`, `php5-mysql`, but it didn't find such packages. After I modified `/etc/webmin/virtual-server/latest-scripts/phpmyadmin.pl` to use `mysqli`, phpMyAdmin was successfully upgraded. 

So it didn't handled to replace mysql to mysqli.


Webmin version 1.900
Virtualmin version 6.05
Operating system  CentOS Linux 6.10
PHP Version  7.2.14-1.el6.remi

jcameron:
Can you post a screenshot of the incorrect PHP module attempting to be installed?

garas:
I've investigated more why Virtualmin gets PHP version as 5. It gets it by [FCGIWrapper](https://github.com/virtualmin/virtualmin-gpl/blob/b5d5fea4cd383f2d10ea41033f02100213842b60/php-lib.pl#L1049) in `https.conf`.

`https.conf` still had links to `fcgi-bin/php5.fcgi`, `fcgi-bin/php7.0.fcgi` etc. I ran _Re-Check Configuration_ and it found new PHP version and updated links for all domains to `php7.2.fcgi`.

After that, installer successfully installed phpMyAdmin.

---

Could Virtualmin automatically detect new PHP versions?

Could Virtualmin cleanup `fcgi-bin/` directories and delete `php5.fcgi, php7.0.fcgi, php7.1.fcgi` for uninstalled PHP versions for all domains?

PR #79, opened 2019-02-08
Update backup_form.cgi

PitWenkin:
Fix broken destination segment in scheduled backups edit. Currenltxy it needs double clikc to actually open


jcameron:
Thanks for the patch!

PR #105, opened 2019-09-27
Fix MariaDB 10.3+ password --secure-auth issues in user creation and user password updates.

MarcherTech:
Per title, we were having a grand old time with this building out our servers on mariadb 10.3.18, so this issue is not limited to mariadb 10.4. In addition, the logic changed for user creation needed extended to password updates.


jcameron:
Thanks for the patch!

PR #158, opened 2020-05-03
Add "none" option to php modes to disable per vhost

ardabeyazoglu:
Sometimes, php should be disabled per vhost because there is a lot of case for static websites or nodejs apps which needs a reverse proxy but no php. This is especially important for nginx servers, because for each vhost it creates an unused fpm pool wasting resources. 

Although the commit adds the php mode option "none" to the list, the virtualserver must read and process it, that's why I sent another patch to virtualmin-nginx.


jcameron:
Thanks for this patch - but is there a reason why "none" isn't added to the return from supported_php_modes?

I do link the idea, but there are quite a few place in Virtualmin where the PHP mode has to be checked.

ardabeyazoglu:
@jcameron That was what I did first and then removed, because i thought this should not be a counted as a supported mode. However still not sure which is better. If you think it is safer and more maintainable, I will change it that way

ardabeyazoglu:
@jcameron You're right, i read the code a bit more and changed the commit to be more generic.

jcameron:
FYI, I was inspired by this change and have implemented "none" mode for PHP for inclusion in the next Virtualmin release.

marclaporte:
So we can close this?

@jcameron In cases like this, can you share commit URLs so we can learn?

Thanks!

jcameron:
The commits are : 

https://github.com/virtualmin/virtualmin-gpl/commit/a39193763fc64369b685ba90d237690d2a7713cd#diff-aa98c4c01f2f6cde518949f907eb960ca25f3f210a756e7729ad4c503bdd8ddd
https://github.com/virtualmin/virtualmin-gpl/commit/0dbef09addcf57ba8398c661472d411ece8e9ae6#diff-aa98c4c01f2f6cde518949f907eb960ca25f3f210a756e7729ad4c503bdd8ddd
https://github.com/virtualmin/virtualmin-gpl/commit/782aff708de0cb90996104a84a2e33d6f1f3265f#diff-aa98c4c01f2f6cde518949f907eb960ca25f3f210a756e7729ad4c503bdd8ddd

I will close this request ...

PR #165, opened 2020-05-20
Add support for s3 compatible servers with aws-cli

theshaun:
Currently, when aws-cli is installed, the s3_endpoint setting is ignored. 

This change allows for the use of s3 compatible servers that require signature version 4 signing (such as Backblaze B2 - https://www.backblaze.com/b2/docs/s3_compatible_api.html)

This change makes the assumption that the endpoint will be https, as there is no provision to define ports or security in settings

### DIFF
--- a/s3-lib.pl
+++ b/s3-lib.pl
@@ -1024,12 +1024,16 @@ return 1;
 # Run the aws command for s3 with some params, and return output
 sub call_aws_cmd
 {
-my ($akey, $params) = @_;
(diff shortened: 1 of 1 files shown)


theshaun:
Hi All, just touching base on this one. Making sure I've done everything correctly.

I'd like to apply this to my production servers but worried a virtualmin upgrade might wipe out the change

jcameron:
Thanks for the patch!

PR #168, opened 2020-05-25
Delete old content styles 1/2

iliarostovtsev:


iliarostovtsev:
As it's private repo, and there is not much change the second part I just committed directly.

https://github.com/virtualmin/virtualmin-pro/commit/26b0027a9b7f69d747bbd0f931a927e91561ced1

PR #197, opened 2020-08-17
Add to display nicely module version and type

iliarostovtsev:
Let's display version on index page and features and plugins nicely.

#### GPL
image image 

#### Pro
image image

#### GPL
image image

#### Pro
image image


virtual-server-lib-funcs.pl
```diff
@@ -18002,6 +18002,26 @@ sub needs_xfs_quota_fix
 return 2;
 }
 
+sub get_module_version_and_type
+{
+my ($list, $gpl) = @_;
+my ($v_major, $v_minor, $v_type) = 
+		$module_info{'version'} =~ /(\d+).(\d+)\.(\w+)/;
```
jcameron, line 18009: Does this syntax where the regexp automatically returns all the matching groups work in all Perl versions?

iliarostovtsev: It must work in Perl 5.10+, perhaps, even the earlier ones too.

jcameron: I always thought that =~ would just return true if it matched, false otherwise

iliarostovtsev: It's all about parenthesis in variables assignment. I wish I could use non destructive modifier `/r`, which would look even nicer but that is Perl 5.14+ only, if I remember correctly.

iliarostovtsev: > I always thought that =~ would just return true if it matched, false otherwise

It returns `$1`, `$2`, and `$3` as of matched group, and assigns them to the variables in parentheses.

PR #224, opened 2020-11-05
Added api script to execute webmin custom commands

websmurf:
This pull request is based this discussion here: https://forum.virtualmin.com/t/execute-custom-commands-through-the-api/107254/2

I'm by no means a Perl programmer, so please look it over thoroughly, but I got pretty far (enough for my needs in any case).
It creates an API integration to execute custom commands and return the output generated there.

Usage: 

```
virtualmin exec-cmd --id "1598540631" --parameter1 "value" --parameter2 "value"
```

The ID refers ofcourse to the command that should be executed.
There parameters will be added to the request to ./run.cgi and are based on the defined custom command.

It can be called over the remote api as well


jcameron:
This looks useful, but the way it runs the command isn't really following the recommended practices. It would be better to call an API in the `custom` module instead..

swelljoe:
Also, I think this should be a `webmin` subcommand in the Custom Commands module instead of for `virtualmin`. (Good idea, though.)

shoulders:
this is 3-4 years old 😄

shoulders:
this commit should be accepted or removed as it is 3-4 years old

jcameron:
It's a reasonable concept, but it shouldn't be implemented in this way (faking a CGI request)

PR #225, opened 2020-11-06
Update de - change of "Virtuelmin"

HonkXL:
2x appearance of Virtuelmin, but the correct Name is "Virtualmin".


iliarostovtsev:
Thanks!

PR #226, opened 2020-11-06
Update de - correcting some strange words

HonkXL:
Correcting bad sentences, incorrect words and fixed some bad spellings.


iliarostovtsev:
Thank you very much!

PR #248, opened 2021-02-01
Fixing the "Add sub-domain DNS records to parent domain?” doesn’t work for aliases issue

searosin:
This pull request is based on this discussion here: https://forum.virtualmin.com/t/add-sub-domain-dns-records-to-parent-domain-doesnt-work-for-aliases/109187

It would be nice to have the possibility to create alias virtual servers with no separate DNS zone file, just adding the necessary DNS records to the parent zone file in case it's a subdomain of the parent.


jcameron:
Thanks for the patch!

PR #249, opened 2021-02-01
Suggest installing aws-cli when webmin fails to upload to S3

fabiomontefuscolo:
As seen at https://www.virtualmin.com/node/25170, aws-cli is needed for buckets outside US. Then, it would be nice to give this hint to the user. However, I am not sure if it is the best place to put this message.

### DIFF
--- a/s3-lib.pl
+++ b/s3-lib.pl
@@ -969,7 +969,7 @@ while(defined($buf = &read_http_connection($h, 1024))) {
 &close_http_connection($h);
 
 if ($line !~ /^HTTP\/1\..\s+(200|30[0-9])(\s+|$)/) {
-	return (0, "Upload failed : $line");
+	return (0, "Upload failed : $line\nTry installing aws-cli");
 	}
 elsif (!$rheader{'etag'}) {
 	return (0, "Response missing etag header : $out");


jcameron:
What's the error message for non-US buckets though? It would be even better to check for that any show a more accurate error message.

s3-lib.pl
```diff
@@ -969,7 +969,7 @@ sub s3_part_upload
 &close_http_connection($h);
 
 if ($line !~ /^HTTP\/1\..\s+(200|30[0-9])(\s+|$)/) {
-	return (0, "Upload failed : $line");
+	return (0, "Upload failed : $line\nTry installing aws-cli");
```
iliarostovtsev, line 972: How will that be printed on the UI?

@jcameron I suggest 

~~~
Try installing `awscli` package using package manager
~~~

.. as there is no such package as `aws-cli` nor such command.

iliarostovtsev (commented):
How will that be printed on the UI?

@jcameron I suggest 

~~~
Try installing `awscli` package using package manager
~~~

.. as there is no such package as `aws-cli` nor such command.

jcameron:
Sure, that could work

PR #250, opened 2021-02-01
Update hardcoded S3 regions list

fabiomontefuscolo:
Add all regions to the S3 script


jcameron:
Thanks for the patch!

PR #256, opened 2021-03-14
Add consistency in features sorting

iliarostovtsev:
I wasn't happy about the way features are displayed across UI, especially, changing set of features globally, pretty much resorted them when displayed under _Edit Virtual Server_.

The following PR fixes this problem and provides control the way it's displayed by ordering and grouping features accordingly, example:

image

image

As you can see features are logically grouped and when features are disabled globally or when bundles changed (LAMP/LEMP), it will keep features at the expected place, which is important for usability.

The same logic for _Create Virtual Server_ is coming in a bit but this commit is already tested, working, independent and ready to be merged.

### DIFF
--- a/edit_domain.cgi
+++ b/edit_domain.cgi
@@ -380,16 +380,7 @@ if (!$d->{'disabled'}) {
 			}
 		}
 	features_sort(\@grid, \@grid_order_initial);
-
-	my @grid_left = @grid;
-	my $grid_tnum = scalar(@grid);
-	my @grid_right = splice(@grid_left, ($grid_tnum / 2) + ($grid_tnum % 2 ? 1 : 0));
-	my $style_force_no_border = 'style="border:0 !important;"';
-	my $style_flex_cnt = 'style="display: flex; align-items: flex-start; justify-content: center;"';
-	my $lgftable = &ui_grid_table(\@grid_left, 1, undef, undef, $style_force_no_border);
-	my $rgftable = &ui_grid_table(\@grid_right, 1, undef, undef, $style_force_no_border);
-	my $ftable = "<div $style_flex_cnt>" . ($lgftable .$rgftable) . "</div>";
-	print &ui_table_row(undef, $ftable, 2);
+	print &ui_table_row(undef, &vui_features_sorted_grid(\@grid), 2);
 	print &ui_hidden_table_end("feature");
 	}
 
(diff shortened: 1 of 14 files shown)


edit_domain.cgi
```diff
@@ -360,23 +363,32 @@ if (!$d->{'disabled'}) {
...
-	$ftable = &ui_grid_table(\@grid, 2, 100,
-			[ "align=left", "align=left" ]);
+	features_sort(\@grid, \@grid_order_initial, 'grid');
+
+	my @grid_left = @grid;
+	my $grid_tnum = scalar(@grid);
+	my @grid_right = splice(@grid_left, ($grid_tnum / 2) + ($grid_tnum % 2 ? 1 : 0));
+	my $style_force_no_border = 'style="border:0 !important;"';
+	my $style_flex_cnt = 'style="display: flex; align-items: flex-start; justify-content: center;"';
+	my $lgftable = &ui_grid_table(\@grid_left, 1, undef, undef, $style_force_no_border);
```
jcameron, line 389: Any reason to use two separate ui_grid_table calls instead of a singe one in two-column mode?

iliarostovtsev: Yes, the problem comes from shuffling order of features, when used as `ui_grid_table(\@data, 2)`. You can see the difference when trying changing global set of features. This `ui_grid_table` sub is not meant to keep the order, which we need.

iliarostovtsev:
Added for form domain - 

image

edit_domain.cgi
jcameron: Yeah, ui_grid_table will order items like :
```
1 2
3 4
5 6
```
etc..

iliarostovtsev: Yes, this is not we want. We need to keep order. I added 15 more commits below, and what I'll do, I will abstract this logic into a separate sub, rather than copy/pasting the same logic across multiple files (around 6-7).

Finishing final tests.

iliarostovtsev:
Okay, Jamie. Just made last commit and finished. All is polished, checked and _**ready**_!

You can have a final look.

For the future releases I will try to combine separate blocks, like [this one](https://github.com/virtualmin/virtualmin-gpl/pull/256/commits/9e2a1e25a5e58a2613b3565855a64a61b66f526d), into one, and sort features and plugins together (to achieve combined sorting, as we already have in grid list, where features and plugins are sorted together).

iliarostovtsev:
Are there any issues that you'd want me to fix, Jamie?

PR #257, opened 2021-03-14
Add better error style and look when CA is wrong

iliarostovtsev:
Jamie, I have done few more tests and it seems that this new patch doesn't look good enough (UI wise). I have done few changes to make it look nicer.

**New:**

image

**Old:**

image

### DIFF
--- a/cert_form.cgi
+++ b/cert_form.cgi
@@ -65,41 +65,33 @@ if (&domain_has_ssl_cert($d)) {
 	$info = &cert_info($d);
 	$chain = &get_website_ssl_file($d, 'ca');
 	
-	# Warn if the CA is wrong
-	my $check_ssl_ca = sub {
-	if ($check_ssl_ca_done) {
-		return;
-		}
-	if ($chain) {
-		my $cainfo = &cert_file_info($chain, $d);
-		if ($cainfo &&
-		    ($cainfo->{'o'} ne $info->{'issuer_o'} ||
-		     $cainfo->{'cn'} ne $info->{'issuer_cn'})) {
-			print &ui_table_row('',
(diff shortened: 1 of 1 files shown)


cert_form.cgi
```diff
@@ -64,30 +64,43 @@ if (&domain_has_ssl_cert($d)) {
...
+		if (ref($v)) {
+			print &ui_table_row($text{'cert_'.$i},
+				&ui_links_row($v), 3);
+			if ($i eq 'type') {
+				&$check_ssl_ca();
+				}
+			}
+		elsif ($v) {
+			print &ui_table_row($text{'cert_'.$i}, $v);
+			if ($i eq 'type') {
```
jcameron, line 101: Instead of creating that anonymous sub, why not just move all the logic to below this if statement, like :

```
if ($v && $i eq "type") {
  if ($chain) {
    my $cainfo = &cert_file_info($chain, $d);
    etc..
```

iliarostovtsev: Well, yes we could, but then it would be printed below **Other domain names** row, which makes much little sense, and having it printed right under **Signed by CA**, puts the message just in the right place. The downside is more code.

jcameron: No I mean it would still be printed in the same place in the UI, just the code would be moved down (but kept inside the loop).

iliarostovtsev: Oh, okay. Got it. There - https://github.com/virtualmin/virtualmin-gpl/pull/257/commits/fc7a96ec2433cd1f077fc2b7cefbb3c2cdc86c5d

jcameron:
Thanks for the patch!

PR #259, opened 2021-03-15
Fixed translation of Let's Encrypt

HonkXL:
Let's Encrypt was translated in de.auto -> is a name that should not be translated.


iliarostovtsev:
Great, thanks.

PR #273, opened 2021-03-27
To make Letsencrypt http-01 challenge work with proxy, http requests …

derjohn:
…to acme challenge URL must be excluded from being forwarded to the proxy backend.


jcameron:
Thanks!

Issue #109, opened 2019-10-09
PostgreSQL offered in 'Features and Plugins' even if not installed/configured


PitWenkin:
I've never installed PostgreSQL, however `PostgreSQL database` is listed among `Features and Plugins`.
![PostgreSQL-1](https://user-images.githubusercontent.com/45786226/66490513-93207680-eab1-11e9-8bd3-c52d8c5df903.png)

The whole `PostgreSQL Database Server`-module is unused, and shows the following:
![PostgreSQL-2](https://user-images.githubusercontent.com/45786226/66490522-961b6700-eab1-11e9-83fd-9569e216ad82.png)

I checked if some PostgreSQL-packages where installed but the only package matching 'postgre' is postgrey from Postfix 

Operating system 	Ubuntu Linux 18.04.3
Webmin version 	1.930
Usermin version 	1.780
Virtualmin version 	6.07
Authentic theme version 	19.39-2 

https://github.com/virtualmin/virtualmin-gpl/blob/1a70293487dba271845f6782c91f03ad31291738/edit_newfeatures.cgi#L20

PitWenkin:
Still the case on a newly installed system:

Webmin version | 1.973
Usermin version | 1.823
Virtualmin version | 6.16

iliarostovtsev:
Hi,

Do you expect it to be hidden on _Features and Plugins_ page if _PostgreSQL_ package isn't installed?

iliarostovtsev:
What would you expect? Would you like having features that are unused being hidden? @jcameron Jamie, what do you think of hiding core features that cannot be used, like in case of PostgreSQL, when system package is not installed? Do you mind hiding them?

PitWenkin:
It's a bit strange to see features listed I cannot use.

If at least they were somehow marked as not installed …?

iliarostovtsev:
> If at least they were somehow marked as not installed …?

Yes, I think we could do it, unless Jamie thinks that hiding it completely would be a better option.

swelljoe:
This shouldn't be a theme-specific implementation. Better if it goes into Virtualmin proper, if we're doing it. But, I also am not sure if this is the right thing. Features and Plugins is a root-user page, and it's the source of truth on what features are available in the UI elsewhere. It's the one place where I don't think we should be trying to hide everything.

[...]

PR #276, opened 2021-04-06
Add a way to remove un-used database(s) from global feature's list #109

iliarostovtsev:
This PR addresses this https://github.com/virtualmin/virtualmin-gpl/issues/109#issuecomment-813775214

### DIFF
--- a/edit_newfeatures.cgi
+++ b/edit_newfeatures.cgi
@@ -18,7 +18,19 @@ print "$text{'features_desc'}<p>\n";
 # Add rows for core features
 @table_order_initial = ( );
 $n = 0;
-foreach $f (@features) {
+my @features_local = @features;
+
+# Opt-out certain features if not
+# enabled in Webmin Users module
+my %features_exclude_check = ('mysql' => 'mysql',
+                              'postgres' => 'postgresql');
+foreach my $mod (keys %features_exclude_check) {
+	if (!&foreign_available($features_exclude_check{$mod})) {
+		@features_local = grep {$_ ne $mod} @features_local;
+		}
+	}
+
+foreach my $f (@features_local) {
 	local @acts;
 	push(@acts, ui_link("search.cgi?field=$f&what=1",
 		                $text{'features_used'}));


jcameron:
I don't think this is a good idea, because if Postfix is installed but Webmin is looking in the wrong place for it, the user won't even know.

iliarostovtsev:
Postfix? Did you mean to say PostgreSQL?

By default PostgreSQL is enabled in _acls_ so all will be seen and displayed as before, just as that admin capable user would have an opportunity to disable it, and only then it won't be displayed in Features and Plugins page. It's quite rational to me and provides the only way to control the display of core Virtualmin features.

iliarostovtsev:
.. and besides, if Webmin is looking at the wrong place it's a bug or use of unsupported system. Nothing would work smoothly anyway in this case scenario.

jcameron:
OK I see now .. the idea is good, but I prefer a different implementation which I've just checked in.

iliarostovtsev:
Cool. Thanks. I'll give it a try tomorrow.

iliarostovtsev:
> OK I see now .. the idea is good, but I prefer a different implementation which I've just checked in.

It works fine! I like the implementation more. Thank you.

I have a question though -- what else but post-installation wizard changes `$config{'mysql'}` and `$config{'postgres'}` config options?

I mean this check:
https://github.com/virtualmin/virtualmin-gpl/blob/0be3423407facbf97d39348a5d162b65c16ad807/edit_newfeatures.cgi#L23-L25

.. is config `mysql={1|0}` and `postgres={1|0}` is only set from post-installation wizard?

jcameron:
Yes only the post-install wizard and the features and plugins page.

iliarostovtsev:
So what would be the way to re-enable a feature that is no longer present on Features and Plugins page (was disabled)? Re-run post-installation wizard?

jcameron:
Yes that would be the simplest way. There's a link on the left menu to re-run the wizard.

PR #317, opened 2021-10-19
Track Let's Encrypt renewed certificates

tr3pan:
When certbot is installed only one Let's Encrypt account is used and all renewed/issued certificates are tied to it. Let's Encrypt has [rate limits](https://letsencrypt.org/docs/rate-limits/) of `a maximum of 300 New Orders per account per 3 hours`.  `apply_letsencrypt_cert_renewal` function is not aware of that limit and the log is full with `too many new orders recently`. The worst part is the renewal is triggered by `collectinfo.pl` which by default runs every 5 minutes.

The fix is not very smart and potentially might block the collectinfo.pl. Suggestions for improvement are welcome.


jcameron:
I'm curious how this limit is being hit on your system. Do you have a very large number of domains with Let's Encrypt certs that are renewing at the same time?

PR #322, opened 2021-10-22
Update s3-lib.pl

acosonic:
Reading config from config values, since tmp file had akey listed twice and shellouting wasn't working


iliajie:
I will close this, as it was fixed a while ago.

Thanks!

PR #334, opened 2021-11-29
BUG FIX : create-route CLI bug Redirect URL is impossible to set

jackharrispeninsulainteractive:
Hi guys!

Thanks for developing virtualmin, its such a great product.

We have discovered a bug in the create-route cli call for the remote api that means it currently impossible to create a redirect to external url. The issue that the dir variable seems to be getting set prior regardless of is a alias is provided in the call and as such throws out the dir check.

As a solution to this we have renamed dir to webDir and now everything works as intended!

Hopefully this temp fix might help you further troubleshoot this issue further.

Hopefully we can get this fixed implemented in the main branch!

Kind Regards,
Jack Harris

### DIFF
--- a/create-redirect.pl
+++ b/create-redirect.pl
@@ -59,7 +59,7 @@ while(@ARGV > 0) {
 		$url = shift(@ARGV);
 		}
 	elsif ($a eq "--alias") {
-		$dir = shift(@ARGV);
+		$webdir = shift(@ARGV);
 		}
 	elsif ($a eq "--regexp") {
 		$regexp = 1;
@@ -93,8 +93,8 @@ if ($url) {
 		&usage("The --redirect flag must be followed by a URL or ".
 		       "a URL path");
 	}
-elsif ($dir) {
-	$dir =~ /^\/\S+$/ && -d $dir ||
+elsif ($webdir) {
+	$webdir =~ /^\/\S+$/ && -d $webdir ||
 		&usage("The --alias flag must be followed by a directory");
 	}
 else {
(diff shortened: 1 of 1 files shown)


jcameron:
What args are you calling the create-redirect command with when it fails for you before this patch? It seems to be able to create redirects just fine to me ...

jackharrispeninsulainteractive:
Hi Jcameron, Here is the small PHP function we have been using


    public function createRedirect(string $domain, string $path, string $redirect, bool $http = false,bool $https = false ,bool $regexp = false): string
    {
        $program = "create-redirect";
        $params = "&domain=".$domain;
        $params .= "&path=".$path;
        $params .= "&redirect=".$redirect;
        if($http) {
            $params .= "&http";
        }
        if($https){
            $params .= "&https";
        }
        if($regexp) {
            $params .= "&regexp";
        }
        return $this->build($program,$params);
    }

    private function build(String $program = "", String $params = ""): String
    {
        $query = WEBMIN_URL;
        $query .= "?program=";
        $query .= $program;
        $query .= $params;
        $query .= "&json=1";

        $ch = curl_init($query);
        curl_setopt($ch, CURLOPT_USERPWD, WEBMIN_USERNAME.":".WEBMIN_PASSWORD);
        curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
        curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, false);
        curl_setopt($ch, CURLOPT_HEADER, false);

        $result = curl_exec($ch);

        curl_close($ch);

        return $result;
    }

jackharrispeninsulainteractive:
Our issue was with redirecting to external urls not internal folders

jcameron:
I did some testing, but wasn't able to re-produce this problem.  Which Virtualmin version are you running there?

PR #344, opened 2021-12-18
Add support for updating FPM config on domain rename

iliarostovtsev:
Hi, Jamie.

Check this commit out. It adds support for changing PHP directives on domain's home directory rename.

Ref.:
https://forum.virtualmin.com/t/session-problems-after-domain-name-change/113348


iliarostovtsev:
Perhaps, **`restore_web`** needs to call it too?

jcameron:
Thanks for the patch!

iliarostovtsev:
Should we merge it?

iliarostovtsev:
Also, I think **`restore_web`** needs this new function call as well, right, Jamie?

PR #345, opened 2021-12-26
Add support for random usernames upon domain creation

iliarostovtsev:
This PR adds support for creating usernames (altering prefix) based on the new config option called _"Regular expression custom pattern"_ and newly added `substitute_pattern` Webmin API.

It will now be possible to define custom pattern for new domains, like `u-[0-9]{4}`, which will be used upon domain creation as a username, group, home directory, database user and name and etc.

It was tested with other features, like domain rename and cloning.

This patch does minimum changes as possible to add support for this new functionality.

Ref. Issue:
https://github.com/virtualmin/authentic-theme/issues/1572


jcameron (commented):
Would or should this also apply to mailbox names?

iliarostovtsev:
> Would or should this also apply to mailbox names?

Yes, it also affects default administration email address.

Or you mean when creating a user (i.e. using **`edit_user.cgi?new=`**)?

iliarostovtsev:
We could easily added it to Create User and Create Mail Alias pages when this option is enabled. This wouldn't be hard. The hard part is this PR.

iliarostovtsev:
Does it work for you?

virtual-server-lib-funcs.pl
```diff
@@ -11339,26 +11384,32 @@ sub compute_prefix
...
-		for(my $i=0; $i<@p; $i++) {
-			local $testp = join("-", @p[0..$i]);
-			local $pclash = &get_domain_by("prefix", $testp);
-			if (!$pclash) {
-				$prefix = $testp;
-				last;
-				}
+		# Extract prefix based on the user pattern
+		if ($config_longname && $config_longname !~ /^[0-9]$/) {
+			$prefix = &generate_random_available_user($config_longname);
```
jcameron, line 11389: I don't think domain prefixes should ever be random like this - they need to be constant and predictable.

virtual-server-lib-funcs.pl
```diff
@@ -7328,20 +7328,55 @@ sub validate_database_name
...
 }
 
+# generate_random_available_user(pattern)
+# Generates free user available to be used as
+# Unix user, group, and virtual server user
+sub generate_random_available_user
+{
+my ($pattern) = @_;
+$pattern ||= 'u-[0-9]{4}';
+my $user = $main::generated_random_available_user;
```
jcameron, line 7338: What is this global variable used for?

iliarostovtsev: As it is random it must be created only once to match across.

virtual-server-lib-funcs.pl
iliarostovtsev: That's what the initial request was about.

They are predictable in the way admin configures it.

iliarostovtsev: It can be configure to be:

`u-[0-9]{2}` - would produce `u-11`

or

`[a-z]{1}_d-[0-9]{1}` - would produce `x_d-2`

The whole point was that an admin can define the pattern for the username.

jcameron: Wasn't the original request for random domain owner usernames? The prefix is only used for mailboxes..

virtual-server-lib-funcs.pl
jcameron: Does this track the last randomly generated user, to prevent clashes?

iliarostovtsev: > Does this track the last randomly generated user, to prevent clashes?


The random name cannot be assigned [unless it is free](https://github.com/virtualmin/virtualmin-gpl/pull/345/files#diff-a8b0a88367118c8f13666eed07287b13ecc217c8fdfaf6a34d51b226dc196f92R7344-R7346) in all 3 cases - a Unix user, group and Virtualmin prefix.

virtual-server-lib-funcs.pl
iliarostovtsev: > Wasn't the original request for random domain owner usernames? The prefix is only used for mailboxes..

@tpnsolutions mentioned:

> I think it'd be cool if either in the theme or Virtualmin there could be a "username" generator. Similar to how there's a setting in Virtualmin Configuration, it would be neat if this feature allowed you to define a "pattern", then generated a random username based on this.

> The prefix is only used for mailboxes

In Virtualmin we want those to be in sync and we have those synced in most of the times, correct? With default options, if I create a domain the username across (Unix username, db username, admin mailbox name, and etc) are in sync. If I set a _custom username_ upon domain creation it is being used familiar way. My patch does exactly the same.

tpnsolutions: Hey, it was less about having a prefix and more about following a pattern.

In my original post on the topic I mentioned I was using a username like "u1234" where by "u" was at the beginning of each username and the 4-digits that followed were random.

I later clarified when we were discussing the best way to implement that a regex pattern would make sense.

Hence I'd be able to define my pattern to work how I want it (above) but others could define their own pattern to work their way.

Thanks for the effort as always!

jcameron: @tpnsolutions if the domain owner is named like u1234 , do you also want mailboxes in the domain to be named like bob-u1234 ?

virtual-server-lib-funcs.pl
jcameron: Wouldn't it be better to pass this in as a parameter, rather than using a global variable?

virtual-server-lib-funcs.pl
tpnsolutions: no no no, I'd only want the initial domain account to make use of this format... Meaning when I create a "Top-Level Virtual Server" I'd like the custom username format... Mailboxes and FTP/SFTP users should continue using the format of "username@domain.com"

tpnsolutions: Right now, when I create a "Virtual Server" via the GUI, I enter a custom username manually... The idea here is to suggest a username just like a random password is suggested when enabled.

tpnsolutions: Frankly, when I add a domain via the CLI, I have my script generate a random username and password, then feed it to the CLI... If the username is already in use, the CLI simply gives me an error, then I re-run and 99% of the time the next username hasn't been used...

tpnsolutions: Anything smarter then what my script is already doing is just a bonus, but the primary thing was introducing the functionality to the GUI so I don't always have to create via CLI, and so I can setup customers (who I manage the server for) up in a maner where they maintain my username format when they use the panel.

virtual-server-lib-funcs.pl
iliarostovtsev: Yes, it would. But doing it seemed to me way less invasive task. As it had to be called not only upon domain creation but also cloning and renaming (with automatic option) and probably somewhere else. In my patch it does it in place of where the program decides which username/group will be returned. As calling this sub would produce a different result each time, the only straight forward solution was calling it only once and storing a result as a global variable.

iliarostovtsev: Where are we at with this PR?

jcameron: Can we make the randomly generated part of the pattern deterministic within the same process? So if `substitute_pattern` is called multiple times in a row with the same input, it will return the same results.

iliarostovtsev: > Can we make the randomly generated part of the pattern deterministic within the same process? So if substitute_pattern is called multiple times in a row with the same input, it will return the same results.

Well, this is exactly what it's doing with this PR and this is the reason why I used a global variable to store the same result at the first place.

iliarostovtsev: @jcameron Jamie, look, if you don't like randomized user, group mailbox and etc. names as an option, we can forget about this PR. Not a problem at all!

Alternatively, I could create a link (button), next to username fields, here and there, to have an ability to generate a username based on the pattern using a button, familiar way we do for generating a password.

jcameron: I do like the idea in general, I'm just not in love with using a global variable to pass the username around ... it's just not good coding practice.  Also, what happens if `generate_random_available_user` uses the same name for the user and group, but the group name is already in use?

iliarostovtsev: I do not like using global variables either. I just don't know how to do it otherwise in this case.

> Also, what happens if `generate_random_available_user` uses the same name for the user and group, but the group name is already in use?

The sub will be running until all three are free (user, group and prefix).

iliarostovtsev: > I do like the idea in general, I'm just not in love with using a global variable to pass the username around ... it's just not good coding practice.

Actually I know how to fix it without using a global variable. There, have a look at this patch please - https://github.com/virtualmin/virtualmin-gpl/pull/345/commits/0049aabb7688846be9769f95243e6facc1c6592e

jcameron: Isn't that `state $user` variable basically the same as a global variable? 

I think it might be simpler to just submit the current change, and I will take a look to see how the global can be avoided?

iliarostovtsev: > Isn't that state $user variable basically the same as a global variable?

No. `state` always declares a lexically scoped variable, like `my`, it's just that with `state` declaration a variable not getting re-initialized.

> I think it might be simpler to just submit the current change, and I will take a look to see how the global can be avoided?

I already did avoid using global. 🙂

jcameron: Oh I see it's like a "static" variable in Java. Ok, that seems better ..

iliarostovtsev:
Okay, thanks for merging it.

Jamie, there is one tiny issue that I am not sure how to fix.

When this new option is enabled, and you go to create a new virtual-server, and for **Administration username** you select **Custom username** instead of default **Automatic**, then the new name is actually used for everything but a new databased name:

image

Although, the database username is correct and set to custom:

image


The same happens if you enable database feature later. It uses randomized database name, while everything else seems to work.

Can you have a look? I am not sure how to fix that.

jcameron:
Do you mean that the database login is still a random name?

iliarostovtsev:
> Do you mean that the database login is still a random name?

Database login -- is okay!

Default database name -- is random (check screenshot above).

jcameron:
Oh that's because the default database name comes from the domain prefix.

iliarostovtsev:
> Oh that's because the default database name comes from the domain prefix.

Yeah, but how to properly avoid this when **`$in{'vuser'}`** is set?

jcameron:
It's not simple with the current code. Would it make more sense if the default DB name followed the domain name?

Actually, maybe the prefix shouldn't follow the randomly generated username?

iliarostovtsev:
> Actually, maybe the prefix shouldn't follow the randomly generated username?

The point was to keep things in sync when the random option is enabled. Why prefix shouldn't be, it would make things inconsistent then, when this option is enabled?

Default database name is the only disobedient element here and only when **`$in{'vuser'}`** is set:

image

tpnsolutions:
Wow, you guys have been on fire with this feature :-) Most appreciated the attention it's been getting lately.

@iliarostovtsev as we discussed a while back, a simple "regex" expression that then generates a random password based on it is all that's needed.

Frankly, I'd have been satisfied if like the "Admin Password" field it just generated something in that "custom username" field that was based on the "regex" and worse case, if during creation it was discovered this username was taken already, it simply errored out like it does in the CLI when you try to specify an existing domain or username for instance.

Upon error in this manner, the theory would be you'd go back a page, and it'd generate (or you'd click a link to generate) a new username which hopefully wasn't taken this time.

Given it'd be populating the "custom username" field, I figured again in theory this would reduce the overhead of development overall as when you specify a "custom username" already, it gets used everywhere else that it's been configured to be used.

Just a thought :-)

Rock on, but K.I.S.S.

iliarostovtsev:
I could add an option to display a link (button) next to username field to generate a free to use username based on previously configured pattern.

This would have to be optional though, as I have a hard time understanding why would anyone wanted to have a detached user name and a domain name? When we call a domain `mydomain.com`, by default we get usernames and prefixes with the same name, which is easily recognisable. Having a username and a prefix called like `u-101` makes it somewhat confusing. But if you or someone else need it, it's no problem to add it, as it's still would be buried under configurable options and wouldn't standout by default. This is certainly a niche feature.

But right now, to make things perfectly right I need to figure out with Jamie how to make sure that default database name is not randomized when custom username is supplied, and this new option is enabled.

tpnsolutions:
@iliarostovtsev, I hear you.

There were a few case scenarios I was addressing in my move to a "u1234" (no dash) approach.

1. Username length - some services truncate usernames meaning if you use a "reallyreallyreallyreallyreallylong.com" it would be shortened by that service, meaning some would resemble say "reallyreallyreallyreallyreallylong" and others might be more like "reallyreallyreally" making things inconsistent.

2. Shorter is easier - IMHO shorter usernames when it comes to say "htop", "glances" and other monitoring tools is far easier to read then either really long names or truncated ones. Again this is personal experience and therefore preference.

3. Simple database names - since when I setup a server I use the "less is best" approach, I disable the creation of a database by default, and therefore when I do create databases as needed they look like "u1234_wordpress" and another client might have "u3456_wordpress" -- much easier if you ask me.

4. Similar domains - if you own "mydomain.com" and "mydomain.net" then one of them cannot use the same "mydomain" format but rather needs to append more to it, or otherwise be different. Now one might assume both are going to be pointed to the same place and therefore one will become an alias of the other... But that's not always the case.

While I do understand agree most like the idea of "simple", I've taken over 20 years of web hosting and rationalized my decisions including why I give my servers a name like "web1.tpnservers.com" and "eml1.tpnservers.com", and "sql1.tpnservers.com". Again, a long and time doing this and a whole lot of realizing that naming is critical to future growth where as my names are "role-based" and I can therefore easily identify what that server was being used for 10 years down the road... I've also whenever possible kept (and so far successfully) 3-character acronyms including:

eml => email
web => websites
sql => databases
nas => backups (from network attached storage)
app => special app servers

None of these decisions were made over night, and so even my recommendation for this feature wasn't until I felt it had merit to be something to consider. Ofcourse, most of this is based on my experience and idiology so yeah, perhaps not everyone will or should agree with my views :-)

jcameron:
@tpnsolutions so it sounds like you actually prefer it when the database name prefix is randomly generated as well?

tpnsolutions:
Basically I like the way things work in general, the only difference being that the "username" be customizable using the regex.

In server templates post install I set the database and database template to:

${USER}

and

${USER}_

*** The former never gets used though is just a habit since I turn off the creation of database on Virtual Server creation. Therefore databases when created are given a name of "${USER}_wordpress" for instance where ${USER} is a standard substituted variable within Virtualmin already ***

If we want to get technical. Keeping with my seemingly unusual naming schema, I've adopted a new database model.

${USER}_YYMMDDHHmmSS

I decided this would make all databases somewhat more unique, a bit more secure as they are harder to guess and well a fixed width.

Ex.

u1234_220124233612

There's a method to my madness, but that comes with a fruitful history of being in the industry.

I admin a few clients with hundreds of domains combined and have established ways to not only optimize VM over the years but keep the servers running at peak performance and less at risk to common security threats.

So far, my evolving models have served me well.

jcameron:
Have you considered using Virtualmin's command-line API to create domains? That way you can have full control over the username, DB name and prefix, and could generate them randomly using a shell script..

tpnsolutions:
@jcameron,

Absolutely, I have a script called "domain-add.php" which does exactly what I like it to do. I was merely suggesting this feature to bring the same support to the GUI.

I had envisioned the feature being implemented somewhere inside I believe "Virtualmin Configuration" right next to the "Password" option where you can choose between a Random Password amongst other options.

I merely was suggesting (though it seems the idea got a lot more attention then expected -- heh heh) that on that page you could perhaps provide a "regex" field whereby a "custom username" could be "populated" using the "regex" format just like a "Random Password" is automatically populated based on the settings.

Frankly, if the given username was already taken, Virtualmin would simply halt the creation of the Virtual Server, I'd go back a page, and try again :-) Perhaps a javascript function could be added with a button like "generate username" based on "regex"...

Didn't expect the level of integration that was going on in this thread, though I was leaving things to you smarter people :-)

tpnsolutions:
@jcameron,

I actually considered even writing a Google Chrome extension to tackle this idea, though I figured perhaps a native feature would be so much better :-)

PR #348, opened 2021-12-29
update showpass.cgi to have some whitespace via nbsp to make it more …

verne-wv:
…readable


jcameron:
Thanks for the patch!

PR #349, opened 2021-12-30
update a path for subroutine get_php_fpm_socket_file in php-lib.pl to…

verne-wv:
… keep SELinux happy --- as SELinux already has a rule for /var/run/php-fpm.

### DIFF
--- a/php-lib.pl
+++ b/php-lib.pl
@@ -1998,7 +1998,7 @@ return @rv;
 sub get_php_fpm_socket_file
 {
 my ($d, $nomkdir) = @_;
-my $base = "/var/php-fpm";
+my $base = "/var/run/php-fpm";
 if (!-d $base && !$nomkdir) {
 	&make_dir($base, 0755);
 	}


verne-wv:
I see that I mis-spoke .... I should have said (removing the false comment about a symlink)

change from using /var/php-fpm/ to /var/run/php-fpm/ as SELinux already has a rule for the longer path name

jcameron:
We can't just change this path, as it would break existing installs.

verne-wv:
humm .... I am pretty sure restarting php-fpm will create all the new SOCK files on the fly, but I am not sure how to cleanly get rid of the old ones, as well as the /var/php-fpm/ directory.

or ... I do not fully understand all the pieces that use that subroutine and the files/directories it works with  :-)

does php-fpm, when doing a shutdown, erase all those files?  if so then a real shutdown/startup (instead of a reload) might be enough and it will clean itself up -- unless a reload as well will clean things up.  Of course that kind of change has to be very closely coordinated with an apache restart/reload -- but perhaps that dual restart/reload (php-fpm and apache) is what is typically done anyway by VirtualMin ?

in any case, I wanted to at least get it documented as today, out of the box, the current socket config simply does not work on SELinux and causes php-fpm to fail to restart   :-)

jcameron:
The problem with changing this path is that all existing domains on users' systems will now be incorrectly configured according to Virtualmin, and their FPM mode will not be detected properly.

Maybe a better approach would be to make the path configurable for users setting up new systems?

verne-wv:
I have been tinkering with this for a few minutes, and cannot get anything to break -- made some dummy servers/domains, changed things, made some more, went back and adjusted the older ones, and the newer ones ...

what can I do to prove or dis-prove your very valid concern of --

> now be incorrectly configured according to Virtualmin, and their FPM mode will not be detected properly.

Validate Virtual Servers is happy.  (re)Checking Configuration is happy.

Again, the current path is not defined in SELinux, so nothing will work at all -- of course SELinux can be disabled, or a custom "rule" be added to SELinux for the path.

And since we know for sure it will not work, I don't think

> make the path configurable for users setting up new systems

is the answer.

iliarostovtsev:
> Again, the current path is not defined in SELinux, so nothing will work at all 

Why not allow the path we use in SELinux instead?

verne-wv:
I have no idea how hard it is to get the RedHat/CentoOS/Fedora ecosystem to accept a new SELinux rule  :-)

.... and since I believe many other utilities and packages use the /var/run/xxxx format, I could see they would suggest following that syntax and path .... just my off-the-cuff guess.

and again, I cannot get my test Linux box to break in any manner ... will such an adjustment to VirtualMin really affect existing domains/virtual servers ?

If you cannot see your way to making this change in the base, then the next best change would be in fact to make that path adjustable by the admin --- not sure if that should be at a domain level, or global for the entire physical server.

iliarostovtsev:
You could add SELinux rule yourself, for your own case.

verne-wv:
any progress on this ... this edit needs to be applied to the new v7 release too !!

verne-wv:
its your software, but I am still at a loss as to why my change cannot be accepted.

oh well.

iliajie:
@jcameron Jamie, unless there are permission issues, why else don't we use a path to store domain PHP-FPM socket _conditionally_ based on the system? For Debian/Ubuntu it would be `/var/run/php` and for RHEL derivatives it would be as @verne-wv suggested `/var/run/php-fpm`?

jcameron:
We could totally do that, yeah.

verne-wv:
Back in January I could not get anything to break when intermingling virtual servers with the old and new paths ...  is Checking Configuration and/or Validate Virtual Servers the way to check things?

Where do things fall apart for existing configs when this path is different for newer virtual servers in get_php_fpm_socket_file in php-lib.pl?  Or is just dangerous to not have everything consistent and that is your main concern?

And would my suggested path of /var/run/php-fpm work on all distros?

Verne

jcameron:
@verne-wv the issue is that this change would cause Virtualmin to look in the wrong directory for existing domains.

iliajie:
> @jcameron We could totally do that, yeah.

This is risky but should work https://github.com/virtualmin/virtualmin-gpl/pull/436/files, right?

iliajie:
And more for Apache: https://github.com/virtualmin/virtualmin-gpl/pull/436/commits/ae2750f9ce75d0d47ca6f8d1f0f9505b4940e4f6 and Nginx: https://github.com/virtualmin/virtualmin-nginx/pull/34/files

verne-wv:
@jcameron as you know much more about the code than I, I must defer to your comment --- but I did spend like two seconds looking at php-lib and other code, as well as creating, modifying and doing backup/restores and never got anything to break.

jcameron:
Did you test the case where a domain was created before this change, and then managed afterwards? Does it show the correct FPM socket mode?

iliajie:
> Did you test the case where a domain was created before this change, and then managed afterwards? Does it show the correct FPM socket mode?

Yes, most certainly. I tested it before making a PR and also today even more after you asking. But please do not take my word for it, run deeper tests yourself!

jcameron:
@iliajie I assume the tests were in conjunction with https://github.com/virtualmin/virtualmin-gpl/pull/436/files ?

Backing up a bit - is the original problem that's being solved here that SElinux doesn't allow the original `/var/php-fpm` directory?

iliajie:
> I assume the tests were in conjunction with https://github.com/virtualmin/virtualmin-gpl/pull/436/files ?

Yes, correct.

> Backing up a bit - is the original problem that's being solved here that SElinux doesn't allow the original /var/php-fpm directory?

It started out as SELinux problem and later also happened to be a matter of using distro preferred paths.

jcameron:
I'm OK with fixing the SElinux problem, but not a fan of adding complexity to just be compatible with other distros. For example, what happens if a domain is moved from one distro to another via backup / restore?

iliajie:
That was a way to fix SELinux, right?

About restoring a domain - I didn't test this case scenario but I can assume that if the distro is changed the socket file won't match and the mode detection with a current code may fail!

Unless we fix the code to use regex to match it (socket) disregard of the socket path? As in other parts of the code?

By the way, this must not hold a new 7.2 release.

iliajie:
Closed. All further discussions should be done here https://github.com/virtualmin/virtualmin-gpl/pull/436#issuecomment-1230232964.

PR #351, opened 2022-01-07
BUG-FIX for overwriting aws credentials

dkarvounaris:
This keeps overwriting /root/.aws/credentials with the access key, where the secret key is required, which leads to S3 backups failing "mysteriously".
Wrong variable name is being used.


jcameron:
Thanks for the fix!

PR #379, opened 2022-03-16
Remove download immediately option

iliajie:
I propose deleting this (download immediately) option as it requires opening a new tab, can hang for quite some time on large backups without showing any progress.


jcameron:
Maybe let's wait until the next release, as the "download via link" option hasn't launched yet, and there may be bugs.

iliarostovtsev:
Sounds reasonable. Although, I have tried "download via link" option many times and never experienced any issues.

jcameron:
Let's wait till after this release.

iliajie:
Let's remove it already?

jcameron:
Have we had a release that included the new "download via link" feature?

iliajie:
Yes, it's already been with Virtualmin 7.1 release for a while.

PR #382, opened 2022-04-05
Fix to make sure that actual shell exists

iliajie:


iliajie:
Can we use `ldd $ushell_cmd | grep "=> /" | awk '{print $3}'` to get shell's shared libs and copy it to jail?

swelljoe:
Wait. This does exactly what I insisted cannot be done. We cannot copy arbitrary user shells into the jail. It is not safe to do that. The user may be able to control their shell, and users _cannot_ be allowed to control what goes into the jail.

I said adding rbash was OK. I said adding other shells was OK. But, adding _anything_ the user wants is _not_ OK.

iliajie:
> I said adding rbash was OK. I said adding other shells was OK.

Umm, i got it. Isn't my patch only dealing with shells?

>  But, adding anything the user wants is not OK.

I don't understand how and where I'm adding anything else but shell related? Are we on the same page here? :)

iliajie:
Alright, then we can add more shells to `[basicshell]` but that will be hardcoded. I suppose is that what we want?

swelljoe:
>Umm, I got it. Isn't my patch only dealing with shells?

Users can control their own shells. You cannot allow a user to control what goes into the jail. Your patch copies any user shell into the jail. Thus, the user can potentially control what goes into the jail. This _cannot_ happen.

You can add other shells into the default jail configuration, if you'd like. You can't make it so that a user gets to pick what gets put into the jail.

swelljoe:
To be clear: The "shell" field in passwd can potentially point to _any_ command. You keep saying "shell" as though there is some magical meaning to it. But, login shell can potentially be literally any command on the system. You cannot allow users to control what goes into the jail.

iliajie:
> The "shell" field in passwd can potentially point to _any_ command. You keep saying "shell" as though there is some magical meaning to it. But, login shell can potentially be literally any command on the system. You cannot allow users to control what goes into the jail.

That makes more sense. I will get back. Appreciate your comments!

swelljoe:
> Alright, then we can add more shells to [basicshell] but that will be hardcoded. I suppose is that what we want?

Yes, that's fine. Hardcoded is the only thing that is reasonable/safe!

PR #392, opened 2022-05-06
Subject for completed backup mails are wrong

PitWenkin:
Fully completed backups and backups that where completed but where the deletion of older backups failed have their subject lines switched


jcameron:
Thanks for the fix!

Issue #164, opened 2020-05-18
Postgrey PID unwritable in default location


ChristopherW:
On my CentOS 7 install, with Postgrey 1.34 from EPEL, it installs a systemctl file to

`/etc/systemd/system/multi-user.target.wants/postgrey.service`

This file contains `PIDFile=/var/run/postgrey.pid` and a similar reference in the ExecStart command. These need modifying, as Postgrey cannot write to postgrey.pid inside `/var/run`. It was logging an according error in `maillog` which wasn't stopping it working, but was pointless.

I followed the simple steps in this Debian bugreport: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=630353#msg15 but also had to alter the postgrey.service command (obviously `/etc/init.d/postgrey` doesn't exist with systemd). 

Any opportunity to deploy Postgrey 1.37 in Virtualmin deploy and bake the config fix in?

SupRavII:
mkdir /var/run/postgrey
chown postgrey:postgrey /var/run/postgrey

change /etc/init.d/postgrey:
PIDFILE=/var/run/$NAME.pid
to
PIDFILE=/var/run/$NAME/$NAME.pid

/etc/init.d/postgrey stop
rm /var/run/postgrey.pid
/etc/init.d/postgrey start

jcameron:
Seems like this should be fixed upstream in the Postgrey package?

chris001:
I found a `cron` job script to auto update the existing `postgrey` whitelist each 28 days to the newest one available.
I'll make a PR for this.

swelljoe:
We no longer maintain a postgrey package. We use the EPEL package on Red Hat derivatives. We recommend you use the EPEL package (it's always been the EPEL package, we just added it to our repos for reasons I can't recall).

swelljoe:
Does this problem exist in the EPEL package version 1.37, @grzegorz-janoszka ?

grzegorz-janoszka:
@swelljoe I use Debian package 1.36-5.2. I will see if there is a newer version of the package.

grzegorz-janoszka:
I have postgrey 1.37 and it looks much better. I had no idea there were so many changes between 1.36 and 1.37.

GaalexxC:
They will never fix this, here's the complete work around

`mkdir /run/postgrey`
`chown -R postgrey:postgrey /run/postgrey`
`chmod 0755 /run/postgrey/`

[...]

PR #421, opened 2022-07-21
Create setup-postgrey-updater.sh

chris001:
Creates cron daily file to update the postgrey whitelist.

### DIFF
--- a/setup-postgrey-updater.sh
+++ b/setup-postgrey-updater.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+# We are going to setup a newer whitelist for postgrey, the version included in the distribution is old
+cat > /etc/cron.daily/virtualmin-postgrey-whitelist << EOF;
+#!/usr/bin/env bash
+
+# Virtualmin
+
+# check we have a postgrey_whitelist_clients file and that it is not older than 28 days
+if [ ! -f /etc/postgrey/whitelist_clients ] || find /etc/postgrey/whitelist_clients -mtime +28 > /dev/null ; then
+    # ok we need to update the file, so lets try to fetch it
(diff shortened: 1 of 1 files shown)


chris001:
Mentioned in #164 Virtualmin has been omitting to regularly check and update the `postgrey` whitelist from its online source, and should have been updating it ever since `postgrey` feature was added, a long time ago!
Reference: the [postgrey repo](https://github.com/schweikert/postgrey#updating-the-whitelist) says to update.

chris001:
@swelljoe  said a UI element should be added to enable/disable this `cron daily` whitelist updater.  To simplify it is possible to eliminate the requirement for UI element, I believe, all that should be needed, for users to be happy, is to enable this updater when `postgrey` is enabled, and disable this updater when `postgrey` is disabled.  IMHO why would a user ever not want to update the whitelist?  My 2c  @iliajie

swelljoe:
Yeah, that's probably a reasonable choice. I hate adding things to the UI if they aren't needed...but, it definitely shouldn't be on when postgrey isn't in use.

chris001:
Depending on how Virtualmin is calling `postgrey`, might need to modify the call, or modify `/etc/default/postgrey`, to be something like this:
```
POSTGREY_OPTS=\"'--inet=10023 --delay=180 --whitelist-recipients=/etc/postgrey/whitelist_clients'\"
```
...because this whitelist file exists in 2 copies on the system, not sure which one `postgrey` is using:
1. `/etc/postgrey/whitelist_clients`  this is the one we're auto updating.
2. `/usr/share/postgrey/whitelist_clients`  this one is not getting auto updated however maybe `postgrey` on Virtualmin is using it so must point `postgrey` to use the one in `/etc/postgrey`.

jcameron:
I wonder if we should just implement something like this as part of Virtualmin's regular background status check process?

shoulders:
this push should be closed because of the arguments I made here https://github.com/virtualmin/virtualmin-gpl/issues/869 and the fact is they are never updated anyway, I think once in 2 years.

PR #422, opened 2022-07-27
Add `.well-known` location work with proxy enabled sites

iliajie:
Add `.well-known` location work with proxy enabled sites


jcameron:
How will this behave for domains that already have `ProxyPass /.well-known/acme-challenge/ !` ?

iliajie:
Those are meant to be added only on initial setup.

iliajie:
.. therefore it makes no difference.

jcameron:
Thanks for the patch!

PR #423, opened 2022-07-29
Fix to drop adding no longer needed `<Proxy *></Proxy>` directives

iliajie:
Fix to drop adding no longer needed `<Proxy *></Proxy>` directives

### DIFF
--- a/feature-web.pl
+++ b/feature-web.pl
@@ -1522,12 +1522,6 @@ if ($d->{'proxy_pass_mode'} == 1) {
 		# SSL proxy mode
 		push(@dirs, "SSLProxyEngine on");
 		}
-	if ($apache::httpd_modules{'core'} >= 2.0) {
-		# Ensure that proxying works
-		push(@dirs, "<Proxy *>",
-			    "allow from all",
-			    "</Proxy>");
-		}
 	}
 elsif ($d->{'proxy_pass_mode'} == 2) {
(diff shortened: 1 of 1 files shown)


jcameron:
Are you 100% sure this isn't needed? Because I've seen some distros that disallowing proxying in httpd.conf or some global include file, and this directive is needed to re-enable it.

iliajie:
I do not know any contemporary systems (Debian 10-11, Ubuntu 18.04-22.04, Alma * and Rocky 8) that would be blocking it globally by default. If that happens 15 years ago, I believe you, but I don't think it's the case anymore. :-) 

Besides, if nowadays it's blocked globally, perhaps we should respect it (unless it's a distro default, which isn't the case).

jcameron:
I've definitely seen it, which is why I added that block. Does it cause any problems to keep it?

iliajie:
Can you refer to where you have seen it (just curious?) This code was added 15 years ago.

> Does it cause any problems to keep it?

Problems, no - but we shouldn't add things to the configs which are not needed, right? Also, I noticed that it's being added at the bottom of the `VirtualHost` config and _below_ any other proxy-related configs -- is this expected?

jcameron:
Yeah, maybe it's not needed anymore on newer systems. I'll merge this PR ..

iliajie:
Alright. Thanks.

PR #424, opened 2022-07-29
Use `Require all granted` directives with Apache 2.4

iliajie:
Use `Require all granted` directives with Apache 2.4


feature-webalizer.pl
```diff
@@ -641,7 +641,7 @@ sub create_webalizer_htaccess
 &print_tempfile(HTACCESS, "AuthUserFile $passwd_file\n");
 &print_tempfile(HTACCESS, "require valid-user\n");
 &print_tempfile(HTACCESS, "<Files .stats-htpasswd>\n");
-&print_tempfile(HTACCESS, "deny from all\n");
+&print_tempfile(HTACCESS, "Require all denied\n");
```
jcameron, line 644: What about older Apache versions?

iliajie: I suggest we ignore it for Virtualmin. Neither distros have it (Apache 2.2), not even CentOS 7. With newer versions of Virtualmin we should focus on systems that are not EOL.

jcameron: Couldn't we just add a simple test for the Apache version here though?

iliajie: I was about to do this (before the commit) but then realized that we're simply going to make Virtualmin code harder for no reason.

Do you really want to add that kind of a test?

jcameron: You could do it with code like : 

```
&require_apache();
if ($apache::httpd_modules{'core'} < 2.4) {
    &print_tempfile(HTACCESS, "deny from all\n");
} else {
    &print_tempfile(HTACCESS, "Require all denied\n");
}
```

iliajie: Yes, of course. I will fix it if you believe it's still needed.

jcameron: Yes, please do.

php-lib.pl
```diff
@@ -1320,7 +1320,7 @@ sub save_domain_php_directory
 			local @lines = (
 				"    <Directory $dir>",
 				"        Options +IncludesNOEXEC +SymLinksifOwnerMatch +ExecCGI",
-				"        allow from all",
+				"        Require all granted",
```
jcameron, line 1323: Also make this conditional on the Apache version

iliajie: Okay, that is done.

iliajie: https://github.com/virtualmin/virtualmin-gpl/pull/424/commits/de8365054fb37206a16482cb5c4feb6874079af3

PR #425, opened 2022-07-29
FPM improvements

iliajie:
1. [Set PHP-FPM pool file permission](https://github.com/virtualmin/virtualmin-gpl/commit/27f86efcd396da1d0455f396fbc76cac7ec8ea76) - This is required for virtual server owners to edit (when allowed) pool config file
  2. [Fix to print better message when no PHP config changed upon saving](https://github.com/virtualmin/virtualmin-gpl/commit/bb55d111f53ddce2d013fa09fc1c96f2cd90ab4e)
  3. [Default children value cannot be lower than max_spare_servers](https://github.com/virtualmin/virtualmin-gpl/commit/bf66dca63b1a8c0963123e36c8788b5c495b273c)
  4. [Fix sanity check (cannot be lower than default pm.max_spare_servers)](https://github.com/virtualmin/virtualmin-gpl/commit/5985266974657d743824585ea18994714742f8f0)
  5. [Fix to update text and help to accommodate both FPM and FCGId](https://github.com/virtualmin/virtualmin-gpl/commit/671bd9b90e1b238c286869669e25382d9db221e6)
  6. [Update help and language translations](https://github.com/virtualmin/virtualmin-gpl/commit/9db72d2d454133e1dce44945c7fc67888c45ee51)
  7. [Fix sanity check (cannot be lower than default pm.max_spare_servers)](https://github.com/virtualmin/virtualmin-gpl/commit/5985266974657d743824585ea18994714742f8f0)


iliajie:
Although, `#1` may be not a good idea, as it could give a user to still edit a pool file aside from limits.

Then we should change it on `phpini` module itself?

iliajie:
No, I have a better idea. Will add a commit to the PR after the walk..

iliajie:
I think I fixed all known PHP related issues (with Apache) ...

save_phpmode.cgi
```diff
@@ -11,7 +11,8 @@ $canv = &can_edit_phpver($d);
 $can || $canv || &error($text{'phpmode_ecannot'});
 &require_apache();
 $p = &domain_has_website($d);
-
+@modes = &supported_php_modes($d);
+($newmode) = grep { $in{'mode'} =~ /^$_/ } @modes;
```
jcameron, line 15: Why use a regexp here?

iliajie: 🙂  This is because (most probably a theme bug, which I will check on a bit later today), when you hit enter to submit a form it adds `mode` two time to the payload which makes `$in{'mode'}` set to something like `fpmfpm` ..

php-lib.pl
```diff
@@ -2251,7 +2251,12 @@ sub create_php_fpm_pool
 	# Create a new file
 	my $tmpl = &get_template($d->{'template'});
 	my $defchildren = $tmpl->{'web_phpchildren'};
-	$defchildren = 9999 if ($defchildren eq "none" || !$defchildren);
+	if ($defchildren eq "none" || !$defchildren) {
+		$defchildren = 9999;
+		}
+	else {
+		$defchildren = $defchildren >= 5 ? $defchildren : 5;
```
jcameron, line 2258: Why this limit of 5 ?

save_phpmode.cgi
jcameron: Hold on, we should fix that bug in instead of working around it here!!

php-lib.pl
iliajie: > Why this limit of 5 ?

Check this updated help, it explains it - [help/phpmode_children.html](https://github.com/virtualmin/virtualmin-gpl/commit/671bd9b90e1b238c286869669e25382d9db221e6#diff-4a8298e41f8c82d19fa5b12a6dfb27e5f79c2b0724ebb3c1c4c352bedd4227dd)

save_phpmode.cgi
iliajie: > Hold on, we should fix that bug in instead of working around it here!!

Absolutely, although this was before I realized that it was a bug! Although, this a small improvement to the code here anyway.

php-lib.pl
jcameron: Will setting it to less than 5 create an invalid config that can't be loaded by PHP?

iliajie: Yes, unless `pm.max_spare_servers` also manually decrease. But we have it hard coded upon pool creation time .. so I decided to make it easy but not obscure ..

save_phpmode.cgi
jcameron: Assuming that bug is fixed, can't this line just be `$newmode = $in{'mode'};`

iliajie: Yes, let's do it. Fixed.

> Assuming that bug is fixed,

Yes, I have been fighting with it today many hours to make it right on the theme side ... finishing final tests ..

php-lib.pl
jcameron: Ok, makes sense!

iliajie:
Thanks. I will create another PR if needed when taking another deeper look into it ..

PR #434, opened 2022-08-24
Add repos check and throw an error if outdated

iliajie:
@swelljoe Joe, check out this implementation of what we were discussing yesterday.


virtual-server-lib-funcs.pl
```diff
@@ -15733,6 +15733,33 @@ sub check_virtual_server_config
...
+		# File exists, but does it contain correct repo links
+		$repolines = &read_file_lines($virtualmin_yum_repo, 1)
+			if (-r $virtualmin_yum_repo);
+		}
+	elsif ($gconfig{'os_type'} eq 'debian-linux') {
+		# File exists, but does it contain correct repo links
+		$repolines = &read_file_lines($virtualmin_apt_repo, 1)
+			if (-r $virtualmin_apt_repo);
+		}
+	# Does repo file has correct links?
```
jcameron, line 15749: What if it's redhat or debian system, but Virtualmin wasn't installed from a repo?

iliajie: You're right. I forgot about that kind of install type. Joe mentioned it too though. Now it's fixed https://github.com/virtualmin/virtualmin-gpl/pull/434/commits/fc1a4ebba16abc15b3435f2205e5893ec90bb61a.

virtual-server-lib-funcs.pl
```diff
@@ -15733,6 +15733,36 @@ sub check_virtual_server_config
...
+	my $repofound;
+	if (-r $virtualmin_yum_repo ||
```
jcameron, line 15751: Why not just use `if ($repolines) {`

iliajie: Because technically the file can be empty.

jcameron: Maybe try `if (defined($repolines)) {`

iliajie: No, as we would have to use globals. If you don't want files to be re-read twice, we can do this - https://github.com/virtualmin/virtualmin-gpl/pull/434/commits/f8a7cdf6bf9d0b6b4d274e326e080c897764b2f0.

iliajie: Oh, wait. Actually it's possible in Perl. There you go - https://github.com/virtualmin/virtualmin-gpl/pull/434/commits/b362789377242d14edffd499f943f34411f94896.

jcameron:
Thanks,  the patch looks good now!

iliajie:
Great, thanks!

Issue #433, opened 2022-08-22
Script installers does not check for supported PHP versions


PitWenkin:
Script installers offers update to Roundcube 1.6.0. Which should not be done on Ubuntu 18.04 as only PHP 7.2. is installed by default:
```
php --version
PHP 7.2.24-0ubuntu0.18.04.13 (cli) (built: Jul  6 2022 12:23:22) ( NTS )
```

[Roundcube dropped support for PHP < 7.3 in 1.6.0
](https://github.com/roundcube/roundcubemail/releases/tag/1.6.0)

There is no mention of this incombatibility during the update progress. I even get emails about 'Virtualmin script upgrades available' which lists the script.

Either there is no check performed for needed php versions or it is broken …

iliajie:
Hello,

Thank you for your suggestion! We will fix that.

PR #435, opened 2022-08-24
Fix PHP version min and max tests for scripts

iliajie:
This patch finally fixes PHP version tests for installable scripts. 

Tested with `b2evolution` (limits **maximum** PHP version to 7.4 for installed script) and `RoundCube` (limits **minimum** PHP version to 7.3 for installed script).

It's important to test against current PHP version that virtual server uses, not all available.

Note: Tried on Ubuntu 22.04 with PHP 7.4 and PHP 8.1.


jcameron:
Can you explain how this fix works compared to the original code?

iliajie:
Yes.

User: Install script version 1
Script: Check if version 1 has PHP deps

And then we check which PHP version is currently setup for virtual server. For example 8.1 but the script has a limit for PHP 7.4 only. So it will show an error.

Now, RoundCube 1.6.0 wants PHP 7.3 at least. We need to check which PHP version is _currently_ enabled for virtual server. We do this now.

Before it was checking on all available PHP versions for a virtual server but this is not what the script expects. We need to test it against currently enabled PHP version.

jcameron:
Is there a particular script and version upgrade from/to that you saw trigger this issue?

iliajie:
Yes, latest submitted RoundCube and its `script_roundcube_php_fullver` are ignored with current Virtualmin `scripts-lib.pl` code. 

This issues, to be reproduced requires that you have more than one PHP version installed. Try older distros with default PHP 7.2 (Ubuntu 18 for instance). Then install PHP 8.1 as well. So the system has PHP 7.2 and PHP 8.2 installed.

Current code when testing `script_roundcube_php_fullver` will test the highest available PHP version on the system, which **_cannot be right_**.

The code in PR will make sure before installing/upgrading a script that given virtual server is using required PHP version.

So, currently, if you have PHP 7.2 configured for a given virtual server (and also PHP 8.1 installed!) then current code will ignore the requirement:

```
sub script_roundcube_php_fullver
{
local ($d, $ver, $sinfo, $phpver) = @_;
my $phpver = &compare_versions($ver, "1.6.0") >= 0 ? "7.3" : 5.6;
return $phpver;
}
```

.. and will allow to install RoundCube 1.6.0 which only works with PHP 7.3+. But a given virtual server has PHP 7.2 setup and configured (not 8.1).. So it won't work.

This PR will test against **_currently_** setup PHP version on a virtual server and not just any available PHP version on the system.

This issue happened for years.

jcameron:
So I tried to re-produce the original issue on a system with PHP versions 5.4.16, 7.2.34, 7.3.33 and 7.4.30 by doing the following : 

1 - Create a test domain called example.com with PHP FPM mode with `virtualmin create-domain --domain example.com --dir --unix --web --dns --mysql --pass smeg`
2 - Set the PHP version for the domain to 7.2.34 with `virtualmin set-php-directory --domain example.com --dir . --version 7.2`
3 - Verified that this has taken effect with `virtualmin list-php-directories --domain example.com --multiline`
4 - Installed the latest Roundcube with `virtualmin install-script --domain example.com --type roundcube --version latest --path / --db mysql example`
5 - Checked that Roundcube works, and that PHP 7.3.33 is in use with `virtualmin list-php-directories --domain example.com --multiline`

So I'm having trouble understanding the original issue this patch is fixing?

iliajie:
The latest version of RoundCube 1.6.0 should only install when PHP 7.3+ is available. You said you have 7.2 setup for the virtual server, right? Then this is a bug! This is what my PR fixes. It must stop you from installing it when PHP 7.3 is not available.

iliajie:
I assume that you wouldn't change the PHP version automatically without a warning, when installing a script as it could break other installed scripts.

iliajie:
But changing the PHP version automatically is an excellent idea in FCGID mode and when per-directory PHP versions are available.

However, it should also work backwards, to change the PHP version downwards.

And in FPM mode, as it doesn't have per directory PHP versions (yet), it should do what PR does, right?

jcameron:
What Webmin has always done is upgrade the PHP version to support the script being installed, if possible. However, it will fail if the required PHP version isn't available at all. And in FPM mode, it can only update the PHP version for the whole domain since per-directory isn't supported (yet).

iliajie:
Excellent! I checked your latest patches. I see what was the initial intention now. I did not know that PHP versions could be automatically changed to accommodate script's needs. It seems that PHP-FPM support was added much later after this feature was initially introduced.

I can also see that using latest patches in FPM mode, now properly shows an error message and doesn't change PHP version globally (when using RoundCube with `php_fullver` sub):

image


However, there is **still a problem** for `b2evolution` script which has a requirement for **maximum** PHP version (7.3 and below) set in `php_maxver_func` sub. In fact it's not even possible to install it at all, even if correct and expected PHP version (<= 7.3) is selected. Disregard of the mode (i.e. the same happens in FPM mode) used. Check screenshots below:


image


image

jcameron:
Ok, the real problem is actually in the unnecessary extra dependencies check. I removed it here : https://github.com/virtualmin/virtualmin-gpl/commit/623f68218063fcab3569941c2c302210fd471e1b

PitWenkin:
As soon as this is available via normal Ubuntu updates I will test it for roundcube


These are the PHP settings for the domain in #433:
PHP settings

iliajie:
Is there a specific reason for you to use _mod_php_?

PitWenkin:
> Is there a specific reason for you to use _mod_php_?

It has been setup up like that by my predecessor, and I don't want to change it now (as roundcude is the only reason PHP is installed on the server, and everytime I tinker with PHP everything goes bellyup)… I might change it when I upgrade the server … but then PHP will be udpated too and the problem solves itself

PitWenkin:
I just got offered [Roundcube 16.1.1 ](https://github.com/roundcube/roundcubemail/releases/tag/1.6.1) which reminded me that this PR is still open.

As I will soon update the related server including PHP, there is not much time left for me to test ;-)

iliajie:
Thanks. I remember we came up with something else (familiar though) with Jamie. I will have to test it all over again with Virtualmin 7.5 and see if it does things right. It actually should do it but I will run tests just in case. I'm on finishing another big project at the moment.

PR #446, opened 2022-09-14
Update existing domain Virtualmin website default page

iliajie:
This patch will solve the problem with existing and outdated Virtualmin website default (landing) page.

* Update template but preserve user existing values on the `index.html` file
* Update template but preserve existing values on the `disabled_by_virtualmin.html` file

### DIFF
--- a/postinstall.pl
+++ b/postinstall.pl
@@ -369,8 +369,10 @@ my %uconfig = &foreign_config("useradmin");
 my $uhome   = $uconfig{'home_base'};
 my $readdir = sub {
     my ($dir) = @_;
+    my @hdirs;
+    return @hdirs if (!-d $dir);
     opendir(my $udir, $dir);
-    my @hdirs = map {&simplify_path("$dir/$_")}
+    @hdirs = map {&simplify_path("$dir/$_")}
       grep {$_ ne '.' && $_ ne '..'} readdir($udir);
     closedir($udir);
     return @hdirs;
--- a/postinstall.pl
+++ b/postinstall.pl
@@ -392,16 +392,18 @@ if (&foreign_check("useradmin")) {
 	                if ($l =~ /\s*<title>(.*?)<\/title>/) {
 	                    $eptitle = $1;
 	                    }
-	                if ($l =~ /\*\s(Virtualmin\sLanding|Website\sDefault\sPage)\sv([\d+\.]+)/) {
+	                
+	                # Test to make sure that given file is Virtualmin website default page
+	                if (!$efix && $l =~ /\*\s(Virtualmin\sLanding|Website\sDefault\sPage)\sv([\d+\.]+)$/) {
 	                    my $tmplver = $2;
-	                    if ($tmplver && &compare_version_numbers($tmplver, "2.0") < 0) {
-	                        $efix++;
-	                        }
+	                    $efix++ if ($tmplver && &compare_version_numbers($tmplver, "2.0") < 0);
 	                    }
+	                $efix++ if ($efix == 1 && $l =~ /\*\sCopyright\s+[\d]{4}\sVirtualmin,\sInc\.$/);
+                    $efix++ if ($efix == 2 && $l =~ /\*\sLicensed\sunder\sMIT$/);
 
 	                # If existing file is a template and needs a fix extract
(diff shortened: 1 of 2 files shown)


postinstall.pl
```diff
@@ -363,6 +363,103 @@ sub module_install
 		}
 	}
 
+# Always update outdated (lower than v2.0)
+# Virtualmin website default (landing) page
+if (&foreign_check("useradmin")) {
```
jcameron, line 368: You can assume this module is always installed

iliajie: Well, yes, but I actually wanted to make variables scoped to the block. Should I remove it?

iliajie: Alright, it will work too. Removed.

postinstall.pl
```diff
@@ -363,6 +363,107 @@ sub module_install
...
+    my @hdirs;
+    return @hdirs if (!-d $dir);
+    opendir(my $udir, $dir);
+    @hdirs = map {&simplify_path("$dir/$_")}
+      grep {$_ ne '.' && $_ ne '..'} readdir($udir);
+    closedir($udir);
+    return @hdirs;
+    };
+my @hdirs = &$readdir($uhome);
+foreach my $hdir (@hdirs) {
```
jcameron, line 381: Instead of looping over home directories, why not loop over all virtual servers from list_domains and for each read the files under the directory set by the 'home' field?

iliajie: Yes, that could also work. But are there advantages for that?

postinstall.pl
```diff
@@ -363,6 +363,107 @@ sub module_install
...
+        my @dpubifiles = &$readdir($dpubdir);
+        @dpubifiles = grep (/^$dpubdir\/(index\.html|disabled_by_virtualmin\.html)$/, @dpubifiles);
+        foreach my $dpubifile (@dpubifiles) {
+            my $dpubifilelines = &read_file_lines($dpubifile, 1);
+            my ($efix, $eptitle, $eerr, $edom, $etitle, $emessage);
+            foreach my $l (@{$dpubifilelines}) {
+            	
+            	# Get beginning of the string for speed and run
+            	# extra check to make sure we have a needed file
+            	$l = substr($l, 0, 256);
```
jcameron, line 394: How long is this string that we need to take a substring??

iliajie: > How long is this string that we need to take a substring??

It can be super long. Up to 280000 bytes.

postinstall.pl
jcameron: It's just much simpler, and will also work if some domain has it's home directory elsewhere

iliajie: Okay, I will rework it then ..

postinstall.pl
```diff
@@ -363,6 +363,107 @@ sub module_install
...
+                    $hashtmp{'dom'}            = $edom     || $d->{'dom'};
+                    $hashtmp{'TMPLTTITLE'}     = $etitle   ||
+                                                 ($ddis ? $text{'deftmplt_website_disabled'} :
+                                                          $text{'deftmplt_website_enabled'});
+                    $hashtmp{'TMPLTCONTENT'}   = $emessage || "";
+                    $hashtmp{'TMPLTSLOGAN'}    = $ddis ? $text{'deftmplt_disable_slog'} :
+                                                         $text{'deftmplt_default_slog'};
+                    $domtmplfilecontent =
+                      &substitute_virtualmin_template($domtmplfilecontent, \%hashtmp);
+                    eval {
```
jcameron, line 453: Why the use of eval here?

iliajie: In case it won't be able to write the file for whatever odd reason.

jcameron: Better to just set the 4th param of open_tempfile_as_domain_user to 1 (so it returns 0 on error opening the file), and only write if the function returns 1.

postinstall.pl
```diff
@@ -363,6 +363,103 @@ sub module_install
 		}
 	}
 
+# Always update outdated (lower than v2.0)
+# Virtualmin website default (landing) page
+my $readdir = sub {
```
jcameron, line 368: Do we even need this as a sub if it's only called once?

PR #450, opened 2022-09-21
Add support to force-refresh domains expiration status

iliajie:
Add support to force-refresh domains expiration status

image


virtual-server-lib-funcs.pl
```diff
@@ -11376,6 +11376,14 @@ sub list_warning_messages
 	}
 if (@expired || @nearly) {
 	my $exp;
+	my $getdnsrefform = sub {
```
jcameron, line 11379: Why put this into a separate sub if it's only called one?

iliajie: Oh, right. It was meant to be called twice at first but then I changed the caller and simply left it that way. Let's just keep it?

iliajie: Okay, removed - https://github.com/virtualmin/virtualmin-gpl/pull/450/commits/4642296ad40d7396022bfb64753c20d52a5771e8

PR #455, opened 2022-09-28
Fix to prefix `other` envs vars to preven clashes

iliajie:


iliajie:
@jcameron Jamie, please check this PR. `$ENV{'SCRIPT_NAME'}` is used by the `miniserv`.

jcameron:
Won't this break users' existing scripts that depend on the old variables? Also, does it even matter if SCRIPT_NAME is overridden?

iliajie:
> Won't this break users' existing scripts that depend on the old variables?

No, because it keeps old variables too.

> Also, does it even matter if SCRIPT_NAME is overridden?

Probably not, but should be expected to work! Nevertheless, it's not even set with the current code (i.e. `script_name`).

jcameron:
Ok, thanks!

PR #458, opened 2022-09-30
Fix to allow resellers virtual servers change PHP mode too

iliajie:


edit_phpmode.cgi
```diff
@@ -23,7 +23,7 @@ print &ui_hidden_table_start($text{'phpmode_header'}, "width=100%", 2,
 			     "phpmode", 1, [ "width=30%" ]);
 
 if ($can) {
-	if (!$d->{'alias'} && $can == 2 &&
+	if (!$d->{'alias'} && $can >= 1 &&
```
jcameron, line 26: Just make it `if (!$d->{'alias'} && $can &&`

iliajie: Fixed and +1 more. - https://github.com/virtualmin/virtualmin-gpl/commit/a4e19c983ce206f88cb91b7af5e702190ce59949

jcameron:
Thanks for the fix!

PR #461, opened 2022-10-02
Add more finesse control over virtual server PHP-FPM pool config

iliajie:
This change is important for keeping virtual server PHP-FPM pool config more balanced and performant.


jcameron:
Looks good, thanks!

PR #466, opened 2022-10-14
Fix DAV directives fully

iliajie:
Fix DAV directives fully


iliajie:
Fixed here in familiar way https://github.com/virtualmin/virtualmin-gpl/commit/24fc2a5977087c2e15dcf754c3fb45aca5f70299

PR #468, opened 2022-10-19
Add a feature to display DNS text records in a column

iliajie:
This PR adds ability to display DNS test records nicely aligned column like.

image


jcameron:
Would it make more sense to do this inside dns_records_to_text when we still have the record fields, rather than re-parsing the records text?

iliajie:
Yes, we could but then there would be no re-usable function. I thought we could use it somewhere else .. Besides, the text size is miserable. 

Although, it's easy to change but I would rather kept it in a separate sub unless you want it to be melded in.

PR #470, opened 2022-10-20
Add help for --letsencrypt and --letsencrypt-always

xxorax:
Just add doc and missing help for --letsencrypt-always

### DIFF
--- a/create-domain.pl
+++ b/create-domain.pl
@@ -66,9 +66,9 @@ used. However, you can override this with the C<--cloud-dns> flag followed by
 either C<local> to host locally, C<services> to use Cloudmin services, or
 the ID of one of the supported providers like C<route53> or C<google>.
 
-Use the --letsencrypt option to generate automaticaly a Let's Encrypt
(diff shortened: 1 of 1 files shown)


create-domain.pl
```diff
@@ -66,6 +66,10 @@ =head1 create-domain.pl
 either C<local> to host locally, C<services> to use Cloudmin services, or
 the ID of one of the supported providers like C<route53> or C<google>.
 
+Use the --letsencrypt option to generate automaticaly a Let's Encrypt
+certificate with auto-renewal. Use  --letsencrypt-always to also run checks
```
iliajie, line 70: This is actually the opposite - _not_ to run any checks. Also, use `C<--letsencrypt>` and `C<--letsencrypt-always>`.

And the right word would be not `generate` but `request`.

xxorax: My bad, I had been foolish by [create_initial_letsencrypt_cert](https://github.com/virtualmin/virtualmin-gpl/blob/master/virtual-server-lib-funcs.pl#8182). Fixed.

PR #471, opened 2022-10-20
Fix mutliline port output

xxorax:
Fix a typo on the port line


iliajie:
Thanks for the patch!

PR #473, opened 2022-10-30
Add file manager link to the top level

iliajie:


virtual-server-lib-funcs.pl
```diff
@@ -12869,6 +12869,18 @@ sub get_domain_actions
 		    'icon' => 'page_edit',
 		  });
 	}
+
+if (&foreign_available("xterm") &&
+    !$d->{'alias'} && $d->{'user'}) {
+	# Link to Terminal for master admin
```
jcameron, line 12875: Should this be visible to the domain owner, or only the master admin? If the latter, you should add `&master_admin()` to the if statement.

iliajie: It has to be shown to all users that have access to `xterm`, like File Manager.

It's convenient to have it shown easily accessible. 

image

jcameron: Right but for other users, will that user= parameter work?

iliajie: No, it will be ignored. Related - https://github.com/webmin/webmin/pull/1770/files

jcameron: Ok, that's reasonable

PR #479, opened 2022-11-24
Allow root login with login link

xxorax:
This allow root login with the `create-login-link`command.

It's working, but I'm not convince about how to handle priviledged accounts like root : there maybe others (sudoers, but I do not use sudo), and root account may be not named "root" (rare cases).

In fact I did not found a way to get all the users allowed to login on Virtualmin, and it's the feature that this command should provide.

Thanks to let me know your suggestions and I will update the PR.


jcameron:
Thanks for this patch!

PR #483, opened 2022-12-04
Fix a new PHP logging

iliajie:
1. If we pass an empty string to to `error_log` in FPM mode, FPM config for virtual server will be saved as such, i.e. `php_value[error_log] =` defaulting to `/usr`. It can be seen using `phpinfo()` function. This issue can be reproduced by disabling logging after it was enabled upon virtual server creation.
  2. PHP logging control should be in PHP Options page. Having it under Website Options is very confusing. Also, it must always be possible to control it, as having it magically disappear is also very confusing.

### DIFF
--- a/php-lib.pl
+++ b/php-lib.pl
@@ -2655,7 +2655,11 @@ my ($d) = @_;
 my $mode = &get_domain_php_mode($d);
 my $phplog;
 if ($mode eq "fpm") {
-	$phplog = &get_php_fpm_ini_value($d, "error_log") || "";
+	$phplog = &get_php_fpm_ini_value($d, "error_log") || 
+	# `error_log` can never be an empty string,
+	# otherwise it will be set to default /usr
+	# if defined but the empty string in FPM
+	undef;
 	}
 elsif ($mode ne "none" && $mode ne "mod_php") {
 	$phplog = "";
--- a/edit_phpmode.cgi
+++ b/edit_phpmode.cgi
@@ -179,6 +179,21 @@ if ($canv && !$d->{'alias'} && $mode ne "mod_php") {
 		}
 	}
 
+if ($mode ne 'none') {
+	$plog = &get_domain_php_error_log($d);
+	$defplog = &get_default_php_error_log($d);
+	$mode = !$plog ? 1 :
+		$plog eq $defplog ? 2 : 0;
+	print &ui_table_row(&hlink($text{'phpmode_plog'}, 'phplog'),
+		&ui_radio_table("plog_def", $mode,
+		[ [ 1, $text{'phpmode_noplog'} ],
+		  [ 2, $text{'phpmode_defplog'},
+		       "<tt>$defplog</tt>" ],
+		  [ 0, $text{'phpmode_fileplog'},
+		    &ui_textbox("plog", $mode == 0 ? $plog : "", 60) ],
+		]));
+	}
+
 print &ui_hidden_table_end();
 
 # Show PHP information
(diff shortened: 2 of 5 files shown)


jcameron:
Since we just want to do a new release soon, can you split this up so that there's a PR for just the bugfix.

iliajie:
All of these changes are bug fixes. Because, previously when changing PHP logging in FCGId mode Apache/FCGId (with Nginx) wasn't restarted. Besides, we should not release a new Virtualmin version with PHP logging controls placed under Websites Options, as in the past we decided to put all PHP related options on one page.

jcameron:
I actually prefer to keep this logging option where it is, because the other webserver log fields are on the same page.

iliajie:
> because the other webserver log fields are on the same page.

I could certainly move it back but we have PHP Options page, where as Website Options page incorporates options for which we don't have separate page, like `Ruby script execution mode` and other options related to website (webserver). PHP logging is less related to a website or a webserver.

php-lib.pl
```diff
@@ -2655,7 +2655,11 @@ sub get_domain_php_error_log
 my $mode = &get_domain_php_mode($d);
 my $phplog;
 if ($mode eq "fpm") {
-	$phplog = &get_php_fpm_ini_value($d, "error_log") || "";
```
jcameron, line 2658: I don't see why this is needed, because `get_php_fpm_ini_value` already returns `undef` if `error_log` isn't set.

iliajie: > I don't see why this is needed, because `get_php_fpm_ini_value` already returns `undef` if `error_log` isn't set.

It's getting triggered in [this code](https://github.com/virtualmin/virtualmin-gpl/commit/0b1d178#diff-aa98c4c01f2f6cde518949f907eb960ca25f3f210a756e7729ad4c503bdd8dddR597-R600). Unless we do it, FPM pool config file gets:

image

And `phpinfo()` will have `error_log` set in FPM mode to:

image

jcameron:
Actually I think this is the best fix : https://github.com/virtualmin/virtualmin-gpl/commit/6d6939bf2c19843208afd23639e6fe860648813b

iliajie:
> Actually I think this is the best fix : [6d6939b](https://github.com/virtualmin/virtualmin-gpl/commit/6d6939bf2c19843208afd23639e6fe860648813b)

Before sending this PR to you, I had done lots of testings. I had tried **_exactly the same_** fix the very first moment, as it seemed as the most obvious. However, running further more deep tests made it fail.

When logging is first enabled on domain creation time (in templates), and then you go and disabled PHP logging, and then changing PHP execution mode to first FCGId, and then again to FPM and back to FCGId it fails with the following error:

image


So, this fix cannot solely alone be the best one. We need to either do what I do in my PR (send an empty string in FCGId mode and `undef` in FPM) or fix that popping up error.

jcameron:
That's a different issue - but this patch should fix it https://github.com/virtualmin/virtualmin-gpl/commit/b415abd2ab8aabb4051cf9d64c4820b6c5716a36

iliajie:
> That's a different issue - but this patch should fix it [b415abd](https://github.com/virtualmin/virtualmin-gpl/commit/b415abd2ab8aabb4051cf9d64c4820b6c5716a36)

Excellent, thank you! It works too. Let's keep it this way, if you like it more like that. But essentially out patches do exactly the same, just slightly different. I will close this PR and make a new one to move logs controls to PHP Options page (to be consistent with the Templates).

PR #485, opened 2022-12-06
Fix PHP logging controls

iliajie:
This PR changes and fixes:

  1. Changes the location of PHP log controls to be placed inside of PHP Options page. This will also correspond with what we currently have in Templates, where PHP log controls are under PHP Options section as well https://github.com/virtualmin/virtualmin-gpl/commit/03f2d546ed07224ab4322ba4c14bc3c8aabd3253
  2. Fixes a problem where in CGI/FCGID mode a new changes to `error_log` was not applied https://github.com/virtualmin/virtualmin-gpl/commit/88bbbf2bdba0549e7b385148bc6fb03abcffa784

### DIFF
--- a/edit_phpmode.cgi
+++ b/edit_phpmode.cgi
@@ -179,6 +179,21 @@ if ($canv && !$d->{'alias'} && $mode ne "mod_php") {
 		}
 	}
 
+if ($mode ne 'none') {
+	$plog = &get_domain_php_error_log($d);
+	$defplog = &get_default_php_error_log($d);
+	$mode = !$plog ? 1 :
+		$plog eq $defplog ? 2 : 0;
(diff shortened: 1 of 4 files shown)


edit_phpmode.cgi
```diff
@@ -179,6 +179,21 @@ if ($canv && !$d->{'alias'} && $mode ne "mod_php") {
 		}
 	}
 
+if ($mode ne 'none') {
```
jcameron, line 182: Also check for $mode ne 'mod_php'

iliajie: Fixed. https://github.com/virtualmin/virtualmin-gpl/pull/485/commits/d134b010cbaebea60cd972ef1afafd02fe23d941

PR #490, opened 2022-12-13
Fix `milter-greylist` username for RHEL systems

iliajie:


ratelimit-lib.pl
```diff
@@ -35,7 +35,7 @@ sub get_ratelimit_init_name
 sub get_ratelimit_user
 {
 my $type = &get_ratelimit_type();
-return $type eq 'redhat' ? 'greylist' :
+return $type eq 'redhat' ? 'grmilter' :
```
jcameron, line 38: Is this the username on *all* CentOS versions and variants? Pretty sure it's `greylist` on CentOS 7 for example.

jcameron: Apologies, you are correct ... I will merge this commit.

PR #494, opened 2022-12-27
Fix to regenerate DNSSEC records upon restore

iliajie:


iliajie:
Now I understand how it works, after looking in to this for hours. All we care about is private and public keys. DS key is just a hash of public DNSKEY. So upon transfer we preserve those files (private and public keys) by restoring them from the backup, and regenerate the records to keep them fresh on the target machine.

iliajie:
Will be suppressed with another upcoming PR.

PR #495, opened 2022-12-28
Fix SPF records upon restore

iliajie:


feature-dns.pl
```diff
@@ -2500,7 +2500,7 @@ sub restore_dns
 	if ($changed) {
 		local $str = &bind8::join_spf($spf);
 		$r->{'values'} = [ $str ];
-		&modify_dns_record($recs, $file, $r);
+		&modify_dns_record($recs, $zonefile, $r);
```
iliajie, line 2503: Also, two lines above this code the line `$str = &bind8::join_spf($spf);` has a bug, because `$spf` gets a new `ipv4` pushed but it's not getting added to the `$str` string. I'm not sure if it's Bind `join_spf` is misbehaving or it can be fixed in here.

iliajie:
> Also, two lines above this code the line `$str = &bind8::join_spf($spf);` has a bug, because `$spf` gets a new ipv4 pushed but it's not getting added to the `$str` string. I'm not sure if it's Bind join_spf is misbehaving or it can be fixed in here.

Jamie, thanks for merging it. Although, it won't be useful unless we also fix what I just mentioned in a blockquote above.

Here is the [example](https://github.com/virtualmin/virtualmin-gpl/files/10313985/ubuntu22-pro-3.virtualmin.dev.tar.gz) backup which you could restore to see what I mean. The first local IP in SPF record is not being updated, while public IP is updated as expected.

jcameron:
In that backup, what were the local and external IPs on the original system?

The current restore code only adds the new system's IPs, it doesn't remove the old ones.

iliajie:
> In that backup, what were the local and external IPs on the original system?

Oh, I see. Yes, on that backup I have two addresses, first local and second public. Local IP (first on the list) is not getting updated and Public IP (last, second on the list) is getting updated properly.

> The current restore code only adds the new system's IPs, it doesn't remove the old ones.

The Public IP is actually getting updated before _this_ code is run. I assume that both IPs need to be updated, right?

jcameron:
Ok I see what you mean now - this commit will fix it : https://github.com/virtualmin/virtualmin-gpl/commit/722e7251eb29d7d5f63564a2d52d0f692fd715a8

PR #496, opened 2022-12-30
Fix DNSSEC upon restore

iliajie:
This patch will:
  1. Toggles DNSSEC to fully refresh records, after private and public keys were changed upon restore
  2. Prints DNSSEC related stuff:
     image

  3. Re-obtains removed DNS lock


feature-dns.pl
```diff
@@ -2514,9 +2541,27 @@ sub restore_dns
 
 &post_records_change($d, $recs, $zonefile);
 &register_post_action(\&restart_bind, $d);
-&$second_print($text{'setup_done'});
-
 &release_lock_dns($d, 1);
+
+# Refresh DNSSEC records in case it was enabled originally,
+# as `post_records_change` cannot do it because public and
+# private keys were restored afterwards
```
jcameron, line 2548: What do you mean by "restored afterwards" ? post_records_change is called right at the end of the restore_dns function, after the keys have been copied into place.

iliajie: Yes, right. It is called at the very end, but it would fail to sign the zone (i.e. with `sign_dnssec_zone_if_key`), as we call `enable_domain_dnssec`, and only _then_ we overwrite public and private keys from the back up (which have different id) and delete the key with the expected reference number (just generated using `enable_domain_dnssec`). The failing part is in eval, and you cannot see the error, but the error is expected:

```
DNSSEC signing after records change failed : dnssec-signzone: warning: dns_dnssec_keylistfromrdataset: error reading ./Kubuntu22-pro-3.virtualmin.dev.+008+49576.private: file not found
```

So, we either should disable and then again re-enable DNSSEC to update the records using the new key or update records on the zone manually to include the reference id for the new key, or rename the keys that are restored. Doing latter two could be risky and bring new potential unforeseen bugs, so I went down with a simple, safe and working solution to just toggle DNSSEC for the domain 🙂 .

feature-dns.pl
```diff
@@ -2411,13 +2411,26 @@ sub restore_dns
...
+			&$second_print($text{'setup_failed'});
+			}
+		else {
+			&$second_print($text{'setup_done'});
+			}
+		&$outdent_print();
+
+		# As lock was removed while enabling
+		# domain DNSSEC, re-obtain it.
+		&obtain_lock_dns($d);
```
jcameron, line 2426: The lock functions keep a reference count, so there's no need to re-lock here, because enable_domain_dnssec won't actually release the lock.

feature-dns.pl
jcameron: Ok I see now ... damn, dnssec is complicated!

iliajie: Yes, it is very confusing!

feature-dns.pl
iliajie: Oh, great to know! I will remove it then.

iliajie:
Thanks for merging it.

feature-dns.pl
iliajie: You already did it. Thank you!

PR #498, opened 2023-01-01
Add PHP-FPM pool limits support

iliajie:


edit_phpmode.cgi
```diff
@@ -45,12 +45,15 @@ if ($can) {
 	    ($p eq 'web' || &plugin_defined($p, "feature_get_web_php_children"))) {
 		$children = &get_domain_php_children($d);
 		if (defined($children) && $children >= 0) {
-			print &ui_table_row(&hlink($text{'phpmode_children'},
-						   "phpmode_children"),
-				    &ui_opt_textbox("children", $children > 0 ? $children : '', 5,
-				        $mode eq 'fcgid' ?
-				        $text{'tmpl_phpchildrenauto'} : 
-				        &text('tmpl_phpchildrennone', &get_php_max_childred_allowed())));
+			my $dom_limits = &get_domain_resource_limits($d);
```
jcameron, line 48: This will fail on GPL systems - the get_domain_resource_limits function is defined only in pro/resources-lib.pl

iliajie: Oh, you're right. Thanks! I forgot about this test! Fixed now.

save_phpmode.cgi
```diff
@@ -182,6 +160,42 @@ if (&can_php_error_log($mode)) {
...
+	    $nc != &get_domain_php_children($d) && $can) {
+		&$first_print($nc || $mode eq "fpm" ?
+		    &text('phpmode_kidding', $nc || &get_php_max_childred_allowed()) :
+		    $text{'phpmode_nokids'});
+		&save_domain_php_children($d, $nc);
+		&$second_print($text{'setup_done'});
+		$anything++;
+		}
+	# Save with imposed limits
+	elsif($can && $mode eq "fpm" && $dom_limits->{'procs'}) {
```
jcameron, line 176: Why do we need this block when in edit_phpmode.cgi we only show the child process limit if `$dom_limits->{'procs'}` isn't set?

iliajie: We need this to re-impose the limits on PHP change.

jcameron: That seems like something we should set elsewhere, because there are plenty of places where the PHP version can be changed. So I think this code should be removed, and a separate PR created to re-apply limits to new PHP versions.

iliajie: Alright, done!

PR #511, opened 2023-01-25
Add support for enabling/disabling password hashing for existing servers

iliajie:
Jamie, I have taken a deeper look into adding support for switching password hash mode for existing domains. As we [discussed here](https://forum.virtualmin.com/t/i-cant-see-mail-paswords-in-edit-user-page-after-virtualmin-authentic-theme-update/118831/17?u=ilia), if template and domain `hashpass` options don't match then we will show a nice checkbox and let user change `hashpass` mode.

Heavily tested and working well for me, yet still a PR.

image

As you can see there is a new `Hash password` checkbox appears but only of `hashpass` value on the template doesn't equal `hashpass` value on the domain config. In all other, standard cases it won't be shown.

### DIFF
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -10754,7 +10754,7 @@ my ($d, $mode) = @_;
 if (&check_domain_hashpass($d)) {
 	$d->{'hashpass'} = $mode;
 	if (!$mode) {
-		my @encpass_opts = grep { $_ =~ /enc_pass$/ } keys %{$d};
+		my @encpass_opts = grep { /enc_pass$/ } keys %{$d};
 		foreach my $encpass_opt (@encpass_opts) {
 			delete($d->{$encpass_opt});
 			}
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -10763,14 +10763,12 @@ my ($d, $mode, $pass) = @_;
 if ($mode) {
 	# On switching to hashed mode some passwords
 	# services need to be reset to plain text	
-	my @opts = ('mysql_enc_pass');
-	foreach my $opt (@opts) {
-		if ($d->{$opt}) {
-			my $fn = $opt;
-			$fn =~ s/enc_//;
-			$d->{$fn} = $pass;
-			delete($d->{$opt});
-			}
+	my $opt = 'mysql_enc_pass';
+	if ($d->{$opt}) {
+		my $fn = $opt;
+		$fn =~ s/enc_//;
+		$d->{$fn} = $pass;
+		delete($d->{$opt});
 		}
 	}
 }
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -10729,13 +10729,20 @@ sub check_domain_hashpass
 {
 my ($d) = @_;
 my $tmpl = &get_template($d->{'template'});
-my $rs = "";
 if (defined($d->{'hashpass'}) && defined($tmpl->{'hashpass'}) &&
     $d->{'hashpass'} != $tmpl->{'hashpass'}) {
-	$rs = "&nbsp;&nbsp;&nbsp;&nbsp;".&ui_checkbox("hashpass_enable", 1, 
-		$text{'edit_hash'}, $d->{'hashpass'});
+	return 1;
 	}
-return $rs;
+return 0;
+}
+
+# check_domain_hashpass(&domain)
+# Returns hashpass checkbox
+sub get_domain_hashpass_checkbox
+{
+my ($d) = @_;
+return "&nbsp;&nbsp;&nbsp;&nbsp;".&ui_checkbox("hashpass_enable", 1, 
+		$text{'edit_hash'}, $d->{'hashpass'});
 }
 
 # update_domain_hashpass(&dom, mode)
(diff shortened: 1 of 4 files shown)


iliajie:
Updated title.

virtual-server-lib-funcs.pl
```diff
@@ -10722,6 +10722,61 @@ sub get_database_host
...
+
+# post_update_domain_hashpass(&@doms, mode, pass)
+# Apply domain services hashpass clean ups
+sub post_update_domain_hashpass
+{
+my ($d, $mode, $pass) = @_;
+if ($mode) {
+	# On switching to hashed mode some passwords
+	# services need to be reset to plain text	
+	my @opts = ('mysql_enc_pass');
```
jcameron, line 10768: Won't this copy the hashed MySQL password to the plaintext password?

iliajie: No, this will remove previously created (i.e. by `generate_domain_password_hashes()`) MySQL hashed password and set one defined by user, as in:

https://github.com/virtualmin/virtualmin-gpl/blob/master/feature-mysql.pl#L490-L495

jcameron: Why the loop when there's only a single element in the array? I think this block could be a lot simpler..

virtual-server-lib-funcs.pl
```diff
@@ -10722,6 +10722,61 @@ sub get_database_host
...
+{
+my ($d) = @_;
+my $tmpl = &get_template($d->{'template'});
+my $rs = "";
+if (defined($d->{'hashpass'}) && defined($tmpl->{'hashpass'}) &&
+    $d->{'hashpass'} != $tmpl->{'hashpass'}) {
+	$rs = "&nbsp;&nbsp;&nbsp;&nbsp;".&ui_checkbox("hashpass_enable", 1, 
+		$text{'edit_hash'}, $d->{'hashpass'});
+	}
+return $rs;
```
jcameron, line 10738: This function should just return 1 or 0 if enabling hashed passwords is possible, and the actual checkbox should be created in the .cgi script.

virtual-server-lib-funcs.pl
```diff
@@ -10722,6 +10722,61 @@ sub get_database_host
...
+# update_domain_hashpass(&dom, mode)
+# Updates domain hashpass option if needed, and
+# delete all *_enc_pass if disabling hashing
+sub update_domain_hashpass
+{
+my ($d, $mode) = @_;
+if (&check_domain_hashpass($d)) {
+	$d->{'hashpass'} = $mode;
+	if (!$mode) {
+		my @encpass_opts = grep { $_ =~ /enc_pass$/ } keys %{$d};
```
jcameron, line 10750: `$_ =~`  isn't needed

virtual-server-lib-funcs.pl
iliajie: At first I thought that `postgres_enc_pass` should also go there. By the way, should it?

virtual-server-lib-funcs.pl
iliajie: Yeah, you're right. It would be more correct. Fixed now.

Although, I used the sub (now and then) for returning the checkbox UI element because I still think we will re-use this code in `edit_user.cgi` page..

virtual-server-lib-funcs.pl
iliajie: Alright, fixed too. I just wanted to be more explicit.

virtual-server-lib-funcs.pl
jcameron: No, there's no such field

virtual-server-lib-funcs.pl
jcameron: A sub like this should never return a UI element ... just create it in the CGI script.

iliajie: Alright, fixed too. But why not?

virtual-server-lib-funcs.pl
iliajie: > No, there's no such field

Alright, fixed too. By the way, is this a [bug](https://github.com/virtualmin/virtualmin-gpl/blob/master/feature-postgres.pl#L136)?

virtual-server-lib-funcs.pl
jcameron: It's just cleaner coding practice to separate UI from logic.

virtual-server-lib-funcs.pl
jcameron: Yes, I'll fix that

virtual-server-lib-funcs.pl
```diff
@@ -10722,6 +10722,57 @@ sub get_database_host
...
+# delete all *_enc_pass if disabling hashing
+sub update_domain_hashpass
+{
+my ($d, $mode) = @_;
+if (&check_domain_hashpass($d)) {
+	$d->{'hashpass'} = $mode;
+	if (!$mode) {
+		my @encpass_opts = grep { /enc_pass$/ } keys %{$d};
+		foreach my $encpass_opt (@encpass_opts) {
+			delete($d->{$encpass_opt});
```
jcameron, line 10750: Don't we set these fields for all domains at creation time, even when not using hashed passwords?

iliajie: This code applies only to existing domains. It is important to remove all `enc_pass` from domain config, if user switches to plain-text passwords.

jcameron: Ok I see ... I expect most of the time this will do nothing because `generate_domain_password_hashes` is called just afterwards, but it can't hurt.

PR #515, opened 2023-02-17
Fix to check if escaped alias user exists

iliajie:
This PR to address this issued described in details here https://github.com/virtualmin/virtualmin-gpl/commit/fcd7ed451a4a5c7afa893f166b8634b4bac30c2f#commitcomment-100894160.

### DIFF
--- a/change-password.pl
+++ b/change-password.pl
@@ -59,7 +59,7 @@ $d = &get_user_domain($username);
 $d || &error_exit("Not a Virtualmin user");
 @users = &list_domain_users($d);
 ($user) = grep { $_->{'user'} eq $username ||
-		 &replace_atsign_if_exists($_->{'user'}) eq $username } @users;
+		 &replace_atsign($_->{'user'}) eq $username } @users;
 $user || &error_exit("Not a Virtualmin user in $d->{'dom'}");
 $olduser = { %$user };
 
--- a/change-password.pl
+++ b/change-password.pl
@@ -59,7 +59,7 @@ $d = &get_user_domain($username);
 $d || &error_exit("Not a Virtualmin user");
 @users = &list_domain_users($d);
 ($user) = grep { $_->{'user'} eq $username ||
-		 &replace_atsign($_->{'user'}) eq $username } @users;
+		 &replace_atsign_if_exists($_->{'user'}) eq $username } @users;
 $user || &error_exit("Not a Virtualmin user in $d->{'dom'}");
 $olduser = { %$user };
 
--- a/feature-mail.pl
+++ b/feature-mail.pl
@@ -2559,7 +2559,7 @@ if (&foreign_check("dovecot")) {
 		}
 	foreach my $dove (@doves) {
 		&unlink_file($dove."/".$_[0]->{'user'});
-		&unlink_file($dove."/".&replace_atsign_if_exists($_[0]->{'user'}));
+		&unlink_file($dove."/".&replace_atsign($_[0]->{'user'}));
 		}
 	}
 }
@@ -2606,8 +2606,8 @@ if (&foreign_check("dovecot")) {
 	foreach my $dove (@doves) {
 		&rename_file($dove."/".$_[1]->{'user'},
 			     $dove."/".$_[0]->{'user'});
-		&rename_file($dove."/".&replace_atsign_if_exists($_[1]->{'user'}),
-			     $dove."/".&replace_atsign_if_exists($_[0]->{'user'}));
+		&rename_file($dove."/".&replace_atsign($_[1]->{'user'}),
+			     $dove."/".&replace_atsign($_[0]->{'user'}));
 		}
 	}
 }
--- a/feature-mail.pl
+++ b/feature-mail.pl
@@ -153,7 +153,7 @@ if ($config{'mail_system'} != 4) {
 		if ($config{'mail_system'} == 0 && $u->{'user'} =~ /\@/) {
 			# Special case for Postfix @ users
 			$foruser{$pop3."\@".$d->{'dom'}} =
-				&replace_atsign($u->{'user'});
+				&replace_atsign_if_exists($u->{'user'});
 			}
 		}
 	if ($d->{'mail'}) {
@@ -2316,7 +2316,7 @@ elsif ($config{'mail_system'} == 0) {
 			# link from the @ so that the mail server and Webmin
 			# agree.
 			$mfreal = &postfix::postfix_mail_file(
-					&replace_atsign($user->{'user'}));
+					&replace_atsign_if_exists($user->{'user'}));
 			}
 		}
 	elsif ($s == 2) {
@@ -2502,7 +2502,7 @@ local $noat;
 if ($config{'mail_system'} == 0 && $_[0]->{'user'} =~ /\@/) {
 	# Remove real file as well as link, if any
 	local $fakeuser = { %{$_[0]} };
-	$fakeuser->{'user'} = $noat = &replace_atsign($_[0]->{'user'});
+	$fakeuser->{'user'} = $noat = &replace_atsign_if_exists($_[0]->{'user'});
 	local ($realumf, $realtype) = &user_mail_file($fakeuser);
 	if ($realumf ne $umf && $realtype == 0) {
 		&system_logged("rm -f ".quotemeta($realumf));
@@ -2559,7 +2559,7 @@ if (&foreign_check("dovecot")) {
 		}
 	foreach my $dove (@doves) {
 		&unlink_file($dove."/".$_[0]->{'user'});
-		&unlink_file($dove."/".&replace_atsign($_[0]->{'user'}));
+		&unlink_file($dove."/".&replace_atsign_if_exists($_[0]->{'user'}));
 		}
 	}
 }
@@ -2606,8 +2606,8 @@ if (&foreign_check("dovecot")) {
 	foreach my $dove (@doves) {
 		&rename_file($dove."/".$_[1]->{'user'},
 			     $dove."/".$_[0]->{'user'});
-		&rename_file($dove."/".&replace_atsign($_[1]->{'user'}),
-			     $dove."/".&replace_atsign($_[0]->{'user'}));
+		&rename_file($dove."/".&replace_atsign_if_exists($_[1]->{'user'}),
+			     $dove."/".&replace_atsign_if_exists($_[0]->{'user'}));
 		}
 	}
 }
@@ -3169,7 +3169,7 @@ if (&foreign_check("dovecot") && &foreign_installed("dovecot")) {
 			if (-e "$control/$u->{'user'}") {
 				push(@names, $u->{'user'});
 				}
-			local $repl = &replace_atsign($u->{'user'});
+			local $repl = &replace_atsign_if_exists($u->{'user'});
 			if ($repl ne $u->{'user'} && -e "$control/$repl") {
 				push(@names, $repl);
 				}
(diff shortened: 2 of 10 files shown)


virtual-server-lib-funcs.pl
```diff
@@ -9078,18 +9078,27 @@ sub unescape_user
...
 # Replace an @ in a username with -
 sub replace_atsign
 {
-local ($rv) = @_;
-$rv =~ s/\@/-/g;
-return $rv;
+my ($user, $nocheck) = @_;
+my $origuser = $user;
+$user =~ s/\@/-/g;
+if (!getpwnam($user) && !$nocheck) {
```
jcameron, line 9097: No no no, we can't do this ... because this function might be called before the other format user is created.

virtual-server-lib-funcs.pl
```diff
@@ -1361,7 +1361,7 @@ sub create_user
 		}
 	if ($needextra) {
 		$extrauser = { %{$_[0]} };
-		$extrauser->{'user'} = &replace_atsign($extrauser->{'user'});
+		$extrauser->{'user'} = &replace_atsign($extrauser->{'user'}, 1);
```
jcameron, line 1364: What does this extra 1 parameter do?

virtual-server-lib-funcs.pl
iliajie: > because this function might be called before the other format user is created.

Correct. I am aware of that and this is why there is the second `nocheck` param.

virtual-server-lib-funcs.pl
iliajie: > What does this extra 1 parameter do?

This parameter makes sure that it skips the check and always replaces `@` with `-`.

virtual-server-lib-funcs.pl
iliajie: I have double checked all the code over again but yet found only one place in `create_user` sub where we need to always replace `@` with `-` because this is where a new user is created.

We could create a new sub like, `replace_atsign_with_check` but that would mean that we will have to replace existing `replace_atsign` all over with a new  `replace_atsign_with_check`, while `replace_atsign` would be used only once in `create_user` .. so this is why I have added that second `nocheck` param for `create_user` situation where a new user is created ..

jcameron: I really don't think this is the right place to do that check though - it should be done in callers to the function, or alternative in a new function called something like `replace_atsign_if_exists` .

iliajie: Alright, I can do that. Check this patch https://github.com/virtualmin/virtualmin-gpl/pull/515/commits/01b76af0fca657953f7950dcdee767e0ab8f6b4b.

change-password.pl
```diff
@@ -59,7 +59,7 @@ =head1 change-password.pl
 $d || &error_exit("Not a Virtualmin user");
 @users = &list_domain_users($d);
 ($user) = grep { $_->{'user'} eq $username ||
-		 &replace_atsign($_->{'user'}) eq $username } @users;
+		 &replace_atsign_if_exists($_->{'user'}) eq $username } @users;
```
jcameron, line 62: This seems unnecessary because it checks for the unmodified user anyway..

feature-mail.pl
```diff
@@ -2559,7 +2559,7 @@ sub delete_mail_file
 		}
 	foreach my $dove (@doves) {
 		&unlink_file($dove."/".$_[0]->{'user'});
-		&unlink_file($dove."/".&replace_atsign($_[0]->{'user'}));
+		&unlink_file($dove."/".&replace_atsign_if_exists($_[0]->{'user'}));
```
jcameron, line 2562: This change isn't needed because it deletes the file the unmodified user above

feature-mail.pl
```diff
@@ -2606,8 +2606,8 @@ sub rename_mail_file
 	foreach my $dove (@doves) {
 		&rename_file($dove."/".$_[1]->{'user'},
 			     $dove."/".$_[0]->{'user'});
-		&rename_file($dove."/".&replace_atsign($_[1]->{'user'}),
-			     $dove."/".&replace_atsign($_[0]->{'user'}));
+		&rename_file($dove."/".&replace_atsign_if_exists($_[1]->{'user'}),
```
jcameron, line 2609: Same here, no need for this change

lookup-domain-daemon.pl
```diff
@@ -154,7 +154,7 @@ sub handle_one_request
 	}
 @users = &list_domain_users($d, 0, 1, 0, 1);
 ($user) = grep { $_->{'user'} eq $username ||
-		 &replace_atsign($_->{'user'}) eq $username }
+		 &replace_atsign_if_exists($_->{'user'}) eq $username }
```
jcameron, line 157: This also isn't needed

lookup-domain.pl
```diff
@@ -163,7 +163,7 @@
 # Check if the user is approaching his quota
 @users = &list_domain_users($d, 0, 1, 0, 1);
 ($user) = grep { $_->{'user'} eq $username ||
-		 &replace_atsign($_->{'user'}) eq $username } @users;
+		 &replace_atsign_if_exists($_->{'user'}) eq $username } @users;
```
jcameron, line 166: Same here

virtual-server-lib-funcs.pl
```diff
@@ -773,7 +773,7 @@ sub list_domain_users
 	# Remove users with @ in their names for whom a user with the @ replace
 	# already exists (for Postfix)
 	if ($config{'mail_system'} == 0) {
-		local %umap = map { &replace_atsign($_->{'user'}), $_ }
```
jcameron, line 776: No need for this, as it's just filtering out @-format usernames, so if they don't exist nothing will happen

virtual-server-lib-funcs.pl
```diff
@@ -1220,7 +1220,7 @@ sub list_all_users_quotas
 	else {
 		# Set quotas based on quota commands
 		my $altuser = $u->{'user'} =~ /\@/ ?
-				&replace_atsign($u->{'user'}) :
```
jcameron, line 1223: Not needed .. perhaps instead just check if altuser exists

virtual-server-lib-funcs.pl
```diff
@@ -1633,13 +1633,13 @@ sub modify_user
 	&foreign_call($usermodule, "made_changes");
 
 	if ($config{'mail_system'} == 0 && $_[1]->{'user'} =~ /\@/) {
-		local $esc = &replace_atsign($_[1]->{'user'});
+		local $esc = &replace_atsign_if_exists($_[1]->{'user'});
```
jcameron, line 1636: No needed, as $esc will be empty if the @-format user doesn't exist

virtual-server-lib-funcs.pl
```diff
@@ -1633,13 +1633,13 @@ sub modify_user
...
 		local @allusers = &list_all_users_quotas(1);
 		local ($oldextrauser) = grep { $_->{'user'} eq $esc } @allusers;
 		if ($oldextrauser) {
 			# Found him .. fix up
 			$extrauser = { %{$_[0]} };
-			$extrauser->{'user'} = &replace_atsign($_[0]->{'user'});
+			$extrauser->{'user'} = &replace_atsign_if_exists($_[0]->{'user'});
```
jcameron, line 1642: Same here

virtual-server-lib-funcs.pl
```diff
@@ -9078,18 +9078,35 @@ sub unescape_user
 # Converts a username into a suitable alias name
 sub escape_alias
 {
-local ($escuser) = @_;
+my ($escuser) = @_;
+my $origuser = $escuser;
 $escuser =~ s/\@/-/g;
+if (!getpwnam($escuser)) {
```
jcameron, line 9084: Shouldn't this function also have an if-exists version?

virtual-server-lib-funcs.pl
iliajie: I have fixed all mentioned above to just `replace_atsign` instead of `replace_atsign_if_exists`.

virtual-server-lib-funcs.pl
iliajie: Isn't that what `replace_atsign_if_exists` was doing?

virtual-server-lib-funcs.pl
iliajie: I think not, because `escape_alias` sub checks for an alias that has to exists.

virtual-server-lib-funcs.pl

virtual-server-lib-funcs.pl
```diff
@@ -4762,7 +4762,7 @@ sub alias_type
         @rv = (2, $1);
         }
 elsif ($_[0] =~ /^\\(\S+)$/) {
-	if ($1 eq $_[1] || $1 eq "NEWUSER" || $1 eq &replace_atsign($_[1]) ||
+	if ($1 eq $_[1] || $1 eq "NEWUSER" || $1 eq &replace_atsign_if_exists($_[1]) ||
```
jcameron, line 4765: I don't think this needs to be if_exists either

virtual-server-lib-funcs.pl
```diff
@@ -11707,7 +11724,7 @@ sub get_user_domain
 	local $u;
 	foreach $u (@users) {
 		if ($u->{'user'} eq $_[0] ||
-		    &replace_atsign($u->{'user'}) eq $_[0]) {
+		    &replace_atsign_if_exists($u->{'user'}) eq $_[0]) {
```
jcameron, line 11727: Nor this one

iliajie: Also fixed.

virtual-server-lib-funcs.pl
```diff
@@ -17331,7 +17348,7 @@ sub count_domain_users
 		}
 	if ($config{'mail_system'} == 0) {
 		# Don't double-count Postfix @ and - users
-		local $noat = &replace_atsign($u->{'user'});
+		local $noat = &replace_atsign_if_exists($u->{'user'});
```
jcameron, line 17351: This one doesn't have to change either

feature-mail.pl
```diff
@@ -2316,7 +2316,7 @@ sub create_mail_file
 			# link from the @ so that the mail server and Webmin
 			# agree.
 			$mfreal = &postfix::postfix_mail_file(
-					&replace_atsign($user->{'user'}));
```
jcameron, line 2319: In this case, we should just not create the link if there is no alternate user

virtual-server-lib-funcs.pl
iliajie: > This one doesn't have to change either

Does it mean that the regular `@` user will be counted either way (before this code, right?).

feature-mail.pl
iliajie: Alright fixed here https://github.com/virtualmin/virtualmin-gpl/pull/515/commits/61f50780923415b5bb62a370baa0709ce4db8322. Looks good?

I have run more tests over again and it seems to work.

Although, as I just realized that perhaps a new `replace_atsign_if_exists` sub shouldn't _always_ return escaped user which contains an `@` sign?

jcameron: Well it should return the @-sign user if the user without an @ doesn't exist.

iliajie: Well, I see now. Then what I did before was incorrect. Check this patch https://github.com/virtualmin/virtualmin-gpl/pull/515/commits/73bba5aefd95b8e2511aa68b5bba26a6529e749d. It should also address issues when migrating.

PR #523, opened 2023-03-03
Add PHP log controls improvements

iliajie:
Hello, Jamie!

This PR adds important improvements to how PHP error log file is controlled:

  1. Only master admin can set custom log file path
  2. If master admin has imposed custom log file path users won't be able to change it
  3. Respect `Server Owner Limits / Can edit PHP and website options` option before displaying PHP error log file option
  4. Don't loose error log path when changing PHP versions

### DIFF
--- a/edit_phpmode.cgi
+++ b/edit_phpmode.cgi
@@ -191,14 +191,18 @@ if (&can_php_error_log($mode)) {
 	$defplog = &get_default_php_error_log($d);
 	$mode = !$plog ? 1 :
 		$plog eq $defplog ? 2 : 0;
-	print &ui_table_row(&hlink($text{'phpmode_plog'}, 'phplog'),
-		&ui_radio_table("plog_def", $mode,
-		[ [ 1, $text{'phpmode_noplog'} ],
+	my $opts = [ [ 1, $text{'phpmode_noplog'} ],
 		  [ 2, $text{'phpmode_defplog'},
-		       "<tt>$defplog</tt>" ],
-		  [ 0, $text{'phpmode_fileplog'},
-		    &ui_textbox("plog", $mode == 0 ? $plog : "", 60) ],
-		]));
+		       "<tt>$defplog</tt>" ]
+		];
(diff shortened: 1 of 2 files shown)


jcameron:
What's the danger of allowing a custom PHP error log path? It doesn't seem like it could break anything other than the user's own PHP sites?

jcameron:
Maybe the part to preserve the log path when changing PHP versions can be split out?

iliajie:
> What's the danger of allowing a custom PHP error log path? It doesn't seem like it could break anything other than the user's own PHP sites?

The danger is minimal, because even though a user puts PHP log to write to `/etc/passwd` it will fail to work due to permissions. However, it's a mess. Also, an admin may want to force a specific path for PHP logs and wouldn't want a user to change that.

> Maybe the part to preserve the log path when changing PHP versions can be split out?

Well, if we fundamentally disagree on the fact that an admin should have the full control over where user PHP logs are stored, then I could fix that. But I think, the user shouldn't be able to change the path for storing his/her PHP logs outside of home? Although, the additional work may require though (if you think), to let the user set the PHP to anything inside home dir?

jcameron:
Yeah I think what would be best for domain owners is disallowing custom paths outside the home dir ...

iliajie:
I forgot, where we at with this PR?

jcameron:
I'm still unsure as to why we need to restrict where logs can be written, if they are only written with domain owner permissions?

chris001:
Drawbacks to letting users write logs outside their own `/home`.  A malicious user or script could write an arbitrary malicious script to the php log location outside of home, and possibly exploit that script to achieve root privilege escalation.  They could also scan that php log directory for files and directories with unintended loose permissions and modify, delete, or use them to exploit the system or other user/customer accounts.

jcameron:
Now I think about it, maybe we shouldn't allow domain owners to change the PHP log path at all? We disallow this for Apache access and error logs..

PR #537, opened 2023-03-25
Fix resellers part [GPL]

iliajie:
This PR is dependent/related to another [PR](https://github.com/virtualmin/virtualmin-pro/pull/34) in Pro!

* Fix to properly modifying resellers dependent ACLs on add, remove and modify users (Unix or Webmin)
* Fix to factor out `refresh_reseller_user` to a separate sub

### DIFF
--- a/feature-unix.pl
+++ b/feature-unix.pl
@@ -178,6 +178,7 @@ if (defined(&supports_resource_limits) && $tmpl->{'resources'} ne 'none' &&
 	}
 
 &release_lock_unix($_[0]);
+&setvar('unix-user-created', 1, $module_name);
 return 1;
 }
 
@@ -408,6 +409,9 @@ if (!$d->{'parent'}) {
 	&release_lock_unix($d);
 	&release_lock_cron($d);
 	}
+
+&register_post_action(\&refresh_reseller_user, $d, undef)
+	if (&getvar('webmin-user-deleted', $module_name));
 
 # Update any groups
 &build_denied_ssh_group(undef, $d);
--- a/feature-webmin.pl
+++ b/feature-webmin.pl
@@ -1029,6 +1029,18 @@ local $oldwd = $oldd->{'parent'} ? &get_domain($oldd->{'parent'}) : $oldd;
 if ($wd->{'webmin'}) {
 	&modify_webmin($wd, $oldwd);
 	}
+&refresh_reseller_user($d, $oldd);
+}
+
+# refresh_reseller_user(&domain, [&old-domain])
+# Update resellers
+sub refresh_reseller_user
+{
+local ($d, $oldd) = @_;
+my $has_oldd = $oldd ? 1 : 0;
+$oldd ||= $d;
+local $wd = $d->{'parent'} ? &get_domain($d->{'parent'}) : $d;
+local $oldwd = $oldd->{'parent'} ? &get_domain($oldd->{'parent'}) : $oldd;
 if ($wd->{'reseller'} && $virtualmin_pro) {
 	# Update all resellers on the domain
 	foreach my $r (split(/\s+/, $wd->{'reseller'})) {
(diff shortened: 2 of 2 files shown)


jcameron:
I don't like the user of `unix-user-created` as a global variable that effects other functions ... it makes figuring out the code difficult. Can't we pass this down as a parameter if needed?

Also, why is this needed?

iliajie:
> Also, why is this needed?

It's needed to let know Webmin create user function to register a post action. It's only triggered when a Unix user was created and only after Webmin user creation, i.e. called at the very end.

Upon delete it has to happen in reverse to properly update the ACL file for a reseller.

> Can't we pass this down as a parameter if needed?

I'm not sure at the moment. Can we?

iliajie:
> > Can't we pass this down as a parameter if needed?
> 
> I'm not sure at the moment. Can we?

I don't think we can easily make it work.

jcameron:
Can you explain further what problem this code is fixing though?

iliajie:
> Can you explain further what problem this code is fixing though?

Yeah, this code makes sure that when a reseller is created or deleted, or when domain associated with a reseller is deleted, all configs are synced properly. When we create a domain the Unix administration user is created first and then Webmin user is created, however when deletion happens, mentioned actions happen in reverse (first Webmin user is deleted and only then Unix administration user).

We need to sync resellers accordingly, in particular for `allowed_path` to work in File Manager when and after the process of adding/removing Unix user was finished and domain is saved.

This was the initial idea behind using globals to understand which action happened and when. Although, I have done more tests and we don't really have to check for anything, we can just always sync resellers. I have updated the code in PR not to set globals and just always perform the sync.

iliajie:
> Can you explain further what problem this code is fixing though?

Yeah, this code makes sure that when a reseller is created or deleted, or when domain associated with a reseller is deleted, all configs are synced properly. When we create a domain the Unix administration user is created first and then Webmin user is created, however when deletion happens, mentioned actions happen in reverse (first Webmin user is deleted and only then Unix administration user).

We need to sync resellers accordingly, in particular for `allowed_path` to work in File Manager when and after the process of adding/removing Unix user was finished and domain is saved.

This was the initial idea behind using globals to understand which action happened and when. Although, I have done more tests and we don't really have to check for anything, we can just always sync resellers. I have updated the code in PR not to set globals and just always perform the sync.

iliajie:
> Can you explain further what problem this code is fixing though?

Yeah, this code makes sure that when a reseller is created or deleted, or when domain associated with a reseller is deleted, all configs are synced properly. When we create a domain the Unix administration user is created first and then Webmin user is created, however when deletion happens, mentioned actions happen in reverse (first Webmin user is deleted and only then Unix administration user).

We need to sync resellers accordingly, in particular for `allowed_path` to work in File Manager when and after the process of adding/removing Unix user was finished and domain is saved.

This was the initial idea behind using globals to understand which action happened and when. Although, I have done more tests and we don't really have to check for anything, we can just always sync resellers. I have updated the code in PR not to set globals and just always perform the sync.

jcameron:
So in practice is this only a problem when deleting a domain? I'm trying to understand the sequence of events that would trigger this issue..

iliajie:
Deleting and adding. You can create a reseller with Unix user and switch to that reseller login. Then try to add and delete virtual servers, and onserve for that `reseller.acl` file in `/etc/webmin/filemin` directory. Allowed paths won't be updated correctly with the current code. The PR does it.

jcameron:
I think this is a simpler fix : https://github.com/virtualmin/virtualmin-gpl/commit/b2bbd328912ff32a5a06fa8a40deaf8d30cf2bd3

iliajie:
> I think this is a simpler fix : b2bbd32

Won't it fall into the loop?

Nevertheless, I will only be able to try it tomorrow.

jcameron:
Ok let me know how it goes ...

iliajie:
> Ok let me know how it goes 

Thanks! Excellent! It solves the original issue.

I will close this PR.

PR #550, opened 2023-04-22
Update system_info.pl

josecarre:
Fix unsecured http links in `sysinfo.cgi` for Virtualmin Documentation towards secured https links.


jcameron:
Thanks for the patch!

PR #564, opened 2023-05-17
Add to use a new API to get domain hostname

iliajie:
Related to https://github.com/webmin/webmin/pull/1908 and [unable-to-install-opencart#120408](https://forum.virtualmin.com/t/virtualmin-pro-unable-to-install-opencart-using-the-autoinstaller/120408/31)

The main issue in [unable-to-install-opencart#120408](https://forum.virtualmin.com/t/virtualmin-pro-unable-to-install-opencart-using-the-autoinstaller/120408/31) is actually because a user has a redirect setup from `dom.com` to `www.dom.com`, and this is the actual reason why database setup in that ticket was failing! Before, we were checking  manually only for `www.dom.com`, however technically, a redirect can be setup to any alias, e.g. `www3.dom.com`, so I think this is the way to fix it generally, by simply testing the redirect and making sure it's inside of the main domain.

Please give it a closer look, Jamie. It worked for me in my tests!

### DIFF
--- a/feature-status.pl
+++ b/feature-status.pl
@@ -61,8 +61,7 @@ sub make_monitor
 {
 local ($d, $ssl) = @_;
 local $tmpl = &get_template($d->{'template'});
-local $host = $d->{'dns'} ? "www.".$d->{'dom'}
-			  : &get_domain_http_hostname($d);
+local $host = &get_domain_http_hostname($d);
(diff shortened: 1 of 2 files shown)


tshpr:
I didn't add any redirect, so I suppose it ships with OpenCart's archive. I will double test and let you know.

Issue #568, opened 2023-05-29
Send automatic reply option for user in its virtual domain


gstlouisgit:
Operating system | CentOS Linux 7.9.2009
Webmin version | 2.021 | Usermin version | 1.861
Virtualmin version | 7.7 | Authentic theme version | 20.21

I have had a chance to fiddle with the Send automatic reply for a mailbox using custom Subject:
settings in VM -> edit users -> Mail forwarding settings
Everything looks to be working well 

However, when I removed this setting by simply unchecking the box, I can see that it does not remove the *.txt files it has created for those mailboxes in /home/domain/

so in case you care about that...

jcameron:
I think we chose to do that on purpose so as not to delete users' autoreply messages..

iliajie:
It's a bit tricky but doable. Jamie, check [this PR](https://github.com/virtualmin/virtualmin-gpl/pull/569).

PR #569, opened 2023-05-30
Add ability to always display autoreply if present

iliajie:
Ref.: https://github.com/virtualmin/virtualmin-gpl/issues/568#issuecomment-1568520013


jcameron:
Thanks!

Issue #520, opened 2023-03-01
Mismatch of CLI vs Web version of PHP because CLI version is server-wide


marclaporte:
"Yes, this is possible" but "Note that this command will set default PHP version for the system and not for a specific user or a virtual server." Source: https://forum.virtualmin.com/t/changing-default-version-of-php-cli-for-virtual-host/117171

See also: 

- https://forum.virtualmin.com/t/changing-default-php-cli-version-on-the-server/60103
- https://forum.virtualmin.com/t/setting-cli-php-version-5-6/54449
- https://forum.virtualmin.com/t/change-the-default-php-version-for-command-line-interface/107682
- https://forum.virtualmin.com/t/multiple-php-versions-but-not-adhering-selected-version-in-cli/109177
- https://dev.tiki.org/item7820-Installing-package-is-hardly-possible-on-shared-hosting-without-an-option-to-select-the-PHP-CLI-to-use#threadId9986

Since this causes support requests, we added to [Tiki Check](https://doc.tiki.org/Server-Check)
![php-cli](https://user-images.githubusercontent.com/1004261/222068028-2232a674-a0d3-4fe8-89b9-fe2e1c0b0dad.png)

To address this in ClearOS, here is a wrapper we had made a while back:
https://github.com/WikiSuite/app-php-engines/blob/master/packaging/php_wrapper

Thanks!

iliajie:
Hello, Marc.

Thanks for the heads-up!

> To address this in ClearOS, here is a wrapper we had made a while back

The easier solution would be to create an alias in user's home `.bashrc` file (in case user's shell is set to `bash`).

Jamie, perhaps this is something we could do on Virtualmin side automatically?

jcameron:
I don't think a .bashrc alias will fix the PHP version that a script sees. 

Also, on a system with multiple PHP versions installed, `/usr/bin/php` absolutely can be a different version to the one run via FPM or fCGId.

iliajie:
> I don't think a .bashrc alias will fix the PHP version that a script sees.

Yes, a script would have to call the needed PHP version explicitly.

jcameron:
Virtualmin can generally figure out what the correct CLI PHP command is - I wonder if there's a way we can pass it to Tiki?

iliajie:
Create a symlink to `~/bin/php`?

marclaporte:
> I wonder if there's a way we can pass it to Tiki?

[...]

PR #584, opened 2023-07-11
Create Bash startup profile for PHP alias work on sub-server basis

iliajie:


marclaporte:
For the record: Relates to https://github.com/virtualmin/virtualmin-gpl/issues/520

PR #598, opened 2023-07-26
Fix not to break if Virtualmin plugin cannot be loaded

iliajie:


virtual-server-lib-funcs.pl
```diff
@@ -19257,8 +19257,19 @@ sub load_plugin_libraries
...
+				&foreign_require($pname, "virtual_feature.pl");
+				};
+			if (!$@) {
+				$loaded++;
+				}
+			else {
+				my $err = $@;
+				$err = &html_strip($err);
+				$err =~ s/[\n\r]+/ /g;
+				&miniserv::log_error("Failed to load $pname : $err");
```
jcameron, line 19271: No no, you can't call miniserv functions from within webmin ... what if this code wasn't run via miniserv?

Instead, just print to STDERR

iliajie: Correct! Fixed on the latest commit.

virtual-server-lib-funcs.pl
```diff
@@ -19257,8 +19257,24 @@ sub load_plugin_libraries
...
+				&foreign_require($pname, "virtual_feature.pl");
+				};
+			if (!$@) {
+				$loaded++;
+				}
+			else {
+				my $err = $@;
+				$err = &html_strip($err);
+				$err =~ s/[\n\r]+/ /g;
+				if (defined(&error_stderr)) {
```
jcameron, line 19271: Wait where this this declared? I've not seen this error_stderr function before..

iliajie: It's new. I [just added](https://github.com/webmin/webmin/commit/8ef754b47754973944492444f981d9a913441957) it.

jcameron: Ok makes sense!

PR #615, opened 2023-08-14
Add API to handle host default domain

iliajie:
Jamie, hey! As we discussed earlier with Joe, we needed host default domain to be handled in less confusing way.

This change adds API to control host default domain in versatile way. We will be able to use this API from Virtualmin Config in install script to try to create a domain with Let's Encrypt certificate that would match the hostname. On success a user will be able to login to Virtualmin without any errors in the browser. If request fails we simply remove this domain in place.

Now, this change will also consider the old default domain created by Virtualmin from the post-install wizard page. That old domain will always be visible to the user, and will never be touched automatically by Virtualmin disregard of a new config option.

This change adds a new Virtualmin config option, under *SSL settings* page, i.e.:

image

This option will always be disabled by default. It will also be disabled if Virtualmin Config won't be able to request Let's Encrypt SSL certificate at installation time.

However, if a user will turn this option on at existing system _manually_, here is what's going to happen.

In config check, it will check for the host default domain and will try to enable it, e.g.:

image

If the SSL request fails, the default domain will be instantly discarded.

If SSL sharing in Virtualmin config is enabled even when domain owner doesn't match, then it may create host default domain with SSL sharing, e.g.:

image

Later, if the host default domain was successfully setup, Virtualmin will take care of SSL renewals in the future. However, if the hostname was changed, the host default domain will be automatically regenerated on config check, i.e.:

image

If later an admin wants to disable host default domain in Virtualmin configuration, it can easily be done and on the next config check, the default domain will be completely deleted, e.g.:

image

Lastly, if existing user with an old default domain tries to enable this option, the special case message will be dispalyed:

image

---

Even though the _new_ host default domain will be visible in Virtualmin CLI, it will always be hidden in the UI! It won't be shown in the virtual server select dropdown, in the list of virtual servers page, it cannot be backed up or deleted using UI directly. Also all stats for this domain will be excluded, e.g. on the dashboard, like domain counts, quotas, and etc. Moreover, all features for this domain will be disabled. Mail, databases, PHP and all other will be disabled. All additional DNS records, like _ftp._ or _mail._ will be disabled. Web aliases to admin and webmail won't be created, user shell will be set to `/dev/null`, and all this despite of the template settings.

Nevertheless, the host default domain LE SSL certificate will be added to Postfix SNI map, Apache and Webmin but not Dovecot and MySQL.

Let me know if you have other questions ..


jcameron:
So do we really need to change the list_domains function to hide the default domain? It seems like list_visible_domains is the place to hide it, as that controls what appears in the UI..

iliajie:
> So do we really need to change the list_domains function to hide the default domain?

Have you reviewed my latest code, which is different from the initial, that I made before submitting the PR?

So, yes. We need to apply the filter inside of `list_domains`. `list_domains_filters` will handle it but in a way that can be controlled by `list_domains_excludes`, and without any hardcoded key values.

>  It seems like list_visible_domains is the place to hide it, as that controls what appears in the UI..

It happens in `list_visible_domains` as well. The problem is that we don't have a single general function that would work. There are many different places where `list_domains` is called in different ways. For example, default domain must be excluded from any counts, quotas and etc (shown in the dashboard), from backups, script installs and some other pages.

jcameron:
But is there a case where a caller of just `list_domains` would want to apply a filter? Since this just controls visibility, isn't it better to do the filtering inside of `list_visible_domains` ?

iliajie:
> But is there a case where a caller of just `list_domains` would want to apply a filter?

Yes, of course. All these _.cgi_ [files](https://github.com/virtualmin/virtualmin-gpl/pull/615/files) on the PR.

Also, `collect-lib.pl`, `system_info.pl` and other.

jcameron:
Instead of calling list_domains(list_filters()) as in this PR, why not just change those calls in CGI scripts to list_visible_domains? The purpose of list_visible_domains is to control what appears in the UI ...

iliajie:
> Instead of calling list_domains(list_filters()) as in this PR, why not just change those calls in CGI scripts to list_visible_domains? The purpose of list_visible_domains is to control what appears in the UI

Yes, it's much cleaner solution! Fixed here https://github.com/virtualmin/virtualmin-gpl/pull/615/commits/3b6933f7891aeae933d0d1ed3eed90debe80f065.

system_info.pl
```diff
@@ -15,7 +15,7 @@ sub list_system_info
 my @rv;
 my $info = &get_collected_info();
 my @vposs = $info ? @{$info->{'vposs'}} : ( );
-my @doms = &list_visible_domains();
+my @doms = &list_visible_domains({ 'exclude-alias' => 1 });
```
jcameron, line 18: Why exclude aliases?

webmin_menu.pl
```diff
@@ -41,8 +41,8 @@ sub list_webmin_menu
 push(@rv, { 'type' => 'hr' });
 
 # Get domains and find the default
-my @alldoms = &list_domains();
-my @doms = &list_visible_domains();
+my @alldoms = &list_visible_domains();
+my @doms = &list_visible_domains({ 'exclude-alias' => 1 });
```
jcameron, line 45: Also here, why exclude aliases? Won't this hide them from the left menu?

feature-mail.pl
```diff
@@ -5375,7 +5375,7 @@ sub count_domain_aliases
 
 # Map each virtuser to a domain, except for those owned by plugins or for
 # which the destination is a user
-local %dmap = map { $_->{'dom'}, $_ } &list_domains();
+local %dmap = map { $_->{'dom'}, $_ } &list_visible_domains();
```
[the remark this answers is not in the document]
iliajie: > This should include all domains

I thought it should because **Virtualmin Count** on the dashboard wouldn't be displayed correctly, if host default domain is set to be hidden. But I ran more tests and it appears that you're right. It works just with `list_domains` domains.

webmin_menu.pl
iliajie: We need to use a flag, like `{ 'exclude-alias' => 1 }` to preserve initial logic, as before `list_visible_domains` was used only in two places, and applied the logic to exclude the alias domains. But now we use `list_visible_domains` in so many other new cases where old alias related exclusion is not meant to be used. So I use `{ 'exclude-alias' => 1 }` in those old two places where `list_visible_domains` was use initially.

jcameron: Oh there was a config option to hide aliases before! That seems odd ... maybe we should just drop that config setting, because I don't know what use it is.

iliajie: > Oh there was a config option to hide aliases before! That seems odd ... maybe we should just drop that config setting, because I don't know what use it is.

I don't know. I will leave it up to you! I just made sure that the old logic didn't change.

jcameron: I think we should make `list_visible_domains` return all domains the user has access to, and delete that `hide_alias` config option.

iliajie: > I think we should make `list_visible_domains` return all domains the user has access to, and delete that `hide_alias` config option.

I don't think we should remove it. There is a Virtualmin config option called **Include alias virtual servers in menu?** under **Server administrator permissions** page. It may be a feature that people use? That said, if we remove that, we should remove it from Virtualmin config as well.

> However, we need to be careful to only call list_visible_domains to control what appears in the UI.

Yeah, I think we do already!

jcameron: Yes, we should also remove it from the module config page. Since this PR expands the user of list_visible_domains, there are cases where we don't want aliases to be hidden..

iliajie: I can remove it as I get to my computer..

jcameron: Ok great!

iliajie: > Yes, we should also remove it from the module config page. Since this PR expands the user of list_visible_domains, there are cases where we don't want aliases to be hidden..

Dropped in this commit https://github.com/virtualmin/virtualmin-gpl/pull/615/commits/fd5038b9f179abf0dca6be72578b968721a9b83b.

jcameron: Excellent thanks!

jcameron:
Is this PR ready for final review and merge now?

iliajie:
> Is this PR ready for final review and merge now?

Yes!

iliajie:
Thanks for reviewing and merging it. Although, I have a question. Don't we need to make rename function setup per-ip SSL certificates automatically?

jcameron:
Maybe ... Which rename functions, and why would they need to do this?

iliajie:
> Maybe ... Which rename functions, and why would they need to do this?

Like I do it here "manually", and `rename_virtual_server` doesn't take care of it automatically, e.g.:

https://github.com/virtualmin/virtualmin-gpl/pull/615/commits/003ceb09c561b96d7c5a67467aa927138531a1e7#diff-a8b0a88367118c8f13666eed07287b13ecc217c8fdfaf6a34d51b226dc196f92R19702-R19711

Is this expected?

jcameron:
This should already happen via code in `modify_ssl` which re-requests a Let's Encrypt cert, which in turn will update all service certs.

iliajie:
I'm not sure if that really happened for me. I'll have to double check!

PR #623, opened 2023-08-18
Fix to use a new Cloudflare post validate API

iliajie:
Related: https://github.com/virtualmin/virtualmin-pro/pull/38/files

Ref.: virtualmin/virtualmin-pro@793e5dc#commitcomment-124607039


iliajie:
Also wrong!

PR #624, opened 2023-08-18
Fix to use a new Cloudflare post validate API

iliajie:
Related: https://github.com/virtualmin/virtualmin-pro/pull/38/files

Ref.: virtualmin/virtualmin-pro@793e5dc#commitcomment-124607039

### DIFF
--- a/feature-dns.pl
+++ b/feature-dns.pl
@@ -121,6 +121,10 @@ elsif ($d->{'dns_cloud'} && !$dnsparent) {
 		}
 	my $cfunc = "dnscloud_".$ctype."_create_domain";
 	my ($ok, $msg, $location, $err2) = &$cfunc($d, $info);
+	my $efunc = "dnscloud_".$ctype."_post_validate";
+	($ok, $msg, $location, $err2) =
+		&$efunc($d, $ok, $msg, $location, $err2)
+			if (defined(&$efunc));
 	if ($ok == 0) {
 		&$second_print(&text('setup_ebind_cloud', $msg));
 		return 0;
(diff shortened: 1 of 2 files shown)


feature-dns.pl
```diff
@@ -121,6 +121,10 @@ sub setup_dns
 		}
 	my $cfunc = "dnscloud_".$ctype."_create_domain";
 	my ($ok, $msg, $location, $err2) = &$cfunc($d, $info);
+	my $efunc = "dnscloud_".$ctype."_post_validate";
```
jcameron, line 124: Should this still be called if create_domain failed?

feature-dns.pl
```diff
@@ -815,6 +819,8 @@ sub modify_dns
 	else {
 		my $rfunc = "dnscloud_".$ctype."_rename_domain";
 		my ($ok, $id) = &$rfunc($d, $info);
+		my $efunc = "dnscloud_".$ctype."_post_validate";
```
jcameron, line 822: Should this still be called if rename_domain failed?

feature-dns.pl
iliajie: I don't think so, as we can use it in many other cases to handle other errors, right?

feature-dns.pl
iliajie: > Should this still be called if rename_domain failed?

No, as the error messages and codes are the same. No need to duplicate it.

feature-dns.pl
jcameron: Oh I see ... never mind

PR #651, opened 2023-09-24
Add domain config saver

iliajie:
As reported [here](https://avan.tech/item85186-Virtualmin-Make-it-obvious-to-find-database-credentials-when-we-are-in-a-sub-server#threadId=67530):

> I believe it would be a shame to leave things as they are, as it could create confusion for uninformed users. As noted by Jeff Fortin and as you can see in the screenshots shared by Marc, many users find it confusing and incomprehensible to not find this information in the "Edit Databases" tab even if they are on a sub-server. An informed user will know directly that he must go through a top-level server, but for others it is not easy to guess this.

I made a few changes <sup>**[1](https://github.com/virtualmin/virtualmin-gpl/commit/61b20eefeda9ec7ca51980a2f658948d149bbb84) [2](https://github.com/virtualmin/virtualmin-gpl/commit/8a3a6701c9cf26a625cd6f52cbbe0916a431f83e)**</sup> to display an explanation in UI, why sub-server cannot edit database username and password, e.g.:

image

However, I would like to submit this new `save_domain_config.cgi` for your review, because even though it does all the checks, it still makes it possible to change any options in the domain config.

It may be pretty convenient to have such API, although I'd like to double check if you think it is safe to change any domain config options in case domain owner can edit given domain?


chris001:
Very nice!  My 2 cents:  to make the text in the Information box, "its parent server settings", be a link to the parent server's settings page ("Edit virtual server"), `/virtual-server/edit_domain.cgi?dom=xxxxxxxxxxxxxxxxxxx&xnavigation=1`.  Bonus: to print the name of the parent server in the Information message, `alma9-pro.vrtualmin.dev`.  Although the parent server could be any domain name... e.g. `unrelated.virtualmin.dev`... that's why it's good to print it in the Info message, to remind the user what's their parent server name.

iliajie:
> that's why it's good to print it in the Info message, to remind the user what's their parent server name.

Alright, agreed, something like this would work better:

image

jcameron:
How would this `save_domain_config.cgi` script be called? Also, it's unclear to me what it's supposed to do ...

iliajie:
> How would this `save_domain_config.cgi` script be called? Also, it's unclear to me what it's supposed to do .

It will be called from here https://github.com/virtualmin/virtualmin-gpl/commit/8a3a6701c9cf26a625cd6f52cbbe0916a431f83e#diff-006a8bad9686960c91e741eef63895c5a2c5e0dba265c427ba5611b7f1ec95e3R21-R30.

jcameron:
So it's just supposed to hide that warning? I actually think it would be be better to just have a small link to the parent domain's page that's always visible, and remove the complexity of adding this new CGI script.

iliajie:
The link was already added in this patch https://github.com/virtualmin/virtualmin-gpl/commit/89f0cd41f2ab4e097cba7f4ec9a0a55e2079d3ff.

And I think it would be pretty irritating to have it displayed all the time.

jcameron:
I'm thinking it doesn't have to be a large warning message, just a small link. The current warning seems more likely to scare users, because it looks like an error when in fact nothing is wrong.

Also, separately from that, this CGI script is too powerful as it can modify delete arbitrary fields from a domain's config file based on URL params. A malicious domain owner could probably use it to grant themselves extra permissions, for example by removing the `uid` field!

iliajie:
> Also, separately from that, this CGI script is too powerful as it can modify delete arbitrary fields from a domain's config file based on URL params.

Yes, exactly! That was my initial concern!

> A malicious domain owner could probably use it to grant themselves extra permissions, for example by removing the uid field!

Alright, I thought about it too however, I'm not sure what would that lead to? `uid` being set to `0`?

Anyway, I can lock its power to do specific change only.

> I'm thinking it doesn't have to be a large warning message, just a small link. 

But this is informational message in blue, I'd say it is not meant to represent any errors. We use it everywhere across UI.

Unless you insist, I'd rather limited the power of save domain config script (or even removed it in favor of having hard-coded field to be saved) .. ?

jcameron:
Yes, please change this script to only dismiss this one specific warning, rather than being general-purpose. Or alternately, do we even need a button to dismiss it? Mostly in Virtualmin these kinds of warnings are always shown ..

iliajie:
I would think so, as from a UI standpoint, it is user-friendly to omit elements that are evident or redundant. I personally,  would want such alert to be dismissible, as it has no further value after it was shown.

Check this change https://github.com/virtualmin/virtualmin-gpl/pull/651/commits/7f94fb93ff3e3297e4932fb86d9009c95bf0b231.

jcameron:
Thanks for the patch!

iliajie:
My pleasure!

PR #654, opened 2023-09-27
Add numerous improvements to Jailkit setup and control

iliajie:
Jamie, this PR adds numerous improvements to Jailkit in Virtualmin.

image

1. Take Jailkit out to a separate section
2. Add ability to clean previously configured jail
3. Add ability to copy extra commands to the jail
4. Add ability to copy extra sections to the jail (aside from global config) 
5. Add nice documentation to each option


edit_limits.cgi
```diff
@@ -168,6 +168,27 @@ if (defined(&list_scripts)) {
 	}
 
 print &ui_hidden_table_end("limits");
+
+# Jailkit section
+if (!&check_jailkit_support()) {
+print &ui_hidden_table_start($text{'limits_header4'}, "width=100%", 2,
```
jcameron, line 174: Identation is missing here

save_limits.cgi
```diff
@@ -91,20 +91,69 @@ if (defined($in{'shell'})) {
...
+			$err = &disable_domain_jailkit($d);
+			&error(&text('limits_ejailoff2', $err)) if ($err);
+			$jail_post_clean_disabled = 1;
+			}
+
+		# Clean all previously copied files
+		my $dom_chdir = &domain_jailkit_dir($d);
+		if (-d $dom_chdir && $dom_chdir =~ /^\/.*?([\d]{6,})$/) {
+			if ($1 eq $d->{'id'}) {
+				eval "use File::Find";
```
jcameron, line 112: Can we do this without using an extra perl module?

save_limits.cgi
```diff
@@ -91,20 +91,69 @@ if (defined($in{'shell'})) {
 if (!&check_jailkit_support()) {
 	my $oldjail = &get_domain_jailkit($d);
 	my $jailupd;
+	my $jail_post_clean_enabled = 0;
+	my $jail_post_clean_disabled = 0;
+	$d->{'jail_esects'} = $in{'jail_esects'};
+	$d->{'jail_ecmds'} = $in{'jail_ecmds'};
+
+	# Clear chroot first
+	if ($in{'jail_clean'}) {
```
jcameron, line 100: This looks like a potentially destructive option ... maybe there should be an extra warning?

edit_limits.cgi
iliajie: I'll fix that.

save_limits.cgi
iliajie: No, we shouldn't do it without Perl module, as it's more robust than doing `readdir`. Also, these modules are embedded in Webmin anyway.

save_limits.cgi
iliajie: Like the second page with confirmation button? Yes, this isn't a bad idea. Although, we explain what this option does in the help file. Also this option is always off by default. Besides, there is not much to destroy, unless user did tons of custom modifications, which in this case would be lost.

Do you really believe that it's needed?

jcameron: Depends ... will all the files and directories copied into the jail be deleted? What if the user added something to one of those directories?

Maybe it's better to just copy over new files?

iliajie: > Depends ... will all the files and directories copied into the jail be deleted? What if the user added something to one of those directories?

A user generally cannot add anything to those directories. Admin can. The user should only be able to make changes in the home directory. All those directories and files copied into jail, i.e. `/usr`, `/lib` and etc.  by `jail_init` should be owned by _root_, and therefore a user should not be able to make changes there anyway.

> Maybe it's better to just copy over new files?

Oh, this is the whole entire reason why I created this feature in the first place. No, we need to clean things up completely, because if an admin adds `zip` as an extra command at first, and then wants to remove it, it won't be possible by just copying new files over.

edit_limits.cgi
iliajie: > Identation is missing here

Fixed.

edit_limits.cgi
```diff
@@ -168,6 +168,27 @@ if (defined(&list_scripts)) {
...
+if (!&check_jailkit_support()) {
+	print &ui_hidden_table_start($text{'limits_header4'}, "width=100%", 2,
+			     "jail", 0, [ "width=30%" ]);
+	# Chroot directory
+	my $jail = &get_domain_jailkit($d);
+	print &ui_table_row(&hlink($text{'limits_jail'}, "limits_jail"),
+		&ui_radio("jail", $jail ? 1 : 0,
+		  [ [ 1, $text{'yes'}.($jail ? " (<tt>$jail</tt>)" : "") ],
+		    [ 0, $text{'no'} ] ]));
+	print &ui_table_row(&hlink($text{'limits_jail_clean'}, "limits_jail_clean"),
```
jcameron, line 182: I thunk that instead of this being a checkbox, it should be an extra Save button at the bottom of the page .. like "Save and Reset Jail". That would be more consistent I think..

iliajie: I would agree to have a separate button but only if we factored out Jailkit into a separate page. Adding such separate button now wouldn't fit. This page isn't just about Jailkit.

Also, I'm thinking now, that we could make it a specific Pro feature (not the whole chroot option). I mean that the Jailkit page with all the new controls from this PR, would be a Pro feature, while GPL version would only have the old style chroot enable/disable control.

If GPL version is installed, we display everything as before and advertise Jailkit as a separate Pro feature (by having grayed out link in the nav menu). When Pro is installed, the link on the menu would be displayed as normal. and the option to enable/disable chroot won't be shown Edit Owner Limits page.

What are your thoughts on it?

save_limits.cgi
jcameron: Ok, makes sense ..

edit_limits.cgi
jcameron: Let's keep this simple for now and just replace that radio button to clear the jail with a separate submit button. It's still fine to do this even when the form has other options on it..

save_limits.cgi
jcameron: Also, what's this regexp on dom_chdir doing?

iliajie: I check if the returned directory actually contains domain id. I do this as an extra pre-causion measure, because we delete files recursively. We need to be 100% certain that directory we operate on is the right one.

It may seems that this is an overhead, because `domain_jailkit_dir` returns a path with domain id. However, `$d->{'id'}` can be empty technically (shouldn't but can), and if it is, then it would delete all chrooted domains and its mounted home directories! It would be disastoroues ..

edit_limits.cgi
iliajie: Jamie, I have taken another (fresh) look at it, and decided to replace the radio with a checkbox, with slightly change title. It must be better now! https://github.com/virtualmin/virtualmin-gpl/pull/654/commits/0b1495f2ada2b80a38114a800c4185fa0fadccbe

I really don't like an extra button, sorry!

Check it out:

image

iliajie: Jamie, did you have a moment to check it again?

jcameron: I suppose I can live with that ... but should the checkbox really be enabled by default?

save_limits.cgi
jcameron: Ok ... but also, what is this code below doing? Does it just delete the entire jail?

edit_limits.cgi
iliajie: > but should the checkbox really be enabled by default?

No, never. It is never meant to be enabled by default.

save_limits.cgi
iliajie: > Ok ... but also, what is this code below doing? Does it just delete the entire jail?

Yes, this code deletes everything in jail but `home` directory. As I mentioned earlier it is done as a precious measure, in case home directory wasn't unmounted properly. I realize it should never happen but better safe than sorry. I believe we can only safely remove everything under `chroot/dom-id` directory if the domain is getting removed (which you already do).

edit_limits.cgi
jcameron: Ok great!

save_limits.cgi
jcameron: Hmm, I feel like that could be done a lot more simply by just listing the chroot directory and calling unlink_file on everything except /home . No need to use File::Path at all..

iliajie: Well, yeah .. but why Webmin doesn't have an API to delete given directory with passable list of exclusions?

iliajie: Check this out, Jamie. This patch https://github.com/virtualmin/virtualmin-gpl/pull/654/commits/5e1bf86b805232d07f4d888c64e6f16292b49bbd will work as you suggested but also delete all empty directories under chroot home, if empty. It may be more useful if an admin user was changed while chroot was disabled.

save_limits.cgi
```diff
@@ -91,20 +91,74 @@ if (defined($in{'shell'})) {
...
+		# Clean all previously copied files
+		my $dom_chdir = &domain_jailkit_dir($d);
+		if (-d $dom_chdir && $dom_chdir =~ /^\/.*?([\d]{6,})$/) {
+			if ($1 eq $d->{'id'}) {
+				opendir(my $dh, $dom_chdir) || &error("Cannot open chroot directory $dom_chdir for cleaning: $!");
+				while (my $dir = readdir($dh)) {
+					next if $dir eq '.' or $dir eq '..';
+					my $fpath = "$dom_chdir/$dir";
+					# Remove home directory only if empty
+					if (&is_under_directory("$dom_chdir/home", $fpath)) {
```
jcameron, line 117: Can't you just say `if ($dir eq "home") {`

save_limits.cgi
```diff
@@ -91,20 +91,74 @@ if (defined($in{'shell'})) {
...
+						opendir(my $dhh, $fpath) || &error("Cannot open chroot home directory $dom_chdir for cleaning: $!");
```
jcameron, line 118: Wouldn't it be simpler to just not touch /home at all?

save_limits.cgi
iliajie: Perhaps, but I tested it with more than one directory being under home.

save_limits.cgi
iliajie: No, because chroot could be disabled domain renamed and then there would be a stray old directory name hanging around if chroot re-enabled.

jcameron: I suppose ... but I would rather have simpler code when dealing with deleting directories than the minor minor issue of an old empty directory being left around.

save_limits.cgi
swelljoe: File::Path has been a core module for a decade or two.

save_limits.cgi
iliajie: > but I would rather have simpler code when dealing with deleting directories than the minor minor issue of an old empty directory being left around.

It is the same logic as original, which I changed instead of using `File::Path` .

I tested it quite well at first and wouldn't want to change anything about this code, unless you see any issue with it.

save_limits.cgi
jcameron: What I mean is, there is zero difference between  `&is_under_directory("$dom_chdir/home", $fpath)` and `$dir eq 'home'` because at this point in the loop it doesn't go into any sub-directory.

save_limits.cgi
jcameron: Let's just keep this super simple and not touch *anything* under /home regardless. That's safe because it won't accidentally delete user home directories, and much simpler because they're no need to check for empty dirs.

save_limits.cgi
iliajie: > What I mean is, there is zero difference between `&is_under_directory("$dom_chdir/home", $fpath)` and `$dir eq 'home'` because at this point in the loop it doesn't go into any sub-directory.

Yes, you're right. Fixed on the latest commit.

save_limits.cgi
iliajie: Well, I could do it but we just use `rmdir` which is _always_ safe! I would like to keep things precise and clean if possible. The code just works.

jcameron: Yes, using `rmdir` would be the best option..

jcameron: Please update this to just not touch the /home directory, and remove the code to check if it's empty.

iliajie: > Please update this to just not touch the /home directory, and remove the code to check if it's empty.

Yes, sure but can you explain why? Where my code could fail exactly?

jcameron: It won't fail, but it can be much simpler while only missing the rare case of an extra empty directory under /home. Simplicity is a virtue!

iliajie: .. ok, fixed!

Issue #658, opened 2023-10-03
new subservers php-fpm chroot directory set to another servers chroot


hydazz:
when creating a subserver (havnt tested it with main servers), by default, the `chroot` option in php-fpm is set to the wrong chroot directory, 

![image](https://github.com/virtualmin/virtualmin-gpl/assets/53986978/fd70e003-a8f9-4f9f-8bea-3a430688625a)

I have not edited `chroot = /home/chroot/1687503619238689` here, thats how it generated, `1687503619238689` is the domain id of a completly different domain, though it got every other option correct?

not the biggest of issues, but caused some headaches when diagnosing missing files

hydazz:
the domain for that subdomain is also not a chroot, so that chroot should of been set to `/home/.../domains/storage....`

iliajie:
> thats how it generated, `1687503619238689` is the domain id of a completly different domain

What different domain .. a parent domain?

hydazz:
> > thats how it generated, `1687503619238689` is the domain id of a completly different domain
> 
> What different domain .. a parent domain?

no, a completly different domain
![image](https://github.com/virtualmin/virtualmin-gpl/assets/53986978/f52348cd-87ac-487c-9a31-c1f895b7b22c)

i created the subdomain `storage.<>.services` and the php-fpm chroot was set to the chroot of the `.io` domain (the selected one here)

the `.services` user is also not chrooted, so it does not make sense to why its php is chrooted to /home/chroot in the first place

hydazz:
its also repeatable, just tested it any subdomain (does not apply to domains) created has their chroot set to `chroot = /home/chroot/1687503619238689` by default

hydazz:
deleting the `.io` domain (1687503619238689), the chroot is now set to another domain/id, which is still not the correct one for that domain

iliajie:
I don't think we should add `chroot` record to FPM pool unless jail is enabled. The patch in the PR above.

hydazz:
Indeed, just realised. But still for the users/domains that are chrooted, mystery why they are getting the wrong chroot path?

iliajie:
> mystery why they are getting the wrong chroot path?

I don't believe I'm experiencing the same issue on my end.

AndyAlvarezDev:
@iliajie I don't know if you remember, but in the forum in private, I had told you about this matter a while ago... now the forum doesn't work otherwise I would look for the link...

AndyAlvarezDev:
[...]

PR #659, opened 2023-10-03
Fix to check if jail is enabled first

iliajie:
Jamie, I don't think we need to create `chroot` in FPM pool unless jail for domain is enabled.

https://github.com/virtualmin/virtualmin-gpl/issues/658#issuecomment-1744540392

### DIFF
--- a/php-lib.pl
+++ b/php-lib.pl
@@ -2420,8 +2420,10 @@ else {
 		}
 	}
 my $parent = $d->{'parent'} ? &get_domain_by($d->{'parent'}) : $d;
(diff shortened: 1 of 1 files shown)


jcameron:
The current code is fine, because if chroot isn't enabled, `get_domain_jailkit` won't return anything.

iliajie:
But I think it does return it, i.e. sets chroot in FPM config.

jcameron:
Huh, you're right .... I'll merge this. Thanks for the fix!

Issue #657, opened 2023-10-02
LE wildcard certificate request fails in domain creation time


iliajie:
Hello, Jamie!

If we enable `letsencrypt_wild` to `1` in Virtualmin Configuration, then LE validation fails upon domain creation time. When wildcard is disabled, then it works just fine (upon domain creation time).

I have tried it with Cloudflare.

However, when I try to request SSL certificate afterwards, it works most of the time. If it doesn't work from the first try, it always works from the second.

Perhaps, we don't add the right `_acme-challenge` record in case of wildcard domain?

jcameron:
Does it fail, and if so what error message are you getting?

Or does it just hang?

iliajie:
> Does it fail, and if so what error message are you getting?

It fails with error message:

```
Certbot failed to authenticate some domains (authenticator: manual). The Certificate Authority reported these problems:
  Domain: sub16.cloudmin.dev
  Type:   unauthorized
  Detail: Incorrect TXT record "vXD8MdaeAfRRq3GyDyvyobd3W0inG06422YGcnwWxvQ" found at _acme-challenge.sub16.cloudmin.dev

```

iliajie:
To be clear I was creating parent domain `sub16.cloudmin.dev` with Cloudflare, and wildcard enabled by default in Virtualmin Configuration page.

jcameron:
Does it work as expected if you request an LE cert after the domain is created?

iliajie:
.. with the second attempt 90% of the time.

jcameron:
Hmm, I wonder if the issue is that a totally new domain registered with Cloudflare isn't resolvable right away?

jcameron:
Yeah but without a wildcard, web-based validation can be used..

iliajie:
Yeah, meaning we're not adding the right acme challenge records with Cloudflare? Also, why does it work from the second attempt when renewing manually?

jcameron:
Maybe there's some lag between when the challenge record is added and when it's visible?

PR #660, opened 2023-10-04
Fix DNS LE wildcard validation

iliajie:
Jamie, this is the fix to addres the issue with wildcard LE SSL certificate request discussed here https://github.com/virtualmin/virtualmin-gpl/issues/657.

It should be combined of two parts. I will submit the second patch a bit later today.

### DIFF
--- a/feature-ssl.pl
+++ b/feature-ssl.pl
@@ -3034,19 +3034,26 @@ if ($chain) {
 &update_caa_record($d);
 }
 
-# update_caa_record(&domain)
+# update_caa_record(&domain, [force-letsencrypt])
 # Update the CAA record for Let's Encrypt if needed
 sub update_caa_record
 {
-my ($d) = @_;
+my ($d, $letsencrypt_cert) = @_;
 &require_bind();
 return undef if (!$d->{'dns'});
 return undef if (!$d->{'dns_cloud'} &&
 		 &compare_version_numbers($bind8::bind_version, "9.9.6") < 0);
 my ($recs, $file) = &get_domain_dns_records_and_file($d);
 my @caa = grep { $_->{'type'} eq 'CAA' } @$recs;
-my $info = &cert_info($d);
-my $lets = &is_letsencrypt_cert($info) ? 1 : 0;
+# At this stage the cert is always self-signed,
+# so we need to force it for Let's Encrypt
+my $lets = $letsencrypt_cert;
+if (!$lets) {
+	my $info = &cert_info($d);
+	$lets = &is_letsencrypt_cert($info) ? 1 : 0;
+	}
+# Need delay for DNS propagation
+my $dns_delay = 0;
 if (!@caa && $lets) {
 	# Need to add a Let's Encrypt record
 	&pre_records_change($d);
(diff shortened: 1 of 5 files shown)


jcameron:
I don't quite follow what's going on with the CAA record in this PR. Are you trying to update it to be based on the LE cert before it is issued?

iliajie:
> Are you trying to update it to be based on the LE cert before it is issued?

Before we request LE certificate the CAA record should exist. This patch address it, as CAA was never added on the initial shot.

jcameron:
But what are we setting the CAA record based on if there's no LE cert yet?

iliajie:
> But what are we setting the CAA record based on if there's no LE cert yet?

It's called from `request_domain_letsencrypt_cert` sub .. and as you request LE first time, it is never set, because there is still self-sign certificate.

jcameron:
Ok so does this PR just add the CAA record even when the cert is self-signed?

iliajie:
> Ok so does this PR just add the CAA record even when the cert is self-signed?

Jamie, yes. This PR will add CAA right before the actual LE request, as initial cert at the time is self-signed, and old code didn't consider it ..

jcameron:
I see ... I wonder, should we have add a CAA record all the time even when the cert is self-signed? Would that ever make things worse?

iliajie:
> I see ... I wonder, should we have add a CAA record all the time even when the cert is self-signed? Would that ever make things worse?

I believe we should add that CAA record everytime LE certificate is about to be requested, but only in case of wildcard LE certificate.

jcameron:
What I meant was, why not add a CAA record for all domains even when the initial cert is self-signed?

iliajie:
Because that record is specific to LE. We should not pollute DNS zone config, I think. I'd rather be precise.

I don't think we should always add it, and I thought my patch worked in expected way?

jcameron:
I thought the CAA record was just a way to publish any SSL certificate via DNS?

iliajie:
> I thought the CAA record was just a way to publish any SSL certificate via DNS?

We publish specific to LE record only!

jcameron:
Ok I see what's happening now - I was confused as to what is in the CAA record (it's just the CA, not the cert itself).

Can you split up this PR to extract the part that always updates the CAA record from the retry part?

iliajie:
> Can you split up this PR to extract the part that always updates the CAA record from the retry part?

Yes, sure https://github.com/virtualmin/virtualmin-gpl/pull/662.

feature-ssl.pl
```diff
@@ -3066,6 +3074,10 @@ sub update_caa_record
 	&post_records_change($d, $recs, $file);
 	&reload_bind_records($d);
 	}
+if ($dns_delay) {
```
jcameron, line 3077: I don't think this is the right place for adding a delay! It should go in request_domain_letsencrypt_cert , if anywhere

iliajie: Don't we always want to have a delay after CAA are added?

jcameron: Not necessarily - it might happen separately from setting up Let's Encrypt

iliajie: > Not necessarily - it might happen separately from setting up Let's Encrypt

Yeah, my code did this only for LE. But I think we can completely drop it for now.

PR #661, opened 2023-10-09
Fix to respect MySQL/MariaDB charset options

iliajie:
Jamie, we should respect charset options defined in domain templates or the module. 

[Ref.](https://forum.virtualmin.com/t/how-to-allow-emojies-to-transfer-during-virtualmin-backup-restore/122936/27?u=ilia)

Also, shouldn't we set a fall back option to `utf8mb4` ..?

### DIFF
--- a/feature-mysql.pl
+++ b/feature-mysql.pl
@@ -757,6 +757,7 @@ local ($d, $oldd) = @_;
 
 # Re-create each DB with a new name
 local %dbmap;
+local $tmpl = &get_template($oldd->{'template'});
 my @dbs = &domain_databases($oldd, [ 'mysql' ]);
 foreach my $db (@dbs) {
 	local $newname = $db->{'name'};
@@ -814,7 +815,7 @@ if (%dbmap) {
 		local $mymod = &require_dom_mysql($oldd);
 		local $err = &foreign_call(
 			$mymod, "backup_database", $oldname, $temp, 0, 1, 0,
-			undef, undef, undef, undef,
+			$tmpl->{'mysql_charset'} || $mysql::config{'charset'}, undef, undef, undef,
 			&mysql_single_transaction($d, $db));
 		if ($err) {
 			&$second_print(&text('clone_mysqlbackup',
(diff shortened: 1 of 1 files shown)


jcameron:
Wouldn't it be better to get the character set from the actual database?

Also, was there a user bug report that this fixes?

swelljoe:
Customer report is here: https://forum.virtualmin.com/t/how-to-allow-emojies-to-transfer-during-virtualmin-backup-restore/122936/1

And, yes, it only makes sense to get it from the database.

jcameron:
Makes me wonder why `mysqldump` doesn't just get the correct character set from the database by default though? We can of course fetch this and pass it to `mysqldump`, but it may be more complex than that ...

iliajie:
Because historical _default_ character set was `latin1` until modern installations switched to use `utf8mb4`.

By the way, initial issue happened to be in CentOS 7, which explains it (almost a decade old).

The change in my PR does `$tmpl->{'mysql_charset'} || $mysql::config{'charset'}` where both aren't set by default, so effectively, unless an admin explicitly sets it, it would still be `undef`. I don't think anything can go wrong here, right?

jcameron:
Wouldn't it be better to get the *actual* character set the database is using with SQL like `SELECT @@character_set_database` ?

iliajie:
Not sure .. Isn't it what `mysqldump` supposed to be doing by default?

jcameron:
Good question! And if it is, why can't it dump emojis properly?

chris001:
This emoji issue came up [here](https://github.com/wp-cli/db-command/issues/144) and their solution looks OK?

jcameron:
Yeah, but I want to figure out a solution thats scalable and won't cause breakage. The worst thing that could happen would be if we changed the character set for the dump only to find that it caused corruption in *other* backups somehow!

chris001:
The problem is, `mysqldump` is dumb and defaults to dumping in `UTF-8` charset, which always breaks user emojis, if any exist (they can exist when `utf8mb4` charset is in use).

jcameron:
Ok, I'll work on fixing this by using the database character set.

If anyone who's seeing this can send or attach a sample dump to this bug report that uses the correct character set but cannot be backed up without the correct flag to `mysqldump`, that would be really useful.

jcameron:
The next Webmin release will add a function to get the actual database character set here : https://github.com/webmin/webmin/commit/384913506fceca274c8b1c2e0262b8f0b8034c92

And Virtualmin will call it here : https://github.com/virtualmin/virtualmin-gpl/commit/08758a8a6010d45cc33f5055fe532448031bcf28

iliajie:
Jamie, I like this precise approach much more than blindly trying default charset!

Thanks!

PR #662, opened 2023-10-10
Add ability to silently retry on Let's Encrypt failure

iliajie:


jcameron:
Wouldn't it be cleaner to use a loop rather than a recursive call?

iliajie:
> Wouldn't it be cleaner to use a loop rather than a recursive call?

Most probably not, as we use `state` quite often and rely on it to work correctly. Are you afraid it may end-up in an infinite loop? I thought of it, but I have never seen `state` fail in any way ..

jcameron:
The code would be way way simpler and easier to read if it just used a loop instead of a `state` variable inside a recursive call.

iliajie:
> Fix to use loop instead of state

Like this https://github.com/virtualmin/virtualmin-gpl/pull/662/commits/ca312041ca2df46cfc5898d8e6a9b72407fb6141 .. ?

jcameron:
ok this looks good to merge, once you fix that merge conflict ...

iliajie:
> ok this looks good to merge, once you fix that merge conflict ...

Done!

PR #664, opened 2023-10-16
Fix detection of external IPv4

iliajie:
This PR address [this](https://forum.virtualmin.com/t/load-the-page-create-virtual-server-takes-forever/123035/17) reported issue.

  1. It will show previously cached IPv4 address if available.
  2. Upon config check external IP address cache will be updated

### DIFF
--- a/domain_form.cgi
+++ b/domain_form.cgi
@@ -669,7 +669,7 @@ else {
 
 # Show DNS IP address field
 if (&can_dnsip()) {
-	my $def_dns_ip = &get_dns_ip($resel);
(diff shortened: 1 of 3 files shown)


jcameron:
I will also add caching to the lookup function : https://github.com/virtualmin/virtualmin-gpl/commit/2c626f0085053bb0b2da67ec968fc57b810a86b5

iliajie:
> I will also add caching to the lookup function : [2c626f0](https://github.com/virtualmin/virtualmin-gpl/commit/2c626f0085053bb0b2da67ec968fc57b810a86b5)

Cool, thanks!

Issue #663, opened 2023-10-14
Navigation menu reorganization


iliajie:
I would like to suggest to reorganize some parts of Virtualmin navigation menu, in particular those parts related to virtual server itself.

I think **Administration Options**, **Server Configuration** and **Services** could be replaced by:

  1. **Control Center**
  2. **DNS Configuration**
  3. **Web Configuration**
  4. **Mail Configuration**

Example of items' details:

 * **Control Center** could include:
    * Change Backups Excludes (curr. Excluded Directories and DBs)
    * Change Domain Name
    * Change IP Address
    * Clone Virtual Server
    * Edit Owner Limits
    * Edit Resource Limits
    * List Sub-Servers
    * Manage Extra Admins
    * Move Virtual Server
    * Setup SSL Certificate
    * Switch To Server's Admin
    * Transfer Virtual Server

 * **DNS Configuration** could include:
    * DNS Options
    * DNS Records
    * DNS DKIM Record (as current DomainKey Options page minus options, only full record display)
    * Suggested DNS Records

* **Web Configuration** could include:
   * Configure Website
   * Configure SSL Website
   * Edit Web Pages
   * PHP Options
   * PHP-FPM Configuration
   * Proxy Paths
   * Proxy Settings (curr. Edit Proxy Website)
   * Website Options
   * Website Redirects
   * Preview Website
   * Protected Directories

* **Mail Configuration** could include:
   * Email Settings
   * Signing Keys (as current DomainKey Options page minus DNS record)
   * Signup Email
   * Spam and Virus Delivery
   * SpamAssassin Configuration
   

Additionally **Logs and Reports** would get:
   * Disk Usage

marclaporte:
@nekohayo @fakemoth

jcameron:
Not sure I like "Control Center" as a name ... how about something like "Manage Server" or "Manage Domain" ?

iliajie:
"Management" seems to unspecific to me ..

.. "Manage Server" sounds somewhat alright. Although probably "Manage Virtual Server" would be a bit more fitting to the idea of "virtual servers" rather thank "domains" .. ? But it's a bit too long now ..

Perhaps we could just call it "Server Actions" .. ?

If not what other suggestions do we have?

fakemoth:
[...]

PR #669, opened 2023-10-20
Navigation menu reorganization PR

iliajie:
This change is made in relation to https://github.com/virtualmin/virtualmin-gpl/issues/663. I haven't commited changes for the other Virtualmin modules yet, but I will tomorrow to ensure everything functions as expected.

Let's discuss this PR first! This is how it looks now:

image

image

image

image

Does the menu make more sense now?


jcameron:
Will we also need plugin updates for this change, so that their menu items appear in the right place?

iliajie:
Yes, we do and as I mentioned earlier I will just push the commits after we agree on the PR and merge it.

jcameron:
What will happen if users update Virtualmin but not theme packages though? Will the UI break?

iliajie:
If they install Webmin 2.104 then all icons will display properly. Prior Webmin versions will display default icon instead.

PR #670, opened 2023-10-29
Fix EC support for `cert_file_info` sub

iliajie:
Jamie hi!

I have noticed that **SSL Certificate** page for a domain in Virtualmin always had turquoise button displayed at the bottom of the screen, suggesting to use this SSL certificate as the default in Webmin, Usermin, Dovecot, Postfix and MariaDB, in case virtual server used EC certificate type.

I started digging and found out that EC certificate doesn't have modulus and and exponent properties.

This PR fixes the correct handling for both RSA and EC certificate types.

Also, it fixes the key size detection to support new format.

### DIFF
--- a/get-ssl.pl
+++ b/get-ssl.pl
@@ -73,10 +73,11 @@ if (!$chain) {
 if ($cafile) {
 	print "ca: ",$cafile,"\n";
 	}
-print "type: ",&get_ssl_key_type(
+my $keytype = &get_ssl_key_type(
 		&get_website_ssl_file($d, "key"),
-		$d->{'ssl_pass'}),"\n";
-foreach $i (@cert_attributes) {
+		$d->{'ssl_pass'});
+print "type: $keytype\n";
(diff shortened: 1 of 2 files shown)


get-ssl.pl
```diff
@@ -73,17 +73,22 @@ package virtual_server;
...
+foreach my $i (@cert_attributes) {
 	$v = $info->{$i};
 	if (ref($v)) {
 		foreach my $vv (@$v) {
 			print $i,": ",$vv,"\n";
 			}
 		}
 	elsif ($v) {
+		if ($keytype =~ /^ec/) {
+			$i = "pub" if ($i eq "modulus");
```
jcameron, line 89: Why rename the fields like this?

iliajie: To be consistent with `openssl` command, as EC doesn't have neither `modulus` nor `exponent`.

jcameron: Ok makes sense

PR #671, opened 2023-10-30
Fix certificate key size test for RSA only

iliajie:
Hello Jamie!

After yesterday's PR, as the key size now correctly fetched, the false positive message started to appear for the certificate using EC algo, e.g.:

image

This PR fixes that and also a few small other tightly related issues. You can see 4 commits on this PR to get perspective what was changed.

### DIFF
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -11659,7 +11659,7 @@ if ($ENV{'HTTPS'} eq 'ON') {
 	local $certfile = $miniserv{'certfile'} || $miniserv{'keyfile'};
 	if ($certfile) {
 		local $cert = &cert_file_info($certfile);
-		if ($cert->{'size'} && $cert->{'size'} < 2048) {
+		if ($cert->{'algo'} eq 'rsa' && $cert->{'size'} && $cert->{'size'} < 2048) {
 			# Too small!
 			$small = $cert;
 			}
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -11676,7 +11676,12 @@ if ($small) {
 			   $small->{'issuer_o'},
 			   $small->{'issuer_cn'},
 			   )."<p>\n";
-	$alert_text .= &ui_form_start("@{[&get_webprefix_safe()]}/webmin/edit_ssl.cgi");
+	my $formlink = "@{[&get_webprefix_safe()]}/webmin/edit_ssl.cgi";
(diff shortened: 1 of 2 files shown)


jcameron:
Is this showing a warning for the Webmin SSL cert, or for a domain?

iliajie:
> Is this showing a warning for the Webmin SSL cert, or for a domain?

It's reading certificate for Webmin, but Webmin certificate can also be setup by a domain.

I'm not sure what's the best solution for the links. I'd assume if a domain is found in Virtualmin then it handles Webmin certificate as well.

virtual-server-lib-funcs.pl
```diff
@@ -11670,15 +11670,21 @@ sub list_warning_messages
...
-			   $small->{'issuer_c'} || $small->{'issuer_o'},
-			   )."</b><p>\n";
-	$alert_text .= &ui_form_start("@{[&get_webprefix_safe()]}/webmin/edit_ssl.cgi");
+			   "<tt>$small->{'cn'}</tt>",
+			   $small->{'issuer_o'},
+			   $small->{'issuer_cn'},
+			   )."<p>\n";
+	my $formlink = "@{[&get_webprefix_safe()]}/webmin/edit_ssl.cgi";
+	my $domid = &get_domain_by('dom', $small->{'cn'});
+	if ($domid) {
```
jcameron, line 11681: Shouldn't the whole form be inside this if statement?

iliajie: I'd think so, but you clearly wanted to link to SSL Encryption page in Webmin Configuration.

jcameron: Oh I see, actually it's fine as-is

virtual-server-lib-funcs.pl
```diff
@@ -11670,15 +11670,21 @@ sub list_warning_messages
...
+		$formlink = "@{[&get_webprefix_safe()]}/$module_name/cert_form.cgi?dom=$domid";
+		}
+	$alert_text .= &ui_form_start($formlink);
 	$alert_text .= &ui_hidden("mode", $msg eq 'licence_smallself' ?
-					'create' : 'csr');
+					'create' : $small->{'issuer_o'} =~ /let.*?encrypt/i ?
```
jcameron, line 11686: We have a `is_letsencrypt_cert` function that you can use to detect if a cert is from LE

iliajie: Alright, nice! I have fixed that!

PR #673, opened 2023-10-30
Add phpMyAdmin 6.0.0 devel version  support

iliajie:


scripts/phpmyadmin.pl
```diff
@@ -19,20 +19,32 @@ sub script_phpmyadmin_longdesc
 # script_phpmyadmin_versions()
 sub script_phpmyadmin_versions
 {
-return ( "5.2.1", "4.9.11" );
+return ( "5.2.1", "4.9.11", "6.0.0" );
```
jcameron, line 22: The highest version should be first

iliajie: No, it's devel version.

scripts/phpmyadmin.pl
```diff
@@ -194,25 +206,28 @@ sub script_phpmyadmin_check
...
+	$ver = $ver."-all-languages";
+	}
+else {
+	$ver = $ver."-english";
 	}
-elsif (&compare_versions($ver, "2.10.0") >= 0) {
-	$ver = $ver."--all-languages-utf-8-only";
+$url = "https://files.phpmyadmin.net/phpMyAdmin/$origver/phpMyAdmin-$ver.zip";
+if (&compare_versions($ver, 6) >= 0) {
+	$url = "https://files.phpmyadmin.net/snapshots/phpMyAdmin-$ver.tar.gz";
```
jcameron, line 226: Why can't we use the ZIP file for the 6.x version?

iliajie: Oh, we can. I just thought it was `tar.gz` at the first place.

scripts/phpmyadmin.pl
jcameron: Yeah, but some code assumes that versions are in descending order.

scripts/phpmyadmin.pl
jcameron: Let's keep it consistent as .zip then

scripts/phpmyadmin.pl
iliajie: Hmm, how do we make it right then? I don't want to adversise devel version by default.

scripts/phpmyadmin.pl
iliajie: Alright, fixed!

scripts/phpmyadmin.pl
jcameron: Hmmm ... I feel like we need an extra function for the "recommended version" of each script, which if not set defaults to the highest version. I'll add that.

iliajie: Should we add one more function to prevent showing in the UI that an update is available but cannot update? In this case updates from 5 to 6 should be silenced completely because 6 is the devel version of the script.

jcameron: Isn't there already a can_update function that checks for this?

jcameron: Hmm ... do we want to let users know that a new version theoretically exists that they could upgrade to by removing and re-installing though?

swelljoe: I don't really like including devel versions of stuff.

iliajie: > do we want to let users know that a new version theoretically exists that they could upgrade to by removing and re-installing though?

Yes, but not for devel version though.

> I don't really like including devel versions of stuff.

I think it would be useful for some scripts but it's important not to talk about its updates in the UI at all.

iliajie: I mean not to talk about upgrades 5 (stable) to 6 (devel).

jcameron: Oh do you mean this message : 

image

iliajie: Yeah, this message exactly. If a user has version 5 installed there should be no such message displayed, because 6 is devel version.

jcameron: What if 6.0 was a stable version, but not upgradable to?

iliajie: That would be handled by existing function and just return `0`. But for now, we just need to return `-1` (for example) to make sure update message is never even mentioned to the user.

jcameron: So return -1 from the `script_phpmyadmin_can_upgrade` function to indicate that the upgrade shouldn't even be mentioned?

iliajie: If `-1` isn't currently used, then yes! It would be great!

iliajie: When version 6 is out and (if) won't be upgradeable, we will just return `0` from that function, instead of `-1`. So `1` - should be like, okay, go and upgrade it, `0` - a message showing that there is an upgrade but we cannot make it automatically and `-1` won't even show it.

jcameron: Ok .. but if we change the meaning of the can_upgrade function, we'd need to put out a new Virtualmin release first to interpret -1 properly.

iliajie: But there must be some other way, maybe?

jcameron: Unfortunately not ... so we should add support for a -1 status first in `list_scripts.cgi`

iliajie: Then I will disable devel version 6.0 for now.

jcameron: Ok I added support for -1 return value in https://github.com/virtualmin/virtualmin-gpl/commit/f6c6dba4fe055dc7af0368c4c5320392dadc7663

iliajie: Alright, cool! Thanks!

PR #677, opened 2023-11-01
Fix to prioritize sub-domains over aliases in sorting

iliajie:
I noticed that domains in index page and nav menu aren't sorted correctly. Currently it prioritizes aliases over sub servers, and also doesn't make any difference between alias of top-level server and alias of sub-server e.g.:

image

This changes makes it displayed correctly, e.g.:

image

All tests were done with default sorting mode (`domains_sort` set to `sub`).


iliajie:
For Authentic it would look like this:

image

PR #678, opened 2023-11-03
Fix to match virtual server Apache log with `tee` in any order

iliajie:
Jamie hi!

This fixes the virtual server `error_log` detection used with `tee` command.

If we support it, I wish we support it in most precise way, meaning that we don't expect the user to use specific order when putting multiple log files, meaning that _whatever_ works in Apache itself should work in Virtualmin. This is the phylosophy of Webmin.

Also, it fixes the problem in the original regex, that was catching quotes, if a log file was quoted.

This patch will work with the following formats:

```
ErrorLog '|/usr/bin/tee -a "/var/log/apache2/error.log" "/var/log/virtualmin/ubuntu22-pro.virtualmin.dev_error_log"'
```

and also with

```
ErrorLog "|/usr/bin/tee -a '/var/log/apache2/error.log' '/var/log/virtualmin/ubuntu22-pro.virtualmin.dev_error_log'"
```

just as with

```
ErrorLog '|/usr/bin/tee -a /var/log/apache2/error.log /var/log/virtualmin/ubuntu22-pro.virtualmin.dev_error_log'
```

or

```
ErrorLog "|/usr/bin/tee -a /var/log/apache2/error.log /var/log/virtualmin/ubuntu22-pro.virtualmin.dev_error_log"
```

All those are equaly well supported by Apache.


feature-web.pl
```diff
@@ -1434,7 +1434,7 @@ sub extract_writelogs_path
 			}
 		}
 	}
-elsif ($log =~ /^\|(\$)?(tee|\S+\/tee)(\s+\-a)?\s+("[^"]+"|\S+)/) {
+elsif ($log =~ /^\|(\$)?(tee|\S+\/tee)(\s+\-a)?.*?([^'"\s]+\Q$dom\E[^'"\s]+)/) {
```
jcameron, line 1437: What if the admin is using a path that contains the domain ID, instead of the domain name? Seems like if `tee` is in use, we should return the first log file (as the original regexp did) if none of them contain the domain name.

iliajie: Good point! Fixed here https://github.com/virtualmin/virtualmin-gpl/pull/678/commits/5b61f7a8887c1cbda196bc0311496cf549e94361.

jcameron:
Thanks for the patch!

iliajie:
> Thanks for the patch!

Welcome!

Although, there is a small bug to fix, which I will address on the separate PR.

PR #707, opened 2023-12-06
Add major overhaul for `Edit Users` page

iliajie:
Hey Jamie!

As we discussed earlier, I have finally finished reworking **Edit Users** page!  &nbsp; 🚀 

This change will help to make managing users in Virtualmin in much more clear and distinct way.

This is a quick overview of how the page looks now after the rework:

image

It is much cleaner and displays **Login access** in much more precise way. It will mention all services to which given user has access currently, including **Databases** and **Web**.

**Real name** column will only be displayed if used by any of the users.

**Databases** column is hidden by default now but can be re-enabled to be displayed using Virtualmin configuration. **Login access** column now mentions Database access, but if chosen to be shown, in addition to previously used **Yes**, **No**, and **All**, it will also show a number of databases available to the given user.

Master administrator now by default will be able to create SSH users for the virtual server, but only in case **Custom Shells** are not customized (set to use defaults). If user configured custom shells  then it will respect the user configuration.

I have also reworked and created new help files to consider new (default) mode, where only one user name is created (without creating additional Unix user with a dash).

In addition, new help files were created for each new user mode created, which I will post in the screenshots below to get a clearer idea about what's going on.

The old **`edit_user.cgi`** page was reworked to group each service under specific accordion for logic clarity. It encompasses all old features though but in slightly improved order, for example:

image

**Quota and home directory settings** accordion label is now conditional, depending on the content, i.e. either **Quota Settings** or **Home directory settings** or both.

Finally, let me speak about new buttons and functionality they provide. In fact, the first four buttons, i.e. **Add User**, **Add SSH user**, **Add FTP user** and **Add Mail user** embed functionality of original **`edit_users.cgi`** page. However, unlike **Add User** button, which essentially provides the same functionality as before, the other three button, **Add SSH user**, **Add FTP user** and **Add Mail user** have very specific application with all unrelated features disabled.

Alright, let me show you how those new pages look, and also show the new help tips with detailed explanations about what's available for the given user type when it's first created.

**Add SSH user**

image

image

**Add FTP user**
image

image

**Add Mail user**

image

image

The other two user types are somewhat new as they don't create a separate Unix user. It was always not clear how to create database or webserver users only. Creating the new Unix user for this particular purpose always seemed overkill. Now new **Add database user** and **Add webserver user** will do exactly that.

These kind of users are virtual to Virtualmin and are stored in the domain config the same way associated databases are.

Here is the example of **Add database user** page:

image

image

And here at last is the example of **Add webserver user** page:

image

image

Standalone virtual server's database users can be controlled via CLI already. Webserver users not yet, as it would require some additional logic for extracting currently assigned access for web directories. Although, it isn't a big deal to add it if needed.

The only missing bit is recreating of virtual database users upon restore. It also can be easily added later if needed.

The rest of the functionality in this PR was tested quite thoroughly and must just work.

Please let me know if you have any questions.

☮️


jcameron:
This looks nice! But do all the user types have a Unix user though? It seems like the "database user" only exists in MySQL, or am i getting that wrong?

iliajie:
> But do all the user types have a Unix user though?

No, in particular **Database** and **Webserver** users can be created without the Unix user. Later though it's posible to merge it with Unix type of user by creating standard user.

> It seems like the "database user" only exists in MySQL, or am i getting that wrong?

Yes, that is correct. I always wanted to have an easy way to create database and webserver users associated with virtual server but without the need of creating the Unix user for it.

iliajie:
Where are we at with this? Do you have any other questions about the PR?

It can be done later, after the merge but I'm thinking now to also add ability to control public key for SSH user.

jcameron:
Is a "webserver user" just one who has access to some directories under public_html via a .htaccess file?

Because we also have a "webserver FTP" user which does need a Unix user, but with the same UID as the domain.

iliajie:
> Is a "webserver user" just one who has access to some directories under public_html via a .htaccess file?

Yes, it's just a separate user for a webserver.



> Because we also have a "webserver FTP" user which does need a Unix user, but with the same UID as the domain.

Yes, I'm aware. It is also implemented using **Add FTP user** button, as described at the very beginning of this PR (with screenshots).

jcameron:
I see ... hmm, this is quite a large change and hard to evaluate to be honest! 

Backing up a bit, what is the main motivation behind this change? Do we benefit a lot by creating users that only exist in MySQL or `.htusers` files, rather than the current model where all users are Unix users?

iliajie:
> Do we benefit a lot by creating users that only exist in MySQL or `.htusers` files, rather than the current model where all users are Unix users?

We benefit a lot!

Managing users within application context (like a database or webserver user) is simpler than managing system users, as users are isolated from the operating system, meaning they have no access to system resources outside the scope of the application.

This is particularly important for web applications which our users use quite a lot.

Also, it’s easier to scale an application with specific types of users, like defined in a database or `.htusers` file, especially when scaling across multiple servers or in a **_cloud environment_**.

Moreover, using this approach we better stick to the principle of least privilege (PoLP), avoiding the creation of Unix users when it is not necessary.

Personally, I've found it irritating that setting up a simple database or webserver user isn't more straightforward, and I always had to do it manually, yet it never worked upon backup and restore.

Finally, placing separate buttons on the main page provides a _**much**_ clearer overview of the available options for the virtual server. Many users have wondered about adding database users, which, at present, is not very clear, as deeply nested within the "Add New User" page, and under "Other user permissions" accordion.

> this is quite a large change and hard to evaluate to be honest!

Yet, nothing extraordinary, as I'm just using the standard Virtualmin API for doing it ..

jcameron:
Ok ... I will take another look at the PR shortly.

iliajie:
> Ok ... I will take another look at the PR shortly.

Great, thank you!

iliajie:
> > Ok ... I will take another look at the PR shortly.
> 
> Great, thank you!

I'm opened to discuss this PR or have it to be merged.

create-user.pl
```diff
@@ -361,8 +366,31 @@ package virtual_server;
...
 	}
 
+# Create database user only
+if ($db_only) {
+	my @dbusers = &list_domain_users($d, 1, 1, 1, 0, 1);
+        my ($user_already) = grep { $_->{'user'} eq $user->{'user'} } @dbusers;
+        !$user_already || &error(&text('user_ealreadyexist', $user->{'user'}));
+	$user->{'pass'} = $pass;
+	# Create database user
+        my ($err, $dts) = &create_databases_user($d, $user);
```
jcameron, line 376: Where is this function defined? For some reason I don't see it in this diff view ...

edit_user_db.cgi
```diff
@@ -0,0 +1,104 @@
...
+			$pwfield,
+			2, \@tds);
+
+# List databases
+my @dbs;
+@dbs = grep { $_->{'users'} } &domain_databases($d) if ($d);
+
+# Show allowed databases
+if (@dbs) {
+        print &ui_table_row(undef, "<hr data-row-separator>", 2);
```
jcameron, line 64: Isn't there a ui_table_hr function for this?

edit_user_web.cgi
```diff
@@ -0,0 +1,114 @@
...
+if (!$in{'new'}) {
+        my @webusers = &list_domain_users($d, 1, 0, 1, 1, 1);
+        ($webuser) = grep { $_->{'user'} eq $in{'user'} } @webusers;
+        $webuser || &error(&text('user_edoesntexist', &html_escape($in{'user'})));
+        $webuser_name = &remove_userdom($webuser->{'user'}, $d) || $webuser->{'user'};
+        }
+
+# At first check if we have protected webdirectories in this domain
+my $htpasswd_data;
+foreach my $f (&list_mail_plugins()) {
```
jcameron, line 38: Why is the same loop nested??

create-user.pl
iliajie: It is right here https://github.com/virtualmin/virtualmin-gpl/commit/a2caf96a353fcec1a3707979aba03ab7116f8ae6#diff-a8b0a88367118c8f13666eed07287b13ecc217c8fdfaf6a34d51b226dc196f92R1208.

edit_user_db.cgi
iliajie: You're right! I forgot about using `ui_table_hr`, after running the initial tests with this inline example. Fixed here https://github.com/virtualmin/virtualmin-gpl/commit/5d9fedbceb2b7f482f86ade746e0ad94f7f38379.

edit_user_web.cgi
iliajie: You're right, that's a typo from! Fixed in this patch https://github.com/virtualmin/virtualmin-gpl/commit/b78c28e44bb778265fc82289fac3a17c1a99e841.

create-user.pl
jcameron: Is there a way to see the entire diff with all updates in a single view? I always trouble finding this in github when reviewing PRs ..

iliajie: > Is there a way to see the entire diff with all updates in a single view?

Yes, there is.

You can just add `.patch` to any commit or even PR in GitHub to load the diff, e.g.:

https://patch-diff.githubusercontent.com/raw/virtualmin/virtualmin-gpl/pull/707.patch

jcameron: Thanks, I'll add some comments on that other commit

iliajie: Alright, fully done!

jcameron:
I had another look at this PR, and I noticed there's a lot of duplication between the various edit_user_ scripts. Can these perhaps be combined, or at least some of the duplicate code refactored?

iliajie:
> I noticed there's a lot of duplication between the various edit_user_ scripts.

Thanks for your suggestion.

I made each script intentionally separated for clarity and focused functionality, as combining all the `edit_user_` scripts into a single file could significantly increase complexity, making the code harder to support and maintain, and make it potentially more error-prone. Although, it isn't a problem for me to do it (technically) though.

> Can these perhaps be combined

If complexity is not a concern, then yes, I could do it.

> .. or at least some of the duplicate code refactored?

Could you specify which parts of duplicated code you've noticed that could benefit from refactoring?

iliajie:
> > Can these perhaps be combined or at least some of the duplicate code refactored?
> 
> Could you specify which parts of duplicated code you've noticed that could benefit from refactoring?

Actually, it won't be any much more complex. Alright, let me combine it in one file and optimize the code ..

iliajie:
> I had another look at this PR, and I noticed there's a lot of duplication between the various edit_user_ scripts. Can these perhaps be combined, or at least some of the duplicate code refactored?

Alright, done https://github.com/virtualmin/virtualmin-gpl/pull/707/commits/ee045fa6a3194b0fe28bd6d559a45b5554b4e39f.

jcameron:
Thanks! I will take another look ...

jcameron:
So in this PR, is a "webserver user" one who just have permissions on some web directory via a .htaccess file?

Also, do we still support users having multiple types of permissions (ie. mail, mysql and FTP) ?

feature-web.pl
```diff
@@ -5473,6 +5473,51 @@ sub reset_web
...
+        if ($f eq "virtualmin-htpasswd") {
+                $err = &plugin_call($f, "mailbox_validate", $user, $olduser,
+                                    \%in, $in{'new'}, $d);
+                &error($err) if ($err);
+                }
+        }
+
+# Run plugin save functions
+foreach $f (&list_mail_plugins()) {
+        if ($f eq "virtualmin-htpasswd") {
```
jcameron, line 5498: Same here

feature-web.pl
```diff
@@ -5473,6 +5473,51 @@ sub reset_web
...
+&update_extra_user($d, $user, $olduser);
+}
+
+# delete_webserver_user(&user, &domain)
+# Delete a webserver user
+sub delete_webserver_user
+{
+my ($user, $d) = @_;
+foreach my $f (&list_mail_plugins()) {
+        if ($f eq "virtualmin-htpasswd") {
```
jcameron, line 5513: Also here ... this loop seems unnecessary

feature-mysql.pl
```diff
@@ -1350,6 +1366,26 @@ sub restore_mysql
 		}
 	}
 
+# Restoring virtual MySQL users
+my @dbusers_virt = &list_extra_db_users($d);
```
jcameron, line 1370: What if the user was deleted before restoring? Won't this return an empty array, and so no users will be created?

iliajie:
> So in this PR, is a "webserver user" one who just have permissions on some web directory via a .htaccess file?

Yes, exactly it's just a record in `.htpasswd` file, without creating any other type of users.

The same logic applies to "database users", where a user only created in a database (not Unix user or anything).

It is important to point out that it will be possible to create user "jamie" as a "webserver user", and the user with the same username, i.e. "jamie" as a "database user". However, if later a Unix user created (i.e. using "Add user" or "Add SSH user" or "Add mail user" page) with the same name, i.e. "jamie" then this process of creation of Unix user will "take-over" those extra users.

Also, if a Unix user "jamie" existed before, it won't be possible to create any extra users (i.e. database or webserver) with that username.

> Also, do we still support users having multiple types of permissions (ie. mail, mysql and FTP) ?

What do you mean by "multiple types of permissions"?

feature-web.pl
```diff
@@ -5473,6 +5473,51 @@ sub reset_web
...
+my ($user, $olduser, $d) = @_;
+
+# Encrypt user initial password
+ $user->{'pass'} = &encrypt_user_password($user, $user->{'pass'})
+        if (!$in{'webpass_def'});
+$user->{'pass_crypt'} = $user->{'pass'};
+
+# Validate plugins
+foreach my $f (&list_mail_plugins()) {
+        if ($f eq "virtualmin-htpasswd") {
```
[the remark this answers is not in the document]
iliajie: This could be improved, I agree. I made this call because I wanted to make sure that `virtualmin-htpasswd` is actually availble.

What is the better way to do it? Can we here just directly use `plugin_defined`?

feature-mysql.pl
iliajie: > What if the user was deleted before restoring? Won't this return an empty array, and so no users will be created?

This type of user (extra) is stored in the backup file itself, as a record under `/etc/webmin/virtual-server/users/db`. For instance user `example` would be stored as `example.user`. So, upon restore for every extra user record found in that directory, we just re-create it on the new system.

feature-web.pl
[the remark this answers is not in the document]
iliajie: > Why use a loop here if we're only making the call for one plugin type?

Alright, thanks! Fixed here https://github.com/virtualmin/virtualmin-gpl/pull/707/commits/9c6d956be68da23435598f31981b98e6e4700796.

feature-mysql.pl
jcameron: Oh I see. But what if when restoring *some* of the DB users already exist? Or all of them?

jcameron:
> What do you mean by "multiple types of permissions"?

I mean a user with web access and mysql and FTP. Previously all of these were configured as attributes of a single user..

feature-mysql.pl
iliajie: > But what if when restoring _some_ of the DB users already exist? Or all of them?

Currently, it will just override those DB users, including changing passwords and permissions!

Indeed, is this alright? Or you think we should skip doing it, and instead delete an association to this DB user from restored virtual server?

iliajie:
> I mean a user with web access and mysql and FTP. Previously all of these were configured as attributes of a single user.

Extra users don't have any specific attributes, at least not database or webserver users. If we ever add another user type in the future, its attributes would depend on the service used.

However, as I explained earlier extra users created under **Add database user** and **Add webserver user** can have the same name. This is because users exists separately in independent services. That said, creating database user named `jamie`, and then creating webserver user named `jamie` won't produce any errors.

create-user.pl
```diff
@@ -361,8 +366,30 @@ package virtual_server;
...
+        # Add user to domain list
+	$dbuser->{'user'} = $user->{'user'};
+	$dbuser->{'pass'} = $pass;
+	$dbuser->{'extra'} = 1;
+        $dbuser->{'type'} = 'db';
+        foreach my $db (@{$user->{'dbs'}}) {
+                $dbuser->{'db_'.$db->{'type'}} .=
+                        $dbuser->{'db_'.$db->{'type'}} ? " $db->{'name'}" : $db->{'name'};
+                }
+        &update_extra_user($d, $dbuser);
```
jcameron, line 387: Why the need to call create_databases_user and then update_extra_user ? Shouldn't this be part of create_databases_user ?

feature-mysql.pl
jcameron: So create_databases_user will overwrite a user if he already exists? 

In that case, what if a DB user was created after the backup was taken? Will the current code remove that user when restoring the backup?

feature-postgres.pl
```diff
@@ -719,6 +719,26 @@ sub restore_postgres
 		}
 	}
 
+# Restoring virtual PostgreSQL users 
+my @dbusers_virt = &list_extra_db_users($d);
```
jcameron, line 723: Won't this also return extra MySQL users?

mass_change.cgi
```diff
@@ -12,7 +12,7 @@ $d = &get_domain($in{'dom'});
 
 &obtain_lock_unix($d);
 &obtain_lock_mail($d);
-@users = &list_domain_users($d);
+@users = &list_domain_users($d, 0, 0, 0, 0, 1);
```
jcameron, line 15: W

mass_change.cgi
```diff
@@ -44,6 +44,13 @@ foreach $user (@musers) {
 	&$indent_print();
 	$pop3 = &remove_userdom($user->{'user'}, $d);
 
+	# Skip virtual (extra) users
+	if ($user->{'extra'}) {
```
jcameron, line 48: Why include extra users in the response from list_domain_users , only to skip them here?

modify-user.pl
```diff
@@ -512,7 +512,28 @@ package virtual_server;
 &usage($err) if ($err);
 
 # Save the user
-&modify_user($user, $olduser, $d);
+if ($user->{'extra'}) {
+	if ($user->{'type'} eq 'db') {
+		&modify_database_user($user, $olduser, $d);
+		my %dbuser = %{$user};
+		%dbuser = map { $_, $dbuser{$_} }
```
jcameron, line 519: Why do you need to re-create the user hash here with only a subset of keys?

virtual-server-lib-funcs.pl
```diff
@@ -1062,7 +1062,14 @@ sub list_domain_users
 		}
 	}
 
-if (!$_[4] && $d) {
+# Push domain extra database users
+if ($includeextra) {
+	foreach (&list_extra_db_users($d)) {
```
jcameron, line 1067: Instead just use `push(@users, &list_extra_db_users($d))`

virtual-server-lib-funcs.pl
```diff
@@ -1111,6 +1118,31 @@ sub list_domain_users
 				}
 			}
 		}
+	if ($includeextra) {
+		# If extra database users overlap with system users, which already
+		# handle databases, pop them off the list and remove extra db user
+		my (@extra_database_users) = grep { $_->{'extra'} && $_->{'type'} eq 'db' } @users;
+		if (@extra_database_users) {
+			my @other_users_with_dbs = map { $_->{'user'} } grep { 
+				$_->{'dbs'} && ref($_->{'dbs'}) eq 'ARRAY' &&
```
jcameron, line 1127: No need to check if `$_->{'dbs'}` is an ARRAY ref if it exists, because presumably it would never be a scalar

virtual-server-lib-funcs.pl
```diff
@@ -1119,6 +1151,36 @@ sub list_domain_users
...
+		if (@other_users_with_webs) {
+			@users = grep { 
+				$_->{'type'} ne 'web' ||
+				&indexof($_->{'user'}, @other_users_with_webs) < 0
+			} @users;
+			# Now clear Virtualmin record of extra webusers users
+			# belonging to an actual user account
+			foreach my $webserver_virt (@webserver_virts) {
+				if (&indexof($webserver_virt->{'user'}, @other_users_with_webs) != -1) {
+					&delete_webserver_user($webserver_virt, $d);
```
jcameron, line 1177: This shouldn't be happening when listing users! If there is a clash, it should be resolved when saving the user.

create-user.pl
iliajie: At first we actually create a database user, and then we keep an association for this user in `/etc/webmin/virtual-server/users/db` directory.

feature-mysql.pl
iliajie: > So create_databases_user will overwrite a user if he already exists?

Yes! But this is opened to discussion. I can throw an error upon restore, saying that this user already exists and disassociate it from the virtual server.

> In that case, what if a DB user was created after the backup was taken? Will the current code remove that user when restoring the backup?

When restoring a backup current code will always replace existing database user, in case it's found on the backup. However, as I already mentioned it's an opened discussion.

It's worth pointing out that, if the database user already exists, and then an admin tries to create a new database user using this new feature, i.e. using **"Add database user"** button (or using CLI with `create-user --db-only` command), then the error would be returned in this case, saying that the database user already exists.

feature-postgres.pl
iliajie: This will return just a database user. Whether it's a PostgreSQL user, or MySQL/MariaDB user is irrelevant to this code. It ain't its business to care about it.

mass_change.cgi
iliajie: It is done to avoid confusion. For clarity. Because we select a user on the UI, in the previous page, so we need to show that we're operating on this user .. not just mystically skipping on it.

modify-user.pl
iliajie: This is done to save to `/etc/webmin/virtual-server/users/db/db-user-example1.user` only specific keys and not to mix it with the `$user` object that is being passed to `modify_database_user`.

virtual-server-lib-funcs.pl
iliajie: Thanks for the suggestion! Done in this patch https://github.com/virtualmin/virtualmin-gpl/pull/707/commits/db5e978f0c0cfac0070a1273d1f58fc60f6c89a8.

virtual-server-lib-funcs.pl
iliajie: I'd rather kept it as is, cause if we switched to `use strict;` it could break.

virtual-server-lib-funcs.pl
iliajie: > This shouldn't be happening when listing users!

I agree. This is not the right place. I should really do it at the time we save a new Unix user. I will look into fixing this  tomorrow.

> If there is a clash, it should be resolved when saving the user.

Yes, right!

create-user.pl
jcameron: Seems like it would be cleaner to combine this into a single function though. Presumably we never want to create a user that we don't record in the `users/db` directory?

mass_change.cgi
jcameron: Ok makes sense

virtual-server-lib-funcs.pl
jcameron: Great, thanks!

virtual-server-lib-funcs.pl
jcameron: But where in the code could `'dbs'` ever be set to anything other than an array ref? It's kind of confusing to have this check otherwise, because the reader might think that `'dbs'` could legitimately be a string.

modify-user.pl
jcameron: Wouldn't it be cleaner for both functions to just operate on the same kinds of objects? What extra keys could be in the hash?

feature-postgres.pl
jcameron: But later on you call `&create_databases_user($d, $dbuser_virt, 'postgres');` with the same user objects. What's that going to do if it's a MySQL user?

create-user.pl
iliajie: > Presumably we never want to create a user that we don't record in the users/db directory?

We do, in case we create an old type, Unix based user. We cannot combine it.

modify-user.pl
iliajie: > What extra keys could be in the hash?

There are a bunch of extra and unrelated keys in `$user` object, including array refs.

feature-postgres.pl
iliajie: It simply won't create a user unless a database is associated with it. It acts the same way as before. If we remove access from all databases for a database user, it will be deleted.

In this case, a MySQL user will only be created if there is a `db_mysql` record exists for this extra database user.

create-user.pl
jcameron: Ok, I see now..

feature-postgres.pl
jcameron: Ok I see, the last param is the database type. Makes sense ...

virtual-server-lib-funcs.pl
iliajie: > But where in the code could 'dbs' ever be set to anything other than an array ref? It's kind of confusing to have this check otherwise, because the reader might think that 'dbs' could legitimately be a string.

I have completely reworked this portion of the code. The code that you initially commented on was simply removed.

virtual-server-lib-funcs.pl
iliajie: > This shouldn't be happening when listing users! If there is a clash, it should be resolved when saving the user.

I have reworked this part, and made it much more small and efficient! The patch is here https://github.com/virtualmin/virtualmin-gpl/commit/916135fc1011de247209d6120c07c8011d34ff19.

feature-mysql.pl
jcameron: Ok, if a DB user already exists that was created outside of Virtualmin, then it's reasonable to generate an error.

I think the root issue is that we are relying on a separate config file to track database users which can get out of sync with what actually exists in MySQL. The cleanest solution would be if we could find all users who have access to any of a domain's databases. Could that work, or would we not have enough information?

users-lib.pl
```diff
@@ -0,0 +1,125 @@
+# Functions for managing extra users
+
+# list_extra_users(&domain, user-type, [username])
+# Returns a list of extra users for some domain
+sub list_extra_users
+{
+my ($d, $t, $u) = @_;
```
jcameron, line 7: Better to have a separate get_extra_user function to fetch a single user rather than fetching the whole list and then filtering it ..

modify-user.pl
jcameron: Anytime we have functions that return user objects and update them, they should be able to take the same kind of object without needing the caller to do filtering.

users-lib.pl
iliajie: > Better to have a separate get_extra_user function to fetch a single user rather than fetching the whole list and then filtering it ..

Yes, agreed! I have reworked it here https://github.com/virtualmin/virtualmin-gpl/commit/6a4dd65d8b5cb7320ccdea7a6ab1b6994e3023c2.

I have also fixed scoping bug that was introduced a few commits before, when combining `edit_user.cgi` page. It made current extra db and web users not no re-select active databases/web-directories when editing.

feature-mysql.pl
iliajie: > If a DB user already exists that was created outside of Virtualmin, then it's reasonable to generate an error.

Alright, agreed! We will test if a database user already exists outside of the domain name (which shouldn't happen) we will skip its re-creation process https://github.com/virtualmin/virtualmin-gpl/pull/707/commits/c29ce33f5d9b79f608abb9be8d64d9a91a8c7a90.

modify-user.pl
iliajie: > Anytime we have functions that return user objects and update them, they should be able to take the same kind of object without needing the caller to do filtering.

Agreed! Fixed in this commit https://github.com/virtualmin/virtualmin-gpl/commit/4c5b1cc4ecfa440f08708a6e56ca914d52551349. It will also fix and fully support storing hashed passwords for MySQL user if enabled for domain.

All done. Give it another look please!

virtual-server-lib-funcs.pl
```diff
@@ -1620,6 +1693,9 @@ sub create_user
 if ($_[1]) {
 	&create_jailkit_passwd_file($_[1]);
 	}
+
+# Suppress incompatible extra users
+&suppress_extra_user($_[0], $_[1]);
```
jcameron, line 1698: Does this mean that a Unix user can be created with the same name as an extra user, and it will then take over the extra user's capabilities?

virtual-server-lib-funcs.pl
```diff
@@ -2131,6 +2148,22 @@ sub modify_user
...
+if ($_[0]->{'extra'}) {
+	if ($_[0]->{'type'} eq 'db') {
+		# Delete database user
+		my ($err, $dts) = &delete_databases_user($_[1], $_[0]->{'user'});
+		&error($err) if ($err);
+		# Delete user from domain config
+		&delete_extra_user($_[1], $_[0]);
+		}
+	if ($_[0]->{'type'} eq 'web') {
+		&delete_webserver_user($_[0], $_[1]);
```
jcameron, line 2162: Why doesn't delete_extra_user need to be called when deleting a database user?

virtual-server-lib-funcs.pl
```diff
@@ -6895,8 +7068,10 @@ sub restore_virtualmin
 		}
 	if (-r $file."_admins") {
 		# Also restore extra admins
-		&execute_command(
-			"rm -rf ".quotemeta("$extra_admins_dir/$d->{'id'}"));
+		if ($d->{'id'}) {
```
jcameron, line 7071: When is the domain ID ever not set?

virtual-server-lib-funcs.pl
iliajie: Yes, exactly!

virtual-server-lib-funcs.pl
iliajie: To remove association in Virtualmin. We first delete an actual database user, and then remove a record in Virtualmin under `/etc/webmin/virtual-server/users`.

virtual-server-lib-funcs.pl
iliajie: I must always be set. However, there were issues in the past when domain config was destroyed.

In fact, this is just a precaution for `rf -rf` command, as any variable passed to `rm -rf` command should be checked if defined or if it has an expected value.

jcameron: Ok yeah that makes sense as an extra precaution..

virtual-server-lib-funcs.pl
jcameron: Sorry I meant, why isn't it called when deleting a webserver user?

virtual-server-lib-funcs.pl
jcameron: Hmmm ... that seems quite confusing to me. Wouldn't it be a lot cleaner to disallow this?

iliajie: > Hmmm ... that seems quite confusing to me. Wouldn't it be a lot cleaner to disallow this?

Well, not really! A Unix user created by Virtualmin covers functionality for both extra database user and webserver user. It would be bummer to disallow this, if later an admin wants to create a Unix user with the same username that intersects with either of extra users.

iliajie:
Jamie, I have fixed all issues that you pointed out before.

Today I have also added support for managing SSH public keys for domain users. It is important to have this functionality working in a straight-forward way.

You can see this patch here https://github.com/virtualmin/virtualmin-gpl/commit/d1b4c6eeba5c8780bfeafc0392461bb09c83eafd.

Please let me know if you have any questions.

Thanks!

jcameron:
Cool, thanks! Can you post a screenshot of what this SSH key management UI looks like?

feature-mysql.pl
jcameron: No what I mean was, why do we even need a separate config file of MySQL users at all? Can't we infer them directly from what users exist in MySQL? Or is that not possible because there's no way of associating them with a domain?

users-lib.pl
```diff
@@ -0,0 +1,211 @@
...
+	    $u ne &remove_userdom($u, $d)) {
+		$userclash = undef
+		}
+	}
+# Check for clash with Unix users first
+if (!$userclash) {
+	($userclash) = grep { $_->{'user'} eq $u }
+		&list_domain_users($d, 0, 0, 1, 1);
+	}
+return $userclash ?
```
jcameron, line 61: Is this function supposed to return an error string, a user object?

users-lib.pl
```diff
@@ -0,0 +1,211 @@
...
+{
+my ($d, $dbusername) = @_;
+foreach my $dt (&unique(map { $_->{'type'} } &domain_databases($d))) {
+	my $cfunc = "check_".$dt."_user_clash";
+	next if (!defined(&$cfunc));
+	my $ufunc = $dt."_username";
+	if (&$cfunc($d, &$ufunc($dbusername))) {
+		return &text("user_edbclash_$dt", &html_escape($dbusername));
+		}
+	}
```
jcameron, line 78: You're missing a return statement

users-lib.pl
```diff
@@ -0,0 +1,211 @@
...
+sub update_extra_user
+{
+my ($d, $user, $olduser) = @_;
+my $path = "$extra_users_dir/$d->{'id'}/$user->{'type'}";
+&make_dir($path, 0700, 1) if (!-d $path);
+if ($olduser->{'user'} && $user->{'user'} &&
+    $olduser->{'user'} ne $user->{'user'}) {
+        unlink(&extra_user_filename($olduser, $d));
+	}
+&write_file(&extra_user_filename($user, $d), &extra_user_object($user, $d));
```
jcameron, line 153: You should call lock_file before writing, and unlock_file afterwards

users-lib.pl
```diff
@@ -0,0 +1,211 @@
...
+# that cannot coexist with Unix users
+sub suppressible_extra_users_types
+{
+return ('db', 'web');
+}
+
+# suppress_extra_user(&unix-user, &domain)
+# Remove records of extra user that
+# cannot coexist with Unix user
+sub suppress_extra_user
```
jcameron, line 202: Is this still used? Seems like it shouldn't be necessary if clashes are prevented in the first place

feature-mysql.pl
iliajie: Yes, because we cannot know for sure which user belongs to which server. Also, if there is a lot of users it will be very slow.

virtual-server-lib-funcs.pl
jcameron: I disagree, the admin might accidentally create a unix user with the same name as a database user and end up taking it over (and maybe changing the password). Better to keep the interface and code simple and just not allow clashes.

virtual-server-lib-funcs.pl
jcameron: Seems odd that `delete_webserver_user` calls `delete_extra_user`, but `delete_database_user` does not?

feature-mysql.pl
jcameron: Ok makes sense

virtual-server-lib-funcs.pl
iliajie: Hmm, good point and can easily be the real case scenario .. perhaps there is something else we could do?

iliajie:
> Can you post a screenshot of what this SSH key management UI looks like?


image

virtual-server-lib-funcs.pl
jcameron: The simple solution is just not to allow clashes. This then allows you to remove all the logic for merging users..

virtual-server-lib-funcs.pl
jcameron: What I mean is, it feels like the APIs aren't consistent with each other.

jcameron:
Nice, thanks! Does this only appear for SSH-capable users?

virtual-server-lib-funcs.pl
iliajie: > What I mean is, it feels like the APIs aren't consistent with each other.

It is expected. 

> Seems odd that `delete_webserver_user` calls `delete_extra_user`, but `delete_database_user` does not?

`delete_webserver_user` should always call `delete_extra_user` as it only deals with extra users, whereas  `delete_database_user` does not.

iliajie:
> Does this only appear for SSH-capable users?

On the creation time it would appear for **Add SSH user** and for standard **Add user**. Not for others. Although, it would still be possible to add it later for mail user.

iliajie:
> Does this only appear for SSH-capable users?

And yes, this option appears only if SSH shell is available for domain.

users-lib.pl
iliajie: > You should call lock_file before writing, and unlock_file afterwards

Alright, fixed!

users-lib.pl
iliajie: > Is this function supposed to return an error string, a user object?

It always returns a string!

virtual-server-lib-funcs.pl
iliajie: > The simple solution is just not to allow clashes. This then allows you to remove all the logic for merging users..

The logic is very simple. I wouldn't want to remove it to be honest! It works really well and I tested it quite a lot. Also, it can be a useful feature.

I would prefer to add a warning instead, like a confirmation button if there is a clash. Or simply over take the existing extra database user permissions for the database, unless new given.

users-lib.pl
iliajie: > Is this still used? Seems like it shouldn't be necessary if clashes are prevented in the first place

We also have an extra web user which has to be suppressed.

jcameron: But why? Can an web user and a unix user have the same name?

users-lib.pl
jcameron: Oh I see .. makes sense now!

users-lib.pl
iliajie: Well, this is because creating a Unix user also creates associated database and web users.

If we just throw an error as you suggest, it means in order to create a new Unix user with the same name as an extra user, we would need to delete it first. I think it's an extra hassle, and pretty suboptimal. Nevertheless, silently over-riding an extra also isn't great.

So, what could be possible good solutions:

  a). Show a warning and confirmation button when a Unix user is created and an extra user with the same name exists;
  b).  Do the same as in a) but don't show a warning unless either **Databases access settings** or **Web directories access settings** are configured, i.e. would intersect with any existing extra user with the same name. After this new Unix user is created and later edited, we simply won't show an option to edit either **Databases access settings** and/or **Web directories access settings** as long as corresponded extra user with the same username exists;
  c). We use JavaScript and when Unix user field is filled (blurred) we just show a warning straight away (without submitting the form) that an extra user already exists with the same name, and that it will be overwritten.

I would prefer a) or c). 

But just showing an error and blocking a creation of Unix user, which can effectively manage any type of extra user doesn't seem optimal either.

iliajie: Jamie, have a look. I have added https://github.com/virtualmin/virtualmin-gpl/pull/707/commits/4fe8a84c9af9bfc02bf74219e8511ff5bfcca7b6 the confirmation screen, e.g.:

image

jcameron: I like option A, but we should make sure to do any merging of users at the time the Unix user is created rather than in the API that fetches extra users.

iliajie: > but we should make sure to do any merging of users at the time the Unix user is created

Currently merging happens in `create_user` time. I think you were referring to the old code that I already changed.

iliajie:
Jamie, I have retested it all over again. Fixed a few small bugs. All appears flawless to me now. Please have a final look!

users-lib.pl
jcameron: Ok cool we're good then!

iliajie:
@jcameron Jamie, is there anything else you want me to fix in this PR?

jcameron:
One thing that would be nice is to update functional-test.pl to cover any new API commands or flags for the new user types. That way you can test that web, FTP and SSH user creation and management works as expected..

iliajie:
> One thing that would be nice is to update functional-test.pl to cover any new API commands or flags for the new user types. That way you can test that web, FTP and SSH user creation and management works as expected..

Alright. I have added tests for extra MySQL user here https://github.com/virtualmin/virtualmin-gpl/commit/9cab3e7648417db44be0ec006928054d2c6aa519.

jcameron:
Thanks! Is all the new functionality like creating a website user and an FTP user accessible via the CLI as well?

iliajie:
FTP is rather an old functionality, and it all should be there already.

Web user creation is not available via CLI, at least not yet.

Adding a key to SSH user could be added as well. 

Should I add it too?

iliajie:
> Is all the new functionality like creating a website user and an FTP user accessible via the CLI as well?

@jcameron

Jamie, yes, now _all_ functionalities for creating, modifying and deleting extra database and webserver users available in CLI!

Please have another look!

jcameron:
Excellent, thanks! I aim for 100% coverage of all functionality in the API, so that we can do automated tests before a release. It would be nice if you also updated functional-test.pl to create a website user and test that it works, if that's possible?

iliajie:
> It would be nice if you also updated functional-test.pl to create a website user and test that it works, if that's possible?

Yes! Can do!

iliajie:
@jcameron 

Jamie, alright - all done!

iliajie:
All done! Let's wrap this up!

iliajie:
I think this PR is good to be merged?

create-user.pl
```diff
@@ -361,8 +402,40 @@ package virtual_server;
...
+	# Create initial user
+        $user->{'extra'} = 1;
+        $user->{'type'} = 'web';
+        my $userclash = &check_extra_user_clash($d, $user->{'user'}, 'web');
+        !$userclash || &usage($userclash);
+        # Set initial password
+        $user->{'pass'} = $pass;
+        $user->{'pass_crypt'} = $encpass, delete($user->{'pass'}) if ($encpass);
+        $user->{'pass'} || $user->{'pass_crypt'} || &usage($text{'user_epasswebnotset'});
+        &modify_webserver_user($user, undef, $d, { virtualmin_htpasswd => join("\n", @webdirs) });
```
jcameron, line 429: Shouldn't virtualmin_htpasswd be quoted?

feature-web.pl
```diff
@@ -5473,6 +5473,79 @@ sub reset_web
...
+&update_extra_user($d, $user, $olduser);
+}
+
+# revoke_webserver_user_access(&user, &domain)
+# Remove a webserver user access
+sub revoke_webserver_user_access
+{
+my ($user, $d) = @_;
+if (&plugin_defined("virtualmin-htpasswd", "mailbox_delete")) {
+    &plugin_call("virtualmin-htpasswd", "mailbox_delete", $user, $d);
```
jcameron, line 5534: Indent looks wrong

feature-web.pl
```diff
@@ -5473,6 +5473,79 @@ sub reset_web
...
+return @rv;
+}
+
+# modify_webserver_user(&user, &old-user, &domain, [input-data])
+# Create or update a webserver user
+sub modify_webserver_user
+{
+my ($user, $olduser, $d, $indata) = @_;
+# Set input data
+$indata ||= \%in;
```
jcameron, line 5504: I don't think an internal API function like this could ever read %in ... that's for use in CGI scripts

create-user.pl
iliajie: > Shouldn't virtualmin_htpasswd be quoted?

Well, not necessarily! It works just fine as long as it doesn't have spaces. Although, it isn't a problem to quote it. Fixed on the latest commit.

feature-web.pl
iliajie: Thanks, fixed!

feature-web.pl
iliajie: When it cannot read `%in` it means it's expected `$indata` to be passed.

modify-user.pl
```diff
@@ -508,11 +533,48 @@ package virtual_server;
...
-&modify_user($user, $olduser, $d);
+if ($user->{'extra'}) {
+	if ($user->{'type'} eq 'db') {
+		&modify_database_user($user, $olduser, $d);
+		&update_extra_user($d, $user, $olduser);
+		}
+	if ($user->{'type'} eq 'web') {
+		$user->{'pass'} = $pass if ($pass);
+		&modify_webserver_user($user, $olduser, $d,
+			{ virtualmin_htpasswd =>
```
jcameron: Why does this param need to be a hash if `virtualmin_htpasswd` is the only field that's ever set?

iliajie: Because `modify_webserver_user` expect to get and pass a hash, which can be this or `\%in`.

users-lib.pl
```diff
@@ -0,0 +1,214 @@
...
+        }
+return @dbusers;
+}
+
+# get_extra_db_user(&domain, username)
+# Returns a single extra database user
+sub get_extra_db_user
+{
+my ($d, $u) = @_;
+my @extra_db_users = &list_extra_db_users($d);
```
jcameron, line 110: Couldn't this just read the one extra user?

users-lib.pl
```diff
@@ -0,0 +1,214 @@
...
+my ($extra_db_user) = grep { $_->{'user'} eq $u } @extra_db_users;
+return $extra_db_user;
+}
+
+# list_extra_web_users(&domain)
+# Return a list of extra web users for some domain
+sub list_extra_web_users
+{
+my ($d) = @_;
+my @rv = &list_extra_users($d, 'web');
```
jcameron, line 120: No need to assign to @rv

users-lib.pl
```diff
@@ -0,0 +1,214 @@
...
+return @rv;
+}
+
+# get_extra_web_user(&domain, username)
+# Returns a single extra web user
+sub get_extra_web_user
+{
+my ($d, $u) = @_;
+my @extra_web_users = &list_extra_web_users($d);
```
jcameron, line 129: Same here, it would be better to just read the one user we want?

virtual-server-lib-funcs.pl
jcameron: If we removed automatic merging, would admins have to delete the existing DB user before creating a clashing user with the same permissions?

virtual-server-lib-funcs.pl
```diff
@@ -19667,6 +19880,154 @@ sub save_domain_ssh_pubkey
...
+		my $pubkeyfile = &transname('id_rsa.pub');
+		&write_file_contents($pubkeyfile, $pubkey);
+		&execute_command("$ssh_keygen -l -f $pubkeyfile", undef, \$ssh_keytest_out, \$ssh_keytest_err);
+		($ssh_keytest_err =~ s/\s*\Q$pubkeyfile\E\s*|^\s+|\s+$//g, $ssh_keytest_err =~ s/\.$//,
+			$ssh_keytest_err = &text('validate_esshpubkeyinvalid', ucfirst($ssh_keytest_err)))
+				if ($ssh_keytest_err);
+		}
+	else {
+		($ssh_keytest_err = &text('validate_esshpubkeyinvalid', $text{'validate_esshpubkeyinvalidformat'}))
+			if ($pubkey && $pubkey !~ /^(ssh-rsa|ssh-dss|ecdsa-sha2-nistp256|ecdsa-sha2-nistp384|ecdsa-sha2-nistp521|ssh-ed25519)/);
```
jcameron, line 20025: Is it necessary to hard-code all these key types? What if a new key type comes along?

feature-web.pl
jcameron: I understand, but that should never happen. Callers should always pass in a hash with the correct fields. Also, I think it's always a bad idea to pass %in directly, because that will contain ALL the URL parameters..

modify-user.pl
jcameron: But we control that function! We can make it take whatever args we want, and passing %in down to internal API functions is a bad idea.

users-lib.pl
iliajie: Yes, right. Fixed here https://github.com/virtualmin/virtualmin-gpl/commit/564c9cc0f565d9290dab16a8bb72c94194bf5693.

users-lib.pl
iliajie: Agreed! Fixed here https://github.com/virtualmin/virtualmin-gpl/commit/c3eeb13bd12d19b0b931a33a352fe2e5aaa59a60.

users-lib.pl
iliajie: Yes, it would. Fixed here https://github.com/virtualmin/virtualmin-gpl/commit/c177282738038675a36afbfac6902aef80d6db7e.

virtual-server-lib-funcs.pl
iliajie: No, if you remember there is now a confirmation screen:

image

The user will just work as expected after merging, no need to delete it beforehand. Also, using CLI it will just work without any confirmation.

virtual-server-lib-funcs.pl
iliajie: Well, this is mostly a fall back solution for old systems which don't have `ssh-keygen` command. In fact, we should never reach this code on any contemporary system.

feature-web.pl
iliajie: > I understand, but that should never happen. Callers should always pass in a hash with the correct fields.

Yes, actually I agree! This is fixed here https://github.com/virtualmin/virtualmin-gpl/pull/707/commits/890f63a8d688afc049ae92596496b30d88be71db.

> Also, I think it's always a bad idea to pass %in directly, because that will contain ALL the URL parameters..

This is just exactly what we want in our case, as it will be passed to Virtualmin Htpasswd plugin.

modify-user.pl
iliajie: Bad or not, but this is currently what the plugin function [expects](https://github.com/virtualmin/virtualmin-htpasswd/blob/master/virtual_feature.pl#L82-L86). We can fix it later though, if needed. It's unrelated to this PR.

virtual-server-lib-funcs.pl
jcameron: That's a bit confusing to the admin .. what does "Suppress" mean here? Will the extra user be merged, or just deleted?

virtual-server-lib-funcs.pl
jcameron: Right, but can't we check for this in another way that doesn't involve listing all the key types? What does the output contain in this fallback case?

virtual-server-lib-funcs.pl
iliajie: The extra user is the association in Virtualmin (text file under `users/`) to the actual database user. Virtualmin under the hood, when a new Unix user created, would technically just replaced the old database user with the new one, updating user's password and allowed database list. So technically we are replacing the existing database user.

feature-web.pl
jcameron: My apologies, I forgot that the function being called eventually really does need the URL params input. So this is fine..

modify-user.pl
jcameron: Yes, you're right! it's good then ...

virtual-server-lib-funcs.pl
iliajie: Well, the output is a public key, for example:

```
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQClNgCrdnsYVpMu6Ne6OFecTsGighwB/TgdM7b1W3Z+BHFyXlWAsqjAzdcNyeMP3o18Xst8X3QT/13ly/F9+U+KFvD73mE3QuQUWjO6jmo9CDrf+LZxF0CR+z/HCgBJbxeuCt0/F/xkKTyJdIJaTMYSMIuPHNJXtsVYgOJ1Fvzsq9/0BDUw9lovgoQlhS+ywpheBal7UsB802DXKpmFJQ/l5WJtlxmhNy9zy6Oi483ADoRdwN+1U9YLaw5Rl3bTscLhPj8HBFrEUyk/oJn2TpKoqNOJ2Rw1sFL8mkxHokNam5GSybLosEW9g8sFgteOkht8+uRY9HZqWVbtT6hZCxJP
```

The types are pretty standard. We could use more general regex but that wouldn't be accurate, like at all!

feature-web.pl
iliajie: Alright, sure! No problem.

virtual-server-lib-funcs.pl
jcameron: Maybe use the word "Merge" instead then

iliajie: > Maybe use the word "Merge" instead then

Alright, fixed here https://github.com/virtualmin/virtualmin-gpl/commit/744303eac6e139cebb9109b69607cd5207ab052a.


image

virtual-server-lib-funcs.pl
jcameron: What's the output if it fails though? Actually can we not just use $? to check for errors?

iliajie: > What's the output if it fails though?

It tells the user that passed public key is invalid, i.e.:

```
SSH public key is invalid: ssh-rsa AAAAB3NzaC1yc2EA
```

I just noticed that the user is getting saved prior validating the SSH key. It should be fixed.

iliajie: > I just noticed that the user is getting saved prior validating the SSH key. It should be fixed.

This is was fixed here https://github.com/virtualmin/virtualmin-gpl/commit/d6ec6dadfa52b448c3b0461ee143e1acd8e4b4ce.

jcameron: Wouldn't it be better to just check for the error message then?

iliajie: > Wouldn't it be better to just check for the error message then?

Jamie, there is no command output! This literally a check against plain `$in{'sshkey'}` but only if `ssh-keygen` command isn't available.

iliajie: > > Wouldn't it be better to just check for the error message then?
> 
> Jamie, there is no command output! This literally a check against plain `$in{'sshkey'}` but only if `ssh-keygen` command isn't available.

Jamie, I have refactored [this code block](https://github.com/virtualmin/virtualmin-gpl/pull/707/commits/710cbae74f6932d035b664dbab63967fe33cf418) for better readability and made regex more complete to include a few more known formats. However, this regex should never be reached in reality. It is just a fallback.

iliajie: To make more sense for regex part, I have [added a flag](https://github.com/virtualmin/virtualmin-gpl/pull/707/commits/c8635e8626d73039be65e165cdf6f943ad9e6faa) that will allow to only run regex validation if/when needed.

jcameron: Oh do you mean that this is a check against the user's input of an SSH key, not the output of a command?

Is that case I suppose it does make sense to check like this.

iliajie: > Oh do you mean that this is a check against the user's input of an SSH key, not the output of a command?

Yes.

> Is that case I suppose it does make sense to check like this.

Yes, it is. However, this check is a fall back in case we cannot validate passed SSH public key using `ssh-keygen -l -f public-key"` command.

iliajie:
Finally! We've done it! Thanks!

Issue #709, opened 2023-12-08
FCGI Wrap: False positive


iliajie:
Jamie, look, this code in Virtualmin:

```# If using fcgiwrap, make sure the server is running
if ($d->{'fcgiwrap_port'}) {
    my $st = &get_fcgiwrap_status($d);
    if ($st == 0) {
        return $text{'validate_efcgiwrapinit'};
        }
    elsif ($st == 1) {
        return $text{'validate_efcgiwraprun'};
        }
    }
```

.. produces an error in domain validation test right before LE certificate request, i.e.:

```
Validating configuration for example.com ..
.. errors were found, which will prevent Let's Encrypt from issuing a certificate :

Apache website : No FCGIwrap server bootup action exists!
```

However, my Debian 12 system only has `fcgiwrap`, `fcgiwrap.service`, and `fcgiwrap.timer` units. All of them are running correctly. The `get_fcgiwrap_status` function for `$name` returns for this domain `fcgiwrap-example-com` .. although, this service doesn't exist. However, any of this should be irrelevant, as my system has `suexec-custom` installed.

Apache config for this domain contained:

```
<Directory /home/example.com/cgi-bin>
    allow from all
    AllowOverride All Options=ExecCGI,Includes,IncludesNOEXEC,Indexes,MultiViews,SymLinksIfOwnerMatch
    Require all granted
    SetHandler proxy:unix:/var/fcgiwrap/1663060080110061.sock/socket|fcgi://localhost
    ProxyFCGISetEnvIf true SCRIPT_FILENAME "/home/example.com%{reqenv:SCRIPT_NAME}"
</Directory>
```

As I toggled Apache webserver feature, the config was regenerated, and now looks correct, i.e.:

```
<Directory /home/example.com/cgi-bin>
    Require all granted
    AllowOverride All Options=ExecCGI,Includes,IncludesNOEXEC,Indexes,MultiViews,SymLinksIfOwnerMatch
</Directory>
```

I assume it was a bug that we fixed recently, right?

Well, if so, we should also reset `fcgiwrap_port` option for the domain config, as it still keeps failing _**false-positively**_!

jcameron:
Is this happening when requesting an LE cert at domain creation, or every time you try to get a cert?

Also, did this domain every use `fcgiwrap` mode to execute CGI scripts? The `fcgiwrap_port` field shouldn't be set unless FCGI mode is in use..

iliajie:
[...]

PR #715, opened 2023-12-11
Fix to clean `fcgiwrap` domain config

iliajie:
Jamie, as discussed here:

https://github.com/virtualmin/virtualmin-gpl/issues/709

### DIFF
--- a/backups-lib.pl
+++ b/backups-lib.pl
@@ -2280,6 +2280,8 @@ if ($opts->{'reuid'}) {
 
 local $vcount = 0;
 local %restoreok;	# Which domain IDs were restored OK?
+local $enable_fcgiwrap_func;
+local $disable_fcgiwrap_func;
 if ($ok) {
 	# Restore any Virtualmin settings
 	if (@$vbs) {
@@ -2914,6 +2916,39 @@ if ($ok) {
 					}
 				}
 
+			# If domain had fcgiwrap enabled, check what's up
+			my $fcgiwrap_port;
+			map { $fcgiwrap_port = $_ } grep { /fcgiwrap_port$/ } keys %{$d};
+			if ($fcgiwrap_port) {
+				my $newweb = &domain_has_website($d);
+				my $enable_fcgiwrap;
+				my $can_fcgiwrap = &supports_fcgiwrap();
+				my ($fcgiwrap_port_plugin) = $newweb =~ /-(.*)$/;
+				my $fcgiwrap_port_dir =
+					($fcgiwrap_port_plugin ? "${fcgiwrap_port_plugin}_" : "").
+						'fcgiwrap_port';
+
+				# If not the same webserver, clean domain config
+				# directives first but mark for restore later
+				if ($newweb ne $oldweb) {
+					delete($d->{$fcgiwrap_port});
+					delete($d->{$fcgiwrap_port_dir});
+					$enable_fcgiwrap++;
+					}
+
+				# Need to preserv the state of fcgiwrap for domain as
+				# in original system if supported, otherwise disable
+				my ($fcgiwrap_plugin) = $newweb =~ /-(.*)$/;
+				my $fcgiwrap_plugin_name = "${newweb}::";
+				$fcgiwrap_plugin_name =~ s/-/_/g;
+				$fcgiwrap_plugin = 'apache', $fcgiwrap_plugin_name = undef
+					if (!$fcgiwrap_plugin);
+				$enable_fcgiwrap_func = "${fcgiwrap_plugin_name}enable_${fcgiwrap_plugin}_fcgiwrap"
+					if ($enable_fcgiwrap && $can_fcgiwrap);
+				$disable_fcgiwrap_func = "${fcgiwrap_plugin_name}disable_${fcgiwrap_plugin}_fcgiwrap"
+					if (!$can_fcgiwrap);
+				}
+
 			# Finally, create it
 			&$indent_print();
 			delete($d->{'missing'});
@@ -2943,35 +2978,6 @@ if ($ok) {
 					$d->{'disabled_reason'},
 					$d->{'disabled_why'});
 				&$outdent_print();
-				}
-			
-			# If domain had fcgiwrap enabled, check what's up
-			my $fcgiwrap_port;
-			map { $fcgiwrap_port = $_ } grep { /fcgiwrap_port$/ } keys %{$d};
-			if ($fcgiwrap_port) {
-				my $newweb = &domain_has_website($d);
-				my $disable_fcgiwrap = !&supports_fcgiwrap() || &supports_suexec();
-				my $same_web = $newweb eq $oldweb;
-				my ($fcgiwrap_port_plugin) = $newweb =~ /-(.*)$/;
-				my $fcgiwrap_port_dir =
-					($fcgiwrap_port_plugin ? "${fcgiwrap_port_plugin}_" : "").
-						'fcgiwrap_port';
-				# If not the same webserver, flip and clean domain config directives
-				if (!$same_web) {
-					$d->{$fcgiwrap_port_dir} = $d->{$fcgiwrap_port}
-						if (!$d->{$fcgiwrap_port_dir});
-					delete($d->{$fcgiwrap_port});
-					}
-				# Need to disable fcgiwrap for domain
-				if ($disable_fcgiwrap) {
-					my ($fcgiwrap_plugin) = $newweb =~ /-(.*)$/;
-					my $fcgiwrap_plugin_name = "${newweb}::";
-					$fcgiwrap_plugin_name =~ s/-/_/g;
-					$fcgiwrap_plugin = 'apache', $fcgiwrap_plugin_name = undef
-						if (!$fcgiwrap_plugin);
-					my $cfunc = "${fcgiwrap_plugin_name}disable_${fcgiwrap_plugin}_fcgiwrap";
-					&$cfunc($d) if (defined(&$cfunc));
-					}
 				}
 			}
 		else {
@@ -3084,6 +3090,22 @@ if ($ok) {
 			# Make site the default if it was before
 			if ($d->{'web'} && $d->{'backup_web_default'}) {
 				&set_default_website($d);
+				}
+
+			# If fcgiwrap can be used on this system,
+			# and restored domain had fcgiwrap enabled,
+			# but the webserver is different, then
+			# reenable fcgiwrap for the domain
+			if ($enable_fcgiwrap_func) {
+				&$enable_fcgiwrap_func($d)
+					if (defined(&$enable_fcgiwrap_func));
+				}
+
+			# If the system doesn't have fcgiwrap support
+			# at all and webservers are the same
+			if ($disable_fcgiwrap_func) {
+				&$disable_fcgiwrap_func($d)
+					if (defined(&$disable_fcgiwrap_func));
 				}
 
 			# Run the post-restore command
--- a/backups-lib.pl
+++ b/backups-lib.pl
@@ -2949,7 +2949,7 @@ if ($ok) {
 			my $fcgiwrap_port;
 			map { $fcgiwrap_port = $_ } grep { /fcgiwrap_port$/ } keys %{$d};
 			if ($fcgiwrap_port) {
-				my $newweb = &domain_has_website();
+				my $newweb = &domain_has_website($d);
 				my $disable_fcgiwrap = !&supports_fcgiwrap() || &supports_suexec();
 				my $same_web = $newweb eq $oldweb;
 				my ($fcgiwrap_port_plugin) = $newweb =~ /-(.*)$/;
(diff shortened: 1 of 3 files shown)


iliajie:
> Where is refresh_domain_fcgiwrap defined?

It's right here: https://github.com/virtualmin/virtualmin-gpl/pull/715/files#diff-a8b0a88367118c8f13666eed07287b13ecc217c8fdfaf6a34d51b226dc196f92R11400

jcameron:
Oh I see ...  not sure why I missed it when I first looked at the change.

But I don't see what is going on here? How does this code do anything different from removing fcgiwrap_port ?

jcameron:
I don't think this is the right fix though - instead we should validate the fcgiwrap / suexec setting when the domain is being restored, and make sure it is compatible with the local system. I'll work on a code change for that..

iliajie:
> I don't think this is the right fix though - instead we should validate the fcgiwrap / suexec setting when the domain is being restored, and make sure it is compatible with the local system.

I think I understand now the source of misunderstanding and confusion, as I forgot to mention an important detail. The domain config can have `nginx_fcgiwrap_port` and restored on Apache system with `suexec` support. Apache related code knows nothing about `nginx_fcgiwrap_port`.

As I ran more tests, I think, a better approach would probably be to call `disable_apache_fcgiwrap` and similar function in Nginx for systems which have `suexec` support, however both of those functions should be aware how to clean each other `nginx_fcgiwrap_port` and `fcgiwrap_port` directives upon disabling! This is exactly what my function in this PR is doing. We will need to call it or its logic in the right place.

I understand now why this patch is incomplete. I was testing by restoring Nginx backup with `fcgiwrap` enabled on Apache system. In this case webserver config was just regenerated. After creating Apache backup with `fcgiwrap` and restoring it on Apache system with `suexec` support, I can see that we just copy Apache config to the new system as is, which contains `fcgiwrap` socket configuration. That said, we cannot let this happen and have to do what I have described above, i.e. call disable function which will clear both `nginx_fcgiwrap_port` and `fcgiwrap_port` directives, and properly disable `fcgiwrap`.

I will try to improve my PR later today ..

iliajie:
Jamie, I have reverted previous fixes and [implemented the new](https://github.com/virtualmin/virtualmin-gpl/pull/715/files#diff-70ef8c4868cf7db663c021acf70e7343259956ad2339f096fe6bc81448319a38) fix. Please give it a look and try.

We will need to factor out the existing logic on Nginx side into a separate `disable_nginx_fcgiwrap` sub.

Better this time?

backups-lib.pl
```diff
@@ -2944,6 +2944,35 @@ sub restore_domains
 					$d->{'disabled_why'});
 				&$outdent_print();
 				}
+			
+			# If domain had fcgiwrap enabled, check what's up
+			my $fcgiwrap_port;
+			map { $fcgiwrap_port = $_ } grep { /fcgiwrap_port$/ } keys %{$d};
```
jcameron, line 2950: Can't you just do `$fcgiwrap_port = $d->{'fcgiwrap_port'};`

iliajie: > Can't you just do `$fcgiwrap_port = $d->{'fcgiwrap_port'};`

No, because Nginx has it set to `$d->{'nginx_fcgiwrap_port'}` ..

jcameron: Oh so in that case, shouldn't this code to fix the port be in feature-web.pl and the Nginx module? Then it knows which field needs fixing..

iliajie: > Oh so in that case, shouldn't this code to fix the port be in feature-web.pl and the Nginx module? Then it knows which field needs fixing..

No, because if you restore Nginx backup on Apache system, `feature-web.pl` code won't know anything about `nginx_fcgiwrap_port` record ..

jcameron: So in that case it would just re-setup the webserver config from scratch using the local defaults, so this won't be an issue.

iliajie: > So in that case it would just re-setup the webserver config from scratch using the local defaults, so this won't be an issue.

It will be an issue, because domain config will contain `nginx_fcgiwrap_port` on Apache systems and `fcgiwrap_port` on Nginx. Also, we aren't looking to have it re-setup - we're expecting to have it disabled.

iliajie: Jamie, I have taken a fresh and deep look into the problem over again!

I realized that we should preserve `fcgiwrap` mode even if `suexec` is available, because we do support it in CLI.

The [final latest](https://github.com/virtualmin/virtualmin-gpl/pull/715/commits/1d476955910458634ce544d9c165330fb783f33e) patch will re-enable `fcgiwrap` if virtual server in backup had it.

If webserver is different from what it was on the backup, then we clean domain config from any `fcgiwrap` related directives before restore, and mark it for re-enabling this feature after the restore. When domain restore happened we just re-enable `fcgiwrap` for migrated domain if system supports it.

If webserver is the same from what was on the backup, and because we copy configs directly, we don't touch it at all in case target system has `fcgiwrap` support. If it isn't, we disable `fcgiwrap` for the restored domain after the restore.

To test this feature I used 4 different instances. Two Almas and two Debians with both Apache and Nginx stacks.

backups-lib.pl
```diff
@@ -2914,6 +2916,39 @@ sub restore_domains
...
+				# directives first but mark for restore later
+				if ($newweb ne $oldweb) {
+					delete($d->{$fcgiwrap_port});
+					delete($d->{$fcgiwrap_port_dir});
+					$enable_fcgiwrap++;
+					}
+
+				# Need to preserv the state of fcgiwrap for domain as
+				# in original system if supported, otherwise disable
+				my ($fcgiwrap_plugin) = $newweb =~ /-(.*)$/;
```
jcameron, line 2941: I really think this code should live in feature-web.pl and the Nginx plugin, so that all this complexity of computing the function to call can be moved to the place where the names are known.

iliajie: Jamie, how do we handle this situation in `feature-web` alone, if a back up file can provide both `nginx_fcgiwrap_port` and `fcgiwrap_port` .. we need to know which sub to call depending on the stack.

jcameron: Only one webserver feature will be enabled for the domain though (apache or nginx), so only one of the restore functions will be called. If the other feature's port field is still leftover in the domain config, it doesn't matter..

99.9% of the time the user will be restoring on a system with the same webserver. And in the other 1%, fcgiwrap mode will already be setup using the local default from the template..

iliajie: Maybe but I have personally faced this issue, it is very annoying and most importantly mystifying issue! Not everyone preserves the patterns and restores on the same OS. I'd say the number would be much lower than 99.9%.

The point is to make it work smoothly for the users.  Adding these few lines of code raises complexity a bit, although it is better than restoring a domain that users won't be able to fix on their own. It isn't enough to toggle webserver feature. You'd need to delete domain completely to fix it or know which domain config option is causing it.

I'd suggest giving this PR a real-life try and see. Did something not work for you?

jcameron: So when migrating from Apache to Nginx, it should already work fine without any patch though! Because the different webservers use different fields in the domain config for the fcgiwrap port, and so there should be no clash or failure when restoring on a new system.

iliajie: Yes, because webserver config is regenerated. However, Apache to Nginx migration will leave `fcgiwrap_port` left over in domain  config and mat fail some tests. This PR addresses it as well.

jcameron: Wait what tests will that fail? Are you thinking of the case where a domain is migrated from Apache to Nginx, and back again?

iliajie: > Wait what tests will that fail? Are you thinking of the case where a domain is migrated from Apache to Nginx, and back again?

Any `if` tests for `fcgiwrap_port`.

jcameron: I fixed up a few cases where this was being checked even when Apache is not enabled here : https://github.com/virtualmin/virtualmin-gpl/commit/23a2cd01c251ad41c4066bf44233eb0b72bfb33e

iliajie: > I fixed up a few cases where this was being checked even when Apache is not enabled here : [23a2cd0](https://github.com/virtualmin/virtualmin-gpl/commit/23a2cd01c251ad41c4066bf44233eb0b72bfb33e)

Thanks, Jamie! You'd also want to make similar change to `validate_web` sub! However, this isn't a real fix and just a work-around.

I understand the sentiment, notwithstanding sometimes the simplest solutions just aren't available, no matter how much we both wish for them &mdash; your patch won't address the problem where migration happens from  Apache ↔ Apache system, where `fcgiwrap` was used previously, for example in case of Rocky to Debian migration.

jcameron: Right, my patch on it's own isn't a complete fix. We still need code in `restore_web` that checks if fcgiwrap_port is set but the new system doesn't support fcgiwrap, or conversely if suexec was in use and the system doesn't support suexec. I just think that `restore_web` is the right place for this..

iliajie: > I just think that `restore_web` is the right place for this

If we did this in `restore_web` then we would have to re-implement the same or similar logic in `feature_restore`. I'd prefer have it done in one place ..

jcameron: Yes, but I think that's the right choice - the logic would be simpler and easier to read, as there's no need to have all that code to build the correct function name to call.

iliajie: .. and how would `restore_web` deleted `nginx_fcgiwrap_port` directive then .. 🙂 I mean building function name isn't the only problem ..

jcameron: Does it need to remove that field? It doesn't do anything if Nginx isnt' in use ..

iliajie: Probably not, although `fcgiwrap_port` could in case of Nginx is in use.

It seems to me that splitting code is just going to add more complexity .. If we ever add support for another webserver, we would have to re-implement all the same (neutral) logic over again.

jcameron: Yeah but that's a price worth paying IMHO

iliajie: > Yeah but that's a price worth paying IMHO

Alright, I can do that! However at first, I'd like to understand what goodies are we getting for paying more? 🙂

jcameron: If we add another webserver type in future, all the logic for it is contained inside that webserver's plugin code.

iliajie: I have re-checked the code and it wouldn't be trivial (not at all) to fix it the way you suggest, because the code has to be called at different stages of `restore_domains`, and it depends on its logic.

> If we add another webserver type in future, all the logic for it is contained inside that webserver's plugin code.

It definitely would have a benefit in case the code base for this feature was anyhow different/specific. In this particular case, the logic is identical for all &mdash; we shouldn't break DRY principle.

.. oh, I think I understand the source of confusion now. Have you rechecked recently committed code in this PR? Perhaps you were referring to the code in my initial commits, which I later changed/improved.

Could you please review it?

jcameron: I still see calls to a per-webserver function referenced by `enable_fcgiwrap_func` in the patch?

Also, in Nginx there is no support for suexec mode - fcgiwrap is just enabled always if supported by the system, and otherwise there is no way to run CGI scripts. But in Apache, you can either use suexec, fcgiwrap or neither in some cases. So the logic for handling this during a restore is quite different ..

iliajie: > I still see calls to a per-webserver function referenced by `enable_fcgiwrap_func` in the patch?

Yes, correct, because this logic is built before the domain is actually recreated, and feature itself is getting changed way after, and only if needed (conditionally). `restore_domains` is a general type of function, it should be just fine to have such logic in it.

> Also, in Nginx there is no support for suexec mode - fcgiwrap is just enabled always if supported by the system, and otherwise there is no way to run CGI scripts.

Yes, and we still need to handle it correctly when migrating from a system that support `fcgiwrap` to a system which isn't.

> So the logic for handling this during a restore is quite different

Yes, and my PR considers logic for all mentioned cases!

> But in Apache, you can either use suexec, fcgiwrap or neither in some cases.

Yes, I'm aware of that.

Functions like `enable_fcgiwrap_func` and `disable_fcgiwrap_func` should be created in Nginx side to work similarly to Apache.

It must be possible to disable `fcgiwrap` independently using Virtualmin CLI on Nginx system too. It has to be consistent.

jcameron: You know, what I think we really need is a proper API to enable or disable suexec and list supported modes that calls either the plugin or Apache function. Let me look into that ...

iliajie: Great idea. Although, what does it have to do with the existing bug?

jcameron: It would make this code a lot cleaner. Let's hold off on this PR until I have that ready..

iliajie: Alright, sure! Thanks!

PR #717, opened 2023-12-17
Fix inconsistent output from list-domains --simple-multiline and --multiline

bstramsek:
Use different variable for php_mode to not overwrite has website check (from line 583).
The $p still needed to be website check for php max execution time print.

https://forum.virtualmin.com/t/inconsistent-cli-output-for-virtualmin-list-domains-simple-multiline-vs-multiline/123989


iliajie:
Thanks for the patch.

PR #744, opened 2024-02-01
Fix to check if resulting password meets requirements

iliajie:
Jamie, this PR address the problem discussed [here](https://forum.virtualmin.com/t/virtualmin-password-generator-is-not-including-sometimes-digits-or-upper-letters/124604/4?u=ilia).

In this case it would be better not to use a new `&substitute_pattern()` because required password option isn't a regex.

### DIFF
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -11086,22 +11086,25 @@ my $has_uppercase = sub { return $_[0] =~ /[A-Z]/; };
 my $has_lowercase = sub { return $_[0] =~ /[a-z]/; };
 my $has_digit = sub { return $_[0] =~ /[0-9]/; };
 my $has_special = sub { return $_[0] =~ /[\@\#\$\%\^\&\*\(\)\_\+\!\-\=\[\]\{\}\;\:\'\"\,\<\.\>\/\?\~\`\\]/; };
+my $has_unicode_chars = sub { return $_[0] =~ /(?![A-Za-z])\p{L}/; };
 # Dereference $chars ref only once
 my @chars = @$chars;
 # Determine the required character types based on $chars
-my $complexity_level = 0;
-$complexity_level++ if grep { $has_uppercase->($_) } @chars;
-$complexity_level++ if grep { $has_lowercase->($_) } @chars;
-$complexity_level++ if grep { $has_digit->($_) } @chars;
-$complexity_level++ if grep { $has_special->($_) } @chars;
+my %complexity_levels;
+$complexity_levels{'has_uppercase'} = 1 if grep { $has_uppercase->($_) } @chars;
+$complexity_levels{'has_lowercase'} = 1 if grep { $has_lowercase->($_) } @chars;
+$complexity_levels{'has_digit'} = 1 if grep { $has_digit->($_) } @chars;
+$complexity_levels{'has_special'} = 1 if grep { $has_special->($_) } @chars;
+$complexity_levels{'has_unicode_chars'} = 1 if grep { $has_unicode_chars->($_) } @chars;
 # Check if the number of required types exceeds the password length
-return -1 if ($complexity_level > $len || $complexity_level == 0);
+return -1 if (scalar(keys %complexity_levels) > $len);
 # Check if $pass contains at least one of each required type
 my $valid = 1;
-$valid = 0 if $complexity_level >= 1 && !grep { $has_uppercase->($pass) } @chars;
-$valid = 0 if $complexity_level >= 2 && !grep { $has_lowercase->($pass) } @chars;
-$valid = 0 if $complexity_level >= 3 && !grep { $has_digit->($pass) } @chars;
-$valid = 0 if $complexity_level >= 4 && !grep { $has_special->($pass) } @chars;
+$valid = 0 if $complexity_levels{'has_uppercase'} && !grep { $has_uppercase->($pass) } @chars;
+$valid = 0 if $complexity_levels{'has_lowercase'} && !grep { $has_lowercase->($pass) } @chars;
+$valid = 0 if $complexity_levels{'has_digit'} && !grep { $has_digit->($pass) } @chars;
+$valid = 0 if $complexity_levels{'has_special'} && !grep { $has_special->($pass) } @chars;
+$valid = 0 if $complexity_levels{'has_unicode_chars'} && !grep { $has_unicode_chars->($pass) } @chars;
 return $valid;
 }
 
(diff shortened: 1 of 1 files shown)


virtual-server-lib-funcs.pl
```diff
@@ -11067,9 +11067,39 @@ sub random_password
 foreach (1 .. $len) {
 	$random_password .= $passwd_chars[rand(scalar(@passwd_chars))];
 	}
+# Check resulting password and try again
+# if it doesn't meet the requirements
+return &random_password(@_)
```
jcameron, line 11072: Nice, but why not just use a loop rather than recursion? It would be easier to read ...

iliajie: Alright, how about that https://github.com/virtualmin/virtualmin-gpl/pull/744/commits/e608c275d693e4c556915633d570de4e09f91f02.

jcameron: Looks better, thanks!

iliajie: Alright, cool! Thanks!

If there is anything else, please let me know, otherwise let's just merge it.

virtual-server-lib-funcs.pl
```diff
@@ -11058,18 +11058,53 @@ sub random_password
...
+my ($chars, $len, $pass) = @_;
+my $has_uppercase = sub { return $_[0] =~ /[A-Z]/; };
+my $has_lowercase = sub { return $_[0] =~ /[a-z]/; };
+my $has_digit = sub { return $_[0] =~ /[0-9]/; };
+my $has_special = sub { return $_[0] =~ /[\@\#\$\%\^\&\*\(\)\_\+\!\-\=\[\]\{\}\;\:\'\"\,\<\.\>\/\?\~\`\\]/; };
+# Dereference $chars ref only once
+my @chars = @$chars;
+# Determine the required character types based on $chars
+my $complexity_level = 0;
+$complexity_level++ if grep { $has_uppercase->($_) } @chars;
```
jcameron, line 11093: I haven't seen this `$function->(args)` syntax before. Is it new?

iliajie: It's very very old and more preferred over `&$function()` use.

virtual-server-lib-funcs.pl
```diff
@@ -11058,18 +11058,53 @@ sub random_password
...
+		$random_password .= $passwd_chars[rand($passwd_chars)];
+		}
+	return $random_password
+		if (&random_password_validate(
+			\@passwd_chars, $len, $random_password) != 0);
 	}
 return $random_password;
 }
 
+sub random_password_validate {
```
jcameron, line 11083: Can you explain the logic that this function is doing? I am confused as to why it's setting the complexity_level based on the has_ functions, but later using those same functions to check the complexity_level

iliajie: > Can you explain the logic that this function is doing? 

Yes, sure!

> I am confused as to why it's setting the complexity_level based on the has_ functions, but later using those same functions to check the complexity_level

`@passwd_chars` is not static value. It can contain different type of characters. We test which types of characters the source `@passwd_chars` has and remember it. If `@passwd_chars` has capital letters, than later we need to make sure that the resulting password also has it. The same logic applies to lowercase, digits and special chars.

We just test what basic type of characters given `@passwd_chars` has and later test against exact types in the generated password.

So, if `@passwd_chars` is default `A-Z,a-z,0-9` then the password complexity will be `0 + 3`, which impose resulting password also be level 3 of complexity. So, if a generated password happens to be alphabetical, instead of  being alphanumerical (i.e. not as in default `@passwd_chars`) then we try again.

We could add more levels of complexity. For example to test for Greek or Cyrillic chars or any other.

jcameron: Ah, so it's making sure that all the character types in the list of candidate chars are also in the generated password?

iliajie: Yes, exactly!

iliajie: Though, I'm thinking now that different type of complexity levels could "contribute" to the different types, as we're just adding `++` to each level type.

I will fix that ..

iliajie: Alright, Jamie it's fixed here https://github.com/virtualmin/virtualmin-gpl/pull/744/commits/920532393760aca5c2e5fd293513c88d6e9b6dc1.

virtual-server-lib-funcs.pl
```diff
@@ -11058,18 +11058,54 @@ sub random_password
...
+my %check = (
+	uppercase => qr/[A-Z]/,
+	lowercase => qr/[a-z]/,
+	digit     => qr/[0-9]/,
+	special   => qr/[\@\#\$\%\^\&\*\(\)\_\+\!\-\=\[\]\{\}\;\:\'\"\,\<\.\>\/\?\~\`\\]/,
+	unicode   => qr/(?![A-Za-z])\p{L}/,
+	);
+# Determine required character types
+my %required = map {
+	my $__ = $_;
```
jcameron, line 11098: Is $__ a special perl variable?

virtual-server-lib-funcs.pl
```diff
@@ -11058,18 +11058,54 @@ sub random_password
...
+# Define character type checks in a hash
+my %check = (
+	uppercase => qr/[A-Z]/,
+	lowercase => qr/[a-z]/,
+	digit     => qr/[0-9]/,
+	special   => qr/[\@\#\$\%\^\&\*\(\)\_\+\!\-\=\[\]\{\}\;\:\'\"\,\<\.\>\/\?\~\`\\]/,
+	unicode   => qr/(?![A-Za-z])\p{L}/,
+	);
+# Determine required character types
+my %required = map {
```
jcameron, line 11097: This logic is hard to follow! Can you expand it to use a loop, even if it's more lines of code?

virtual-server-lib-funcs.pl
iliajie: No, just a variable.

virtual-server-lib-funcs.pl
iliajie: > This logic is hard to follow! Can you expand it to use a loop, even if it's more lines of code?

You're right. Fixed.

virtual-server-lib-funcs.pl
jcameron: Probably best to give it a regular variable name in future, as a casual reader might think that $__ is special like $_

virtual-server-lib-funcs.pl

PR #745, opened 2024-02-04
Add to support permalinks with Nginx in WordPress

iliajie:
Jamie, as discussed earlier, this PR will add and remove location directive to support permalinks in WordPress with Nginx.

### DIFF
--- a/scripts/wordpress.pl
+++ b/scripts/wordpress.pl
@@ -411,8 +411,11 @@ my ($d, $opts) = @_;
 if (&domain_has_website($d) eq 'virtualmin-nginx' &&
     &indexof('virtualmin-nginx', @plugins) >= 0) {
 	
-	my $locdir = $opts->{'path'}, $locdir =~ s/\/$//, $locdir ||= '/';
-	my $locdirarg = $locdir; $locdirarg .= '/' if ($locdirarg !~ /\/$/);
+	my $locdir = $opts->{'path'};
+	$locdir =~ s/\/$//;
+	$locdir ||= '/';
+	my $locdirarg = $locdir;
+	$locdirarg .= '/' if ($locdirarg !~ /\/$/);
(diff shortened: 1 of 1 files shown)


scripts/wordpress.pl
```diff
@@ -398,4 +404,114 @@ sub script_wordpress_passmode
...
+}
+
+sub script_wordpress_webserver_delete_records
+{
+my ($d, $opts) = @_;
+# Remove Nginx webserver previously added records
+if (&domain_has_website($d) eq 'virtualmin-nginx' &&
+    &indexof('virtualmin-nginx', @plugins) >= 0) {
+	my $locdir = $opts->{'path'}, $locdir =~ s/\/$//, $locdir ||= '/';
+	my $locdirarg = $locdir; $locdirarg .= '/' if ($locdirarg !~ /\/$/);
```
jcameron, line 479: Split these onto separate lines

iliajie: Alright, done!

scripts/wordpress.pl
```diff
@@ -398,4 +404,120 @@ sub script_wordpress_passmode
...
+				'members' => [
+					{ 'name' => 'try_files',
+					  'words' => [ '$uri',
+						       '$uri/',
+						       ($locdirarg.'index.php?$args') ]}]};
+			&virtualmin_nginx::save_directive($server, [ ], [ $loc ]);
+			}
+		&virtualmin_nginx::flush_config_file_lines();
+		&virtualmin_nginx::unlock_all_config_files();
+		&virtualmin_nginx::apply_nginx();
```
jcameron, line 470: Instead of directly applying the Nginx config here, use : 

`&register_post_action(\&virtualmin_nginx::print_apply_nginx);`

so it's run once at the end of the script install.

iliajie: It actually works! Thanks! Fixed.

PR #746, opened 2024-02-05
Fix to call incremental differential

iliajie:
Jamie, this PR fixes inconsistency in terminology.


jcameron:
I think I prefer the current terminology ... it's what we've been using for years, and what the `tar` command calls it.

iliajie:
Well, I understand that. However, the new docs will use "differential" backups term. Though it isn't hard to change it back to "incremental".

But it isn't right to do, I don't think, because in practice we're creating differential backups, not incremental.

`tar` is actually doing it right, using a snapshot file (full backup) and creating a new incremental. For `tar` calling it incremental is correct. For Virtualmin, it isn't, as it's a differential backup.

[Read more..](https://en.wikipedia.org/wiki/Differential_backup#:~:text=A%20differential%20backup%20refers%20to,since%20the%20last%20incremental%20backup.)

jcameron:
Which docs are you referring to?

iliajie:
> Which docs are you referring to?

The [new](https://virtualmin.dev/docs/data-operations/how-to-backup-virtual-servers/#differential-backups) docs.

To make it right we will mention this change in the upcoming version change log, as otherwise we will keep confusing knowledgeable users about the type of created backups, and users who is unaware about the differences wouldn't care either. A win-win situation in my opinion.

Issue #751, opened 2024-02-07
List Virtual Servers - the settings cog does not go to the correct section


shoulders:
|  SYSTEM INFORMATION||
|----------------------|---------------------------|
|  OS type and version |     Ubuntu Linux 22.04.3     |
|  Webmin version      |     2.105      |
|  Usermin version     |     2.005     |
|  Virtualmin version  |     7.9.0       |
|  Theme version       |     21.09.5      |
|  Package updates     |     All installed packages are up to date   |

### The Issue

When you click on the settings cog on the `List Virtual Servers` page you are taken to the following page and section.

https://server.example.com:10000/config.cgi?virtual-server&xnavigation=1

![image](https://github.com/virtualmin/virtualmin-gpl/assets/319997/ccf656e8-faf0-44c4-8314-b211f3a5e875)

This is the right page, wrong section

### Solution

Update the cog's URL to point to this section so the relevant display/column settings are immediately available

https://server.example.com:10000/config.cgi?module=virtual-server&section=line1.3&search=&nnext=Next&xnavigation=1

![image](https://github.com/virtualmin/virtualmin-gpl/assets/319997/ea44866b-8b84-40e4-8f73-3819e8629951)

iliajie:
Hello,

Well, what you're suggesting is reasonable, however this change would introduce inconsistency with other modules, where index page of a module leads to the module configuration (not specific page of module configuration).

I'm not confident about changing it, although, wouldn't oppose it either.

We should wait for Jamie's comment on it.

jcameron:
I think in this case since there's a section specific to that page, linking to it directly isn't a bad idea. Ilia, if you want to implement this, please send me a PR.

iliajie:
Alright, will do ..

iliajie:
> I think in this case since there's a section specific to that page, linking to it directly isn't a bad idea. Ilia, if you want to implement this, please send me a PR.

Check this [PR](https://github.com/virtualmin/virtualmin-gpl/pull/761).

iliaross:
Reopened, as this was closed by an automated cleanup run. @shoulders, sorry for the noise, and thanks for the contribution. We definitely didn't mean any disrespect.

If it turns out the issue is already resolved for you, you're welcome to close it.

PR #761, opened 2024-02-25
Fix config link to lead to user interface section

iliajie:
Hi Jamie! As discussed here https://github.com/virtualmin/virtualmin-gpl/issues/751.

### DIFF
--- a/index.cgi
+++ b/index.cgi
@@ -32,17 +32,8 @@ if ($single_domain_mode) {
 	exit;
 	}
 
-&ui_print_header(undef, $text{'index_title'}, "", "index", 1, 1, 0,
-	undef, undef, undef, $vtitle);
-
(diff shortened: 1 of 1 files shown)


iliajie:
I'd say, _no_,  as it would be more complicated and involve much more changes.

There is nothing wrong in modifying DOM, considering it's simple, works with all themes and doesn't create any theme-specific dependency.

iliajie:
However, it would work if we could re-use existing config param in `header` sub, i.e. `$_[3]` and if set to a string (rather than `1`) then add it as part of a query..

iliajie:
Okay, I have changed to pass params directly.

Depended PR is here https://github.com/webmin/webmin/pull/2098.

jcameron:
Thanks for these patches!

Issue #758, opened 2024-02-19
Backups not incremental for chrooted domains


ghost:
Hi,

Since the release of Virtualmin version 7.9.0, incremental backups for chrooted domains have been observed to be the same size as full backups. Further information can be found in the following two forum threads:

[Backups not incremental for chrooted domains](https://forum.virtualmin.com/t/backups-not-incremental-for-chrooted-domains/124900)
[Incremental backup same size as full backup](https://forum.virtualmin.com/t/incremental-backup-same-size-of-full-backup/124790)

Thank you.

jcameron:
Yes, it looks like the issue is this chmod command : https://github.com/virtualmin/virtualmin-gpl/blob/master/jailkit-lib.pl#L299

@iliajie you added this code recently .... do we really need that chmod?

iliajie:
> Yes, it looks like the issue is this chmod command : https://github.com/virtualmin/virtualmin-gpl/blob/master/jailkit-lib.pl#L299

`chown` does indeed update file **Change** time (i.e. `ctime`). Although I don't understand why Virtualmin doesn't rely on using **Modify** time (i.e. `mtime`) instead? It doesn't seem right to use `ctime` for this purpose. To be clear, using `chmod` doesn't change `mtime`, only `ctime`.

> you added this code recently .... do we really need that chmod?

Well, I added `-R` flag, to the [original code](https://github.com/virtualmin/virtualmin-gpl/commit/304991d9406cf819a35e9b15ce06b92257709e26), which would affect files under `~/virtualmin-backup`. Sadly, I don't remember why exactly I had to add `-R` flag but it couldn't be by accident.

jcameron:
It's really the tar command that looks at those fields, not Virtualmin. 

I think we should remove that chmod command because I don't see why it's needed. Or maybe only do it when files are first copied?

iliajie:
> It's really the tar command that looks at those fields, not Virtualmin.

I have just checked `backups-lib.pl` and we don't actually use `[10]` anywhere from the `stat`. It's always `[9]`, so you're right, it's not Virtualmin. However, `tar` also relies on `mtime`, right? So how do we get this issue just from running `chmod` over backups, which doesn't change `mtime`?

> I think we should remove that chmod command because I don't see why it's needed.

[...]

PR #763, opened 2024-02-25
Add API to remove write permissions for group

iliajie:
Jamie, this PR addresses [this issue](https://github.com/virtualmin/virtualmin-gpl/issues/758#issuecomment-1955946370).

### DIFF
--- a/jailkit-lib.pl
+++ b/jailkit-lib.pl
@@ -324,7 +324,7 @@ foreach my $zdir ("/usr/share/zoneinfo") {
 	}
 
 # Remove write permissions for group
-&remove_write_permissions_for_group($dir, "home/$home_virtualmin_backup");
+&remove_write_permissions_for_group($dir, "$dir/home/$home_virtualmin_backup");
 
 $d->{'jail_last_copy'} = time();
 return undef;
--- a/jailkit-lib.pl
+++ b/jailkit-lib.pl
@@ -324,7 +324,7 @@ foreach my $zdir ("/usr/share/zoneinfo") {
 	}
 
 # Remove write permissions for group
-&remove_write_permissions_for_group($dir, "$dir/home/$home_virtualmin_backup");
+&remove_write_permissions_for_group($dir, "$dir$d->{'home'}/$home_virtualmin_backup");
 
 $d->{'jail_last_copy'} = time();
 return undef;
--- a/jailkit-lib.pl
+++ b/jailkit-lib.pl
@@ -296,7 +296,6 @@ foreach my $sect ("perl", "basicshell", "extendedshell", "ssh", "scp", "sftp",
 	if ($?) {
 		return &text('jailkit_einit', $err);
 		}
-	&system_logged("chmod -R g-w ".quotemeta($dir)."/*");
 	}
 
 # Use jk_cp to copy in any other files/directories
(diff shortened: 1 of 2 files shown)


security-lib.pl
```diff
@@ -599,6 +600,26 @@ sub set_filepath_permissions_as_domain_user
...
+opendir(my $dh, $dir) || die("Cannot open directory $dir: $!");
+my @entries = readdir($dh);
+closedir($dh); 
+foreach my $entry (@entries) {
+	next if ($entry eq '.' || $entry eq '..');
+	my $path = "$dir/$entry";
+	next if ($exclude && $path =~ /\Q$exclude\E/);
+	my $mode = (stat($path))[2];
+	$mode &= ~S_IWGRP;
+	chmod($mode, $path) || warn("Failed to change permissions for $path: $!");
```
jcameron, line 618: What does warn() do here exactly?

jcameron: Also, do we really want to call `chmod` even if no permssion change is needed?

iliajie: `warn()` should just not fail and print to `miniserv.error` log instead.

We want to remove exec permissions from the group right?

jcameron: Ok, using warn seems fine.

What I meant was, is there any point calling `chmod` if the permissions are already correct? Because this will update the timestamp ...

iliajie: Oh, you're right. Fixed on the latest commit.

jailkit-lib.pl
```diff
@@ -325,6 +323,9 @@ sub copy_jailkit_files
 	&copy_source_dest($zdir, $dir.$zdir);
 	}
 
+# Remove write permissions for group
+&remove_write_permissions_for_group($dir, "home/$home_virtualmin_backup");
```
jcameron, line 327: What is `home/` in this exclude parameter?

iliajie: `home/` refers to the portion of _home_ in `/home/chroot/168953125698899/home/` that precedes `virtualmin-backups` directory.

Probably it would be better to give the full/exact path?

iliajie: .. I mean it shouldn't fail but could if Virtualmin backups directory is set to `chroot` ..

jailkit-lib.pl
```diff
@@ -324,7 +324,7 @@ sub copy_jailkit_files
 	}
 
 # Remove write permissions for group
-&remove_write_permissions_for_group($dir, "home/$home_virtualmin_backup");
+&remove_write_permissions_for_group($dir, "$dir/home/$home_virtualmin_backup");
```
jcameron, line 327: What is this `home` directory though?

iliajie: A `home/` under `/home/chroot/1234567890`.

jcameron: I'm still confused - what is the full path that's being excluded here?

Seems like it should be more like `$dir/$d->{'home'}/$home_virtualmin_backup`

iliajie: I don't think so, as `$d->{'home'}` would be something referring to the normal (not _chrooted_) directory structure, like `/home/domain.com`. Perhaps, `home` should be derived from `useradmin` config `home_base` option?

iliajie: Wait, `$d->{'home'}` should already imply `home_base`, then your suggestion just perfectly right! I will fix that.

iliajie: I have fixed that and a few other bugs.

PR #770, opened 2024-03-16
Fix SSH auth

iliajie:


virtual-server-lib-funcs.pl
```diff
@@ -20056,9 +20056,16 @@ sub get_ssh_key_identifier
 sub add_domain_user_ssh_pubkey
 {
 my ($d, $user, $pubkey) = @_;
+return if (!$d->{'dir'});
+my %user = (%{$user}, ugid => $d->{'ugid'});
+my $sshdir = $user->{'home'}."/.ssh";
+&make_dir_as_domain_user(\%user, $sshdir, 0700);
```
jcameron, line 20062: Instead of creating a fake domain object to pass into these `_as_domain_user` functions, better to create a new one specifically for mailbox users.

iliajie: > better to create a new one specifically for mailbox users.

Not sure I understand? I already do this by:

```perl
my %user = (%{$user}, ugid => $d->{'ugid'});
```

iliajie: .. unless you mean to create something like sub `make_dir_as_user_domain_user()` .. I don't think it's necessary and would be an overkill for this exclusive one-case scenario.

iliajie: .. unless you mean to create something like sub `make_dir_as_mailbox_user()` .. I don't think it's necessary and would be an overkill for this exclusive one-case scenario.

jcameron: Check out this update that will allow you to just pass a Mailbox user hash ref into the various as_user functions : https://github.com/virtualmin/virtualmin-gpl/commit/ad53410b31e1a43ee5fcc7fa3dd8ebc463a95277

iliajie: Thanks! I commented on that patch.

Though, will that be helpful, as there are many more functions involved?

jcameron: Since this is called by all the `_as_domain_user` functions, it should simplify your patch..

iliajie: I will need to re-check the code tomorrow. As of now, I don't understand exactly what else needs to be done ..

jcameron: For this PR, just pass the `$user` hash ref into `make_dir_as_domain_user` and `read_file_lines_as_domain_user` , instead of creating the `%user` hash.

jcameron: I will just merge this in and apply the fix ..

iliajie: Great! I've gone through another round of testing, including the latest patches, and everything on the **Edit Users** page is functioning smoothly!

Please let me know if you find anything else.

PR #771, opened 2024-03-17
Add CGI toggle to user interface

iliajie:


edit_website.cgi
```diff
@@ -88,6 +88,16 @@ if (!$d->{'alias'} && &can_log_paths() &&
 			&ui_textbox("elog", $elog, 60));
 		}
 	}
+# CGI execution mode
+my @cgimodes = &has_cgi_support();
+if (@cgimodes > 0) {
```
jcameron, line 93: Maybe change it to `> 1` since there's no point offering this option if the user only has one choice

iliajie: What about Disable mode?

jcameron: What I mean is, if the system doesn't support fcgiwrap or suexec, then the only option will be "disabled" which it doesn't make much sense to show.

iliajie: Yeah, but `&has_cgi_support();` doesn't return disabled state ..

jcameron:
Thanks for the patch!

Issue #786, opened 2024-04-17
Creating a user via CLI without FTP or SSH access


PitWenkin:
Until recently (before [7.10.0](https://github.com/virtualmin/virtualmin-gpl/releases/tag/7.10.0)) it was possible to create a user that had no primary email, no FTP and no SSH.

`virtualmin create-user --user test --pass Ab123 --noemail --domain example.com`

This user was only able to access a mail account (`test` in this case)

Now the same command adds FTP and SSH access to the account probalby due to #707?

![Screenshot 2024-04-17 at 10 28 55](https://github.com/virtualmin/virtualmin-gpl/assets/45786226/5db2eabb-499c-4136-bd80-e06d14cb654d)
On the screenshot (`virtual-server/list_users.cgi`) above:

- `sub.demo.sub` was created before the update
- `test` was also created before the update but without the `--noemail` flag, therefore creating an account as well as a email address.
- `test1` was created after the update and was supposed to not have FTP or SSH access.

What would be the correct command to create a user account that ONLY has access to a mail account, without creating an actual email address with the name of the account?

Basically the same thing thats done when user `Add mail user` via the GUI and settings `Primary email address enabled` to `No`, which seems to do what I need:
![Screenshot 2024-04-17 at 10 49 48](https://github.com/virtualmin/virtualmin-gpl/assets/45786226/f114ff80-9a61-46f4-95e3-7ab418ea6ae9)

iliajie:
Hello,

> Now the same command adds FTP and SSH access to the account probalby due to #707?

This is somewhat expected since the Virtualmin CLI always runs as root. To address this issue, you should either always use the `--shell` parameter or change the defaults on the **System Customization ⇾ Custom Shells** page. I recommend explicitly using the `--shell` parameter.

> What would be the correct command to create a user account that ONLY has access to a mail account, without creating an actual email address with the name of the account?

You got it all right, except you're missing the `--shell /dev/null` parameter.

iliajie:
> Adding that seems to create both types of account as before.

I'm not sure I understand what you mean. It's supposed to create only one user record.

> Could you please confirm wether this is a new requirement or if this was always required, but somehow did not break until recently?

Yes, it is a new change. We no longer restrict master administrators from assigning any shell to users; this limitation only applies to normal users now, at least through the UI. However, as a side effect in the CLI, you now need to specify a particular shell explicitly.

[...]

PR #787, opened 2024-04-19
Fix to set user shell based on defaults in CLI

iliajie:
The fix for a discussion linked below:

https://github.com/virtualmin/virtualmin-gpl/issues/786#issuecomment-2065635653

### DIFF
--- a/create-user.pl
+++ b/create-user.pl
@@ -84,7 +84,7 @@ if (!$module_name) {
 @OLDARGV = @ARGV;
 
 # Get shells
-@ashells = grep { $_->{'mailbox'} && $_->{'avail'} } &list_available_shells();
+@ashells = grep { $_->{'mailbox'} && $_->{'avail'} } &list_available_shells(undef, undef, 1);
 ($nologin_shell, $ftp_shell, $jailed_shell, $shell) =
 	&get_common_available_shells();
 
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -18255,12 +18255,12 @@ if (@ttdoms > 10) {
 return join(", ", @ttdoms);
 }
 
-# list_available_shells([&domain], [mail])
+# list_available_shells([&domain], [mail], [force-defaults])
 # Returns a list of shells assignable to domain owners and/or mailboxes.
 # Each is a hash ref with shell, desc, owner and mailbox keys.
 sub list_available_shells
 {
-local ($d, $mail) = @_;
+local ($d, $mail, $fdef) = @_;
 if (!defined($mail)) {
 	$mail = !$d || $d->{'mail'};
 	}
@@ -18280,7 +18280,7 @@ if (-r $custom_shells_file) {
 	}
 if (!@rv) {
 	# Master admin should be able to create SSH users by default
-	my $defloginshell_admin = &master_admin() ? 1 : 0;
+	my $defloginshell_admin = (&master_admin() && !$fdef) ? 1 : 0;
 
 	# Fake up from config file and known shells, if there is no custom
 	# file or if it is somehow empty.


jcameron:
Isn't the problem that the wrong default shell is returned from `get_common_available_shells` ?

iliajie:
> Isn't the problem that the wrong default shell is returned from `get_common_available_shells` ?

No, and we didn't even make any change to `get_common_available_shells` in the related PR in the first place..

jcameron:
I don't understand how this PR fixes the problem in `create-user.pl` when that script never even uses the return value from `list_available_shells` : https://github.com/virtualmin/virtualmin-gpl/blob/master/create-user.pl#L87

jcameron:
Hang on, I think there's a deeper problem here - the `default` field returned by `list_available_shells` isn't supposed to change based on who calls the function! It is supposed to indicate which shell is the default for a certain type, but it's not consistent due to this code : 

https://github.com/virtualmin/virtualmin-gpl/blob/master/virtual-server-lib-funcs.pl#L18291

I think you changed this as part of adding support for SSH users, but why?

jcameron:
I just checked in a couple of fixes for this issue, but I'm not 100% sure of your original intentions with the `defloginshell_admin` variable. Why does this change based on the current user?

iliajie:
> I just checked in a couple of fixes for this issue, but I'm not 100% sure of your original intentions with the `defloginshell_admin` variable. Why does this change based on the current user?

Jamie, we cannot roll back functionality that we agreed to add before and that was already presented to the users. Those fixes are breaking the initial functionality that was added in the users' overhaul PR. The point of it was to always allow master administrators to create any type of users for the given virtual server. It didn't change anything in terms of security for non-privileged users. Custom Shells are not configured in a way that allows the master administrator (effectively the _root_ user) to create SSH users with the `/bin/bash` shell. That means it again won't be possible to create SSH users for virtual servers by default, expecting that admins will be able to adjust everything correctly in Custom Shells—I think we shouldn't add extra complexity and confusion because the master administrator, per se, cannot have any limitations. So if something is missing for master administrator login, it will create a false perception that the feature is missing completely.

I'd like to point out that my current PR addresses this issue precisely by creating correctly different types of cache.

The confusion from CLI perspective came from the fact that in the UI we automatically set the needed shell based on the user type being created. If a mail user is created, then the shell is automatically set to `/dev/null`; if an SSH user is created, then the shell is set to `/bin/bash`, and so on. If a default user is being created (i.e., using the Add User button), then the shell was also set to Email and FTP, even though the master administrator was doing it. Yet, it was possible to switch it to Email, FTP, and SSH. With the latest checked-in fixes, it breaks this! I spent a lot of time making sure everything worked flawlessly in the first place!

On the contrary, in CLI, as an admin doesn't have a UI to create a specific type of user or a dropdown to select the shell, it was incorrectly defaulting to `/bin/bash`. This PR addressed the original problem with perfect precision.

I'd like to suggest we revert your recent fixes and work it out from there using this PR, please.

jcameron:
Oh for sure I don't want to break existing functionality! But the code has to be correct and consistent, and do things in a way that makes sense. Just because something works doesn't necessarily mean it is right.

In this specific case, the `default` field in the list of shells returned by `list_available_shells` is supposed to indicate that it is the default shell for a certain type of user. This is something that shouldn't be dependent on what user is calling the function ... if we want to differentiate between shells that the `root` user can assign and what resellers or domain owners can assign, this should be done via a new field or some other selector.

jcameron:
Ok I've been looking through the code that generates the list of shells, which I admit is very confusing!

Who do we want to be able to create SSH users - domain owers, or just the `root` user?

iliajie:
> Who do we want to be able to create SSH users - domain owers, or just the `root` user?

We want `root` user to be able to create any type of user but domain owner should strictly adhere to whatever set in Custom Shells.

jcameron:
Ok, makes sense. I'm re-doing the code now to make this work correctly, because the old code didn't support this use case cleanly..

iliajie:
> Ok, makes sense. I'm re-doing the code now to make this work correctly, because the old code didn't support this use case cleanly..

Alright, tell me please when it's ready..

jcameron:
Ok check out the latest commit. If there's anything that doesn't work, please let me know what functionality is broken..

jcameron:
Oh that should work. Is this on a system with custom shells defined? If not, what output do you get from `grep shell /etc/webmin/virtual-server/config` ?

iliajie:
> Oh that should work. Is this on a system with custom shells defined? If not, what output do you get from `grep shell /etc/webmin/virtual-server/config` ?

I'm pretty sure it doesn't. The custom shells are intact (set as default).

The output is:

```
root@ubuntu22-pro:/etc# grep shell /etc/webmin/virtual-server/config
ftp_shell=/bin/false
unix_shell=/bin/bash /bin/sh
avail_shell=0
shell=/dev/null
defushell=none
```

jcameron:
Ok those look like the standard settings. What are you seeing in the UI?

iliajie:
> Ok those look like the standard settings. What are you seeing in the UI?

In the UI (logged in as master admin) is see a dropdown for **Login permissions** with only **Email and FTP** and **Email only** options whereas **Email, FTP and SSH** is missing.

jcameron:
Hmm, I'm not seeing this..

What output do you get if you run `virtualmin list-available-shells --multiline` with all my latest patches applied?

iliajie:
> What output do you get if you run `virtualmin list-available-shells --multiline` with all my latest patches applied?

```
/bin/false
    Description: Email and FTP
    Login type: ftp
    For mailboxes: Yes
    For administrators: No
    For resellers: No
    Available: Yes
    Default: No
/bin/false
    Description: Email and FTP
    Login type: ftp
    For mailboxes: No
    For administrators: Yes
    For resellers: Yes
    Available: Yes
    Default: No
/dev/null
    Description: Email only
    Login type: nologin
    For mailboxes: Yes
    For administrators: No
    For resellers: No
    Available: Yes
    Default: Yes
/dev/null
    Description: Email only
    Login type: nologin
    For mailboxes: No
    For administrators: Yes
    For resellers: Yes
    Available: Yes
    Default: No
/sbin/nologin
    Description: Email only
    Login type: nologin
    For mailboxes: No
    For administrators: Yes
    For resellers: Yes
    Available: No
    Default: No
/bin/bash
    Description: Email, FTP and SSH
    Login type: ssh
    For mailboxes: No
    For administrators: Yes
    For resellers: Yes
    Available: Yes
    Default: Yes
/bin/dash
    Description: Email, FTP and SSH
    Login type: ssh
    For mailboxes: No
    For administrators: Yes
    For resellers: Yes
    Available: No
    Default: No
/bin/rbash
    Description: Email, FTP and SSH
    Login type: ssh
    For mailboxes: No
    For administrators: Yes
    For resellers: Yes
    Available: No
    Default: No
/bin/sh
    Description: Email, FTP and SSH
    Login type: ssh
    For mailboxes: No
    For administrators: Yes
    For resellers: Yes
    Available: No
    Default: No
/usr/bin/bash
    Description: Email, FTP and SSH
    Login type: ssh
    For mailboxes: No
    For administrators: Yes
    For resellers: Yes
    Available: No
    Default: No
/usr/bin/dash
    Description: Email, FTP and SSH
    Login type: ssh
    For mailboxes: No
    For administrators: Yes
    For resellers: Yes
    Available: No
    Default: No
/usr/bin/rbash
    Description: Email, FTP and SSH
    Login type: ssh
    For mailboxes: No
    For administrators: Yes
    For resellers: Yes
    Available: No
    Default: No
/usr/bin/screen
    Description: Email, FTP and SSH
    Login type: ssh
    For mailboxes: No
    For administrators: Yes
    For resellers: Yes
    Available: No
    Default: No
/usr/bin/sh
    Description: Email, FTP and SSH
    Login type: ssh
    For mailboxes: No
    For administrators: Yes
    For resellers: Yes
    Available: No
    Default: No
/usr/bin/tmux
    Description: Email, FTP and SSH
    Login type: ssh
    For mailboxes: No
    For administrators: Yes
    For resellers: Yes
    Available: No
    Default: No
```

iliajie:
> Hmm, I'm not seeing this

Nevermind, I have checked all recent changes and made a new PR.

jcameron:
Ok so I see an SSH-capable shell is available for domain owners :

```
v/bin/bash
    Description: Email, FTP and SSH
    Login type: ssh
    For mailboxes: No
    For administrators: Yes
    For resellers: Yes
    Available: Yes
    Default: Yes
```

And when creating an SSH user it should be selectable here : https://github.com/virtualmin/virtualmin-gpl/blob/master/edit_user.cgi#L88

iliajie:
I already a [PR](https://github.com/virtualmin/virtualmin-gpl/pull/789) that deals with `available_shells_menu`.

jcameron:
I saw the PR, but I don't get how it solves the problem cleanly - it's just allowing the root user to select *any* shell, when what we want is to allow use of domain owner shells.

iliajie:
> it's just allowing the root user to select _any_ shell, when what we want is to allow use of domain owner shells.

Ah, alright. I now understand your initial thought. We could do that, as I've [just patched](https://github.com/virtualmin/virtualmin-gpl/commit/c4f3f8f0781a2e2c5538a8d94eda1b2f6138c485) it.

However, I don't see it as the right solution. The reason for this was that we cannot guarantee that the **Custom Shells** page won't be changed, where the domain admin has the shell set to `/dev/null` and the mailbox user to `/bin/bash`. I initially wanted it to be flexible, where the _root_ can choose any shell unconditionally. Does that make sense?

jcameron:
Actually I don't think we do want the root user to select *any* shell - at least that's not what Virtualmin has allowed historically. Only the kinds of shells that are relevant for the user..

iliajie:
> Actually I don't think we do want the root user to select _any_ shell

Well, we already discussed it on the initial PR, and released Virtualmin version that allows to do it. 

> at least that's not what Virtualmin has allowed historically.

Yes, that's new.

> Only the kinds of shells that are relevant for the user..

The _root_ user must be able to set any shell for a user, as previous limitation doesn't really go along well with philosophy of _root_-capable user. A domain owner is forced to follow **Custom Shells** rules.

jcameron:
I mean the root user can do this in the Webmin Users and Groups module. I think that in the Virtualmin UI they should be limited to shells that make sense..

iliajie:
Yeah, that's right. Although we can't expect our users to know which shell is right nor where to change it. A much simpler and user-friendly approach is to keep it simple and just allow selection from a dropdown where login permissions clearly state **Email only, FTP, and SSH**. We also provide an SSH public key box for users to enter their public key.

I wouldn't want to make it more complicated for users. And there have been users explicitly asking for this feature. SSH access is a very basic and important out-of-the-box functionality.

I suggest either allowing this to work as it is now, i.e., to always allow a _root_-capable user to set any shell, or better, to always allow the server owner to set any shell for its users. But despite how we do it, the SSH option should always be easily available.

jcameron:
Agreed, we always want the root user to be able to select an SSH-capable shell. I'm not sure if we want domain owners to be able to set SSH shells for their mailboxes though, because what if the domain owner doesn't have SSH access himself?

iliajie:
> Agreed, we always want the root user to be able to select an SSH-capable shell.

Great!



> I'm not sure if we want domain owners to be able to set SSH shells for their mailboxes though, because what if the domain owner doesn't have SSH access himself?

So, if a domain owner was granted such privilege by a root user then it should be okay for the domain owner to set it for other domain users?

jcameron:
I suppose that if the domain owner has an SSH-capable shell, they should be able to set a similar shell on mailboxes?

iliajie:
Yeah, I think so too.

iliajie:
I'm not sure. There is already a select for a shell. If the domain owner is allowed of SSH, then all users created by this user should also be allowed SSH. No need for a new option, I don't think.

jcameron:
Ok I'll look into this...

PR #789, opened 2024-04-24
Fix to allow the root user to assign any shells

iliajie:
In continuation of https://github.com/virtualmin/virtualmin-gpl/pull/787#issuecomment-2073636834

### DIFF
--- a/edit_user.cgi
+++ b/edit_user.cgi
@@ -811,8 +811,9 @@ else {
 				}
 			}
 		print &ui_table_row(&hlink($text{'user_ushell'}, "ushell"),
-			&available_shells_menu("shell", $user_shell, "mailbox",
-					0, $user->{'webowner'}),
+			&available_shells_menu("shell", $user_shell,
+				&can_mailbox_ssh() ? "owner" : "mailbox",
+				0, $user->{'webowner'}),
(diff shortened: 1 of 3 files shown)


jcameron:
Hang on, I don't see why this is needed when the owner type is already passed into available_shells_menu for SSH users here : https://github.com/virtualmin/virtualmin-gpl/blob/master/edit_user.cgi#L97

iliajie:
> Hang on, I don't see why this is needed when the owner type is already passed into available_shells_menu for SSH users here : https://github.com/virtualmin/virtualmin-gpl/blob/master/edit_user.cgi#L97

Because there is another one that is set to `mailbox` user here :

https://github.com/virtualmin/virtualmin-gpl/blob/master/edit_user.cgi#L814

iliajie:
Anyway, here is the expected fix https://github.com/virtualmin/virtualmin-gpl/pull/789/commits/7e2579fa30370042e28c187f36ad6f4553efd89f.

jcameron:
Ok that looks better ..thanks!

iliajie:
> Ok that looks better ..thanks!

Alright! By the way, I didn't add it to `edit_defaults.cgi` page.

PR #807, opened 2024-05-10
Add ability to reinstall webapps dependencies

iliajie:
Hey Jamie!

This PR introduces the ability to reinstall web application dependencies. This is especially useful, for example, when a new version of PHP is installed and a virtual server switches to it, but the newly installed PHP version lacks the necessary extensions for a webapp.

Screenshot 2024-05-10 at 21 00 02
Screenshot 2024-05-10 at 21 17 24


script_reinstall_dependencies.cgi
```diff
@@ -0,0 +1,54 @@
...
+require './virtual-server-lib.pl';
+&ReadParse();
+my $d = &get_domain($in{'dom'});
+&can_edit_domain($d) && &can_edit_scripts() || &error($text{'edit_ecannot'});
+
+# Get the script being started
+my @got = &list_domain_scripts($d);
+my ($sinfo) = grep { $_->{'id'} eq $in{'script'} } @got;
+my $script = &get_script($sinfo->{'name'});
+$sinfo && $script || &error($text{'stopscript_egone'});
```
jcameron, line 13: Move the check for $sinfo being set to line 11, because dereferencing it on line 12 will cause it to become defined even if undef.

script_reinstall_dependencies.cgi
```diff
@@ -0,0 +1,54 @@
...
+
+# Work out PHP version for this domain
+my $phpver;
+my @dirs = &list_domain_php_directories($d);
+foreach my $dir (@dirs) {
```
jcameron, line 18: Can't you do this in one loop with `is_under_directory` ?

script_reinstall_dependencies.cgi
iliajie: What about this https://github.com/virtualmin/virtualmin-gpl/commit/4659c94d715c88e0bdbf402a763374eaffcc45a8 one?

script_reinstall_dependencies.cgi
iliajie: Good question, I've just reused your code in this block. Though, I think it can be reduced to a single loop - https://github.com/virtualmin/virtualmin-gpl/pull/807/commits/c5b06d406180a65f569ad5cf3d797dfbe46c9fbb.

script_reinstall_dependencies.cgi
jcameron: Looks good!

script_reinstall_dependencies.cgi
jcameron: Where did you copy this code from?

iliajie: Backup library, if I remember now correctly.

jcameron: It could be replaced with just : 

```
        if (&is_under_directory($dir->{'dir'}, $sinfo->{'dir'})) {
                $phpver ||= $dir->{'version'};
                }
```

iliajie: What if there are multiple PHP versions set inside `DocumentRoot`?

jcameron: Oh I just realized that for this to work, the directories need to be processed in order from longest to shortest so that the most specific one matches.

iliajie: So the current code with combined loops isn't okay? Do we need two separate loops as in my initial patch?

iliajie: I took a closer look, and you were right. PHP wasn't detected correctly for a directory. There was also a bug that didn't pass the script's directory correctly. Both issues have been fixed in this https://github.com/virtualmin/virtualmin-gpl/pull/807/commits/e5c0270ba64ef2c8ea61b3d16a93eed41d444c14 commit.

jcameron: Why not use is_under_directory like this code : https://github.com/virtualmin/virtualmin-gpl/blob/master/scripts-lib.pl#L1192

iliajie: It seems much broader than using a precise equality test.

Besides, the linked code doesn't correctly sort directories from the longest to the shortest.

jcameron: It sorts from the shortest to longest, but the last (longest) match is the one kept.

We can't always use a precise equality test, becuase the user might set path / to PHP 7, path /foo to PHP 8, and have a script that uses PHP 8 in /foo/bar .

iliajie: Ah, thanks! I get it now! Check this patch https://github.com/virtualmin/virtualmin-gpl/commit/9945738338e481fc33b67efac9f8fa09bba30327 out.

php-lib.pl
```diff
@@ -1927,6 +1927,34 @@ sub get_domain_php_version
...
+my ($d, $dir) = @_;
+return undef if (!$d);
+my $phpver;
+my @pdirs = &list_domain_php_directories($d);
+@pdirs = sort { length($a->{'dir'}) <=> length($b->{'dir'}) } @pdirs;
+foreach my $pdir (@pdirs) {
+	if ($pdir->{'dir'} eq $dir) {
+		$phpver = $pdir->{'version'};
+		last;
+		}
```
jcameron, line 1946: Why is this block needed when the code below will also match if `$pdir->{'dir'}` is the same as `$dir` ?

php-lib.pl
```diff
@@ -1927,6 +1927,34 @@ sub get_domain_php_version
...
 }
 
+# get_domain_php_version_for_directory(&domain, [directory])
+# Get the PHP version used by the domain in some directory.
+# If directory param is not specified, it defaults to the
+# public_html directory.
+# If PHP is disabled or not set, undef is returned.
+sub get_domain_php_version_for_directory
+{
+my ($d, $dir) = @_;
```
jcameron, line 1937: Add the line : 

`$dir ||= &public_html_dir($d);`

php-lib.pl
```diff
@@ -1927,6 +1927,34 @@ sub get_domain_php_version
...
+return undef if (!$d);
+my $phpver;
+my @pdirs = &list_domain_php_directories($d);
+@pdirs = sort { length($a->{'dir'}) <=> length($b->{'dir'}) } @pdirs;
+foreach my $pdir (@pdirs) {
+	if ($pdir->{'dir'} eq $dir) {
+		$phpver = $pdir->{'version'};
+		last;
+		}
+	if (&is_under_directory($pdir->{'dir'}, $dir || &public_html_dir($d))) {
```
jcameron, line 1947: Remove `|| &public_html_dir($d)` here

iliajie: Alright. Fixed.

php-lib.pl
iliajie: This one fixed too.

jcameron:
Thanks for the patch!

PR #810, opened 2024-05-15
Add support for forbidding features when creating or modifying some domains

iliajie:
Hey Jamie!

As discussed in [this thread](https://forum.virtualmin.com/t/new-installer-questions-about-the-default-ssl-certificate/126684/10?u=ilia) this PR protects users from enabling mail for a virtual server that matches hostname.

### DIFF
--- a/domain_setup.cgi
+++ b/domain_setup.cgi
@@ -457,6 +457,11 @@ $cerr = &virtual_server_clashes(\%dom);
 $lerr = &virtual_server_limits(\%dom);
 &error($lerr) if ($lerr);
 
+# Check if features are not forbidden
+foreach my $ff (&forbidden_domain_features(\%dom, 1)) {
+	$dom{$ff} && &error(&text('setup_efeatforbidhostdef', $ff));
+	}
+
 # Update custom fields
 &parse_custom_fields(\%dom, \%in);
 
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -19714,6 +19714,7 @@ sub forbidden_domain_features
 {
 my ($d) = @_;
 my @rv;
+return @rv if ($config{'nocheck_forbidden_domain_features'});
 # Not allowed features for host default domain
 if ($d->{'dom'} eq $config{'defaultdomain_name'}) {
 	push(@rv, 'spam', 'virus', 'mail');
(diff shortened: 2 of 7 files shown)


jcameron:
Is having email enabled when the domain name matches the hostname really not allowed? The linked thread doesn't really go into this at all ..

iliajie:
> Is having email enabled when the domain name matches the hostname really not allowed? The linked thread doesn't really go into this at all ..

This is something [Joe](@swelljoe) has always advised against! It didn't just come up in this thread this time.

verne-work:
will there be a way to override it somehow, someplace?   Not for mail, but we create a virtual server to have apache respond to the server's name so our external Nagios monitoring system can test if apache is running or not.

iliajie:
It won't display mail-related options only, and only if they aren't enabled. 

It will be possible to use Virtualmin CLI to force enable those features.

jcameron:
@verne-work it will still be possible to have a domain with a website enabled that matches the hostname

virtual-server-lib-funcs.pl
```diff
@@ -19708,6 +19708,35 @@ sub list_possible_domain_features
...
+return @rv;
+}
+
+# filter_possible_domain_features(&features, &domain, [creating])
+# Given a list of features, returns only those that aren't forbidden for domain
+sub filter_possible_domain_features
+{
+my ($features, $d, $new) = @_;
+my @forbidden = &forbidden_domain_features($d, $new);
+@$features = grep {
```
jcameron, line 19735: Why not return the new list instead of updating the array reference?

domain_setup.cgi
```diff
@@ -457,6 +457,11 @@ $cerr = &virtual_server_clashes(\%dom);
 $lerr = &virtual_server_limits(\%dom);
 &error($lerr) if ($lerr);
 
+# Check if features are not forbidden
+foreach my $ff (&forbidden_domain_features(\%dom, 1)) {
+	$dom{$ff} && &error(&text('setup_efeatforbidhostdef', $ff));
```
jcameron, line 462: This message is very specific to the error of the domain name matching the hostname, but couldn't forbidden features apply in other situations?

virtual-server-lib-funcs.pl
iliajie: It's a better practice in Perl. We do it all the time.

domain_setup.cgi
iliajie: What other situations? Oh, you mean those that don't potentially exist yet?

iliajie: Currently it's only meant to be used with default host domain.. if there is something else, we'll adjust an error message.

Issue #809, opened 2024-05-14
Prevent selecting spam and virus filtering when there is no mail service


shoulders:
|  SYSTEM INFORMATION||
|----------------------|---------------------------|
|  OS type and version |     Ubuntu Linux 22.04.4     |
|  Webmin version      |     2.111      |
|  Usermin version     |     2.010     |
|  Virtualmin version  |     7.10.0       |
|  Theme version       |     21.10      |
|  Package updates     |      44  package updates are available   |

### the issue

When creating a new virtual server, you can select spam and virus filtering for a mail even when this service is not enabled.

![image](https://github.com/virtualmin/virtualmin-gpl/assets/319997/2aa9ca9b-4ac4-4f7a-b004-9a5d873fc7e8)

### proposed solution

can't select spam and mail filtering if no mail service is selected

### Additional

Some websites will send mail via PHP and sendmail and I do not know how this would be affected.

iliajie:
Hello,

> ### proposed solution
> can't select spam and mail filtering if no mail service is selected

That would involve JavaScript. Though, if you try to submit it as it is, you'd still see an error.

However, automatically removing ticks from Spam and Virus options if mail is disabled would be cool.

jcameron:
@iliajie you could send me a PR for this if you like?

iliajie:
Yes, will do!

iliajie:
Implemented here https://github.com/virtualmin/virtualmin-gpl/pull/811.

PR #811, opened 2024-05-15
Add chained features API

iliajie:
Hello Jamie!

As discussed [here](https://github.com/virtualmin/virtualmin-gpl/issues/809) but will not only disable dependent features but also re-enable.

### DIFF
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -5842,16 +5842,20 @@ sub feature_check_chained_javascript
 {
 my ($f) = @_;
 my $chained = {
-	mail => ['spam', 'virus'],
+	'mail'                 => ['spam', 'virus'],
+	'web'                  => ['ssl', 'status', 'webalizer', 'virtualmin-awstats'],
+	'virtualmin-nginx'     => ['virtualmin-nginx-ssl', 'status', 'webalizer', 'virtualmin-awstats'],
 };
-if (exists($chained->{$f})) {
-	my $deps = join(', ', map { "form.$_ && (form.$_.checked = false)" }
-			@{$chained->{$f}});
-	return "oninput=\"if (!form.$f.checked) { $deps }\"";
+my $cfeature = $chained->{$f};
+if ($cfeature) {
+	my $deps = join(', ', map { "form['$_'] && (form['$_'].checked = false)" }
+			@{$cfeature});
+	return "oninput=\"if (form['$f'] && !form['$f'].checked) { $deps }\"";
 	}
 for my $c (keys %$chained) {
-	return "oninput=\"if (form.$f.checked) ".
-			   "{ form.$c && (form.$c.checked = true) }\""
+	next if (!$config{$c} && &indexof($c, @plugins) < 0);
+	return "oninput=\"if (form['$f'] && form['$f'].checked) ".
+			   "{ form['$c'] && (form['$c'].checked = true) }\""
 		if (grep { $_ eq $f } @{$chained->{$c}});
 	}
 return undef;
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -5836,6 +5836,27 @@ else {
 	}
 }
 
+# feature_check_chained_javascript(feature)
+# Return inline JavaScript code to chain disable/enable dependent features
+sub feature_check_chained_javascript
+{
+my ($f) = @_;
+my $chained = {
+	mail => ['spam', 'virus'],
+};
+if (exists($chained->{$f})) {
+	my $deps = join(', ', map { "form.$_ && (form.$_.checked = false)" }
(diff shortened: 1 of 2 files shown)


virtual-server-lib-funcs.pl
```diff
@@ -5836,6 +5836,27 @@ sub quota_parse
 	}
 }
 
+# feature_check_chained_javascript(feature)
+# Return inline JavaScript code to chain disable/enable dependent features
+sub feature_check_chained_javascript
+{
+my ($f) = @_;
+my $chained = {
+	mail => ['spam', 'virus'],
```
jcameron, line 5845: I wonder if there's other dependencies we could add while we're in there, like webalizer on web ?

virtual-server-lib-funcs.pl
```diff
@@ -5836,6 +5836,27 @@ sub quota_parse
...
+};
+if (exists($chained->{$f})) {
+	my $deps = join(', ', map { "form.$_ && (form.$_.checked = false)" }
+			@{$chained->{$f}});
+	return "oninput=\"if (!form.$f.checked) { $deps }\"";
+	}
+for my $c (keys %$chained) {
+	return "oninput=\"if (form.$f.checked) ".
```
jcameron, line 5853: Is there ever danger that some feature field doesn't exist?

virtual-server-lib-funcs.pl
iliajie: Sure, then we just add `web => ['webalizer', 'virtualmin-awstats']` ..

iliajie:
I have run more tests and fixed a few bugs. I also added more chained features as you suggested earlier! All feels smooth now. Please have a look https://github.com/virtualmin/virtualmin-gpl/pull/811/commits/2d6ba33d7ec78407a27c9206eab13258b2f887e0.

virtual-server-lib-funcs.pl
jcameron: Although I just realized that they can work with Nginx, so it's not that simple :-(

virtual-server-lib-funcs.pl
```diff
@@ -5836,6 +5836,31 @@ sub quota_parse
...
+};
+my $cfeature = $chained->{$f};
+if ($cfeature) {
+	my $deps = join(', ', map { "form['$_'] && (form['$_'].checked = false)" }
+			@{$cfeature});
+	return "oninput=\"if (form['$f'] && !form['$f'].checked) { $deps }\"";
+	}
+for my $c (keys %$chained) {
+	next if (!$config{$c} && &indexof($c, @plugins) < 0);
+	return "oninput=\"if (form['$f'] && form['$f'].checked) ".
```
jcameron, line 5857: Is this changing the checked state of other fields? Shouldn't it be disabling fields that cannot be used?

virtual-server-lib-funcs.pl
iliajie: I already fixed that. All good.

virtual-server-lib-funcs.pl
iliajie: Yes, but it also enables dependent features.

For example, it will disable Spam and Virus if Mail feature is unticked; it will auto-enable Mail if either Spam or Virus is ticked.

jcameron: I think you should just disable dependent features when they cannot be used - going the other way seems quite confusing.

iliajie: I don't think it's confusing at all. On contrary, it doesn't let user choose options that will fail anyway after saving the form.

That said, how it can be confusing selecting AWStats reporting feature and not have Apache/Nginx option enabled?

jcameron: So if the mail feature is disabled, will the checkboxes for spam and virus be disabled? Or just un-checked?

iliajie: > So if the mail feature is disabled, will the checkboxes for spam and virus be disabled? Or just un-checked?

They will be unchecked, i.e. disabled! 😅

jcameron: "disabled" to me means greyed out, not just un-checked

iliajie: Ah, I see. But do we ever gray-out features on **Edit Virtual Server** page? I don't think we do.

This API is made to work only with enabled (not disables) checkboxes. Disabled state is controlled by the server.

jcameron: Ok I see now .. I was confused about what disabling the field meant.

iliajie:
Jamie, nearly forgot to add this when creating a new domain! Check this patch please:

https://github.com/virtualmin/virtualmin-gpl/commit/f4433153dbbb017932b9a65f62331575c7b073c4

jcameron:
Looks good!

PR #814, opened 2024-05-18
Scripts are Webapps

iliajie:
Hello Jamie!

As we discussed earlier, I have changed the "Scripts" label to "Webapps" to better reflect the functionality. I have updated this in the language and help files, as well as all translations.


jcameron:
This is a big change in wording! @swelljoe what do you think?

swelljoe:
I don't like webapps (all in one word) either.

Web Apps or Applications would be better. Applications is probably my preference.

swelljoe:
Web Apps is fine, though.

iliajie:
Thanks for pointing this out, Joe. Fixed on the latest commit.

jcameron:
Ok "Web Apps" it is then! Although I don't think we should rename any API commands or their output..

iliajie:
Perhaps we could create a link?

jcameron:
I'd rather not ... best to keep APIs stable, because we have no idea what scripts that use them exist out there.

iliajie:
Can't we keep the API stable and backward-compatible with all old `-script` versions while using the new `-webapp` instead?

iliajie:
Can't we keep the API stable and backward-compatible with all old API `-script` filenames while also using the new `-webapp`?

jcameron:
I'd prefer not to, because even if we create symlinks the API commands include their own names in their output. Best not to make too many unnecessary changes.

iliajie:
Alright, then we're good to move on.

iliajie:
Although, shouldn't we at least mention in parentheses (web app) in help? I mean like this:

```
root@ubuntu24-pro:~# virtualmin | grep script
install-script            Install one third-party script (web app)
delete-script             Un-install one script (web app) from a virtual server
list-scripts              Display script (web app) installed into some virtual server
list-available-scripts    List known scripts (web apps)
```

jcameron:
Sure that seems reasonable!

iliajie:
> Sure that seems reasonable!

Okay, done https://github.com/virtualmin/virtualmin-gpl/pull/814/commits/613b4ba4949d9284a8faea3705ed573635f27424.

PR #820, opened 2024-06-03
Fix to set default domain serve as default service certificate

iliajie:
This patch addresses what was [discussed here](https://forum.virtualmin.com/t/how-to-manage-default-ssl-cert-for-hostname/126970/15?u=ilia).

### DIFF
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -20379,8 +20379,8 @@ if (&can_default_website(\%dom)) {
 &push_all_print();
 &set_all_null_print();
 foreach my $st (&list_service_ssl_cert_types()) {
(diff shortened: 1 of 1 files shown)


virtual-server-lib-funcs.pl
```diff
@@ -20375,6 +20375,18 @@ sub setup_virtualmin_default_hostname_ssl
 if (&can_default_website(\%dom)) {
 	&set_default_website(\%dom);
 	}
+# Enable as default domain for all services
+&push_all_print();
+&set_all_null_print();
+foreach my $st (&list_service_ssl_cert_types()) {
+	($a) = grep { !$_->{'d'} && $_->{'id'} eq $st->{'id'} }
```
jcameron, line 20382: Add `my` before `($a)`

iliajie: Thanks! Fixed now!

PR #821, opened 2024-06-04
Fix to conditionally sent XFP header and fix `mod_headers` check

iliajie:
Hello Jamie!

This PR fixes ongoing issue as discussed [here](https://github.com/virtualmin/virtualmin-gpl/commit/68b828f18ffc6485a04c53e19ae849f449004911#r142691321).

### DIFF
--- a/balancer-lib.pl
+++ b/balancer-lib.pl
@@ -104,9 +104,11 @@ foreach my $port (@ports) {
 			    'type' => 1,
 			    'value' => "balancer://$balancer->{'balancer'}",
 			    'members' => \@mems };
+		my $balancer_ssl;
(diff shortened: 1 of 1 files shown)


jcameron:
This isn't what we want though, because that header indicates if the *original* protocol was SSL, not the protocol from Apache to the backend.

iliajie:
> This isn't what we want though, because that header indicates if the _original_ protocol was SSL, not the protocol from Apache to the backend.

Ah, you're right! What about this change then:

https://github.com/virtualmin/virtualmin-gpl/pull/821/files

jcameron:
Thanks, looks good!

iliajie:
Welcome!

Issue #305, opened 2021-08-26
website proxy does not support websockets


iateadonut:
On a server, go to Server Configuration -> Edit Proxy Website

Set "Proxy enabled?" to "Yes". Proxy to url: http://localhost:12809/

Websockets are not enabled.

To fix this, in the apache directives, under RewriteEngine on, the following directives need to be put in place by virtualmin (currently you have to do this by hand) - assuming the previous value is "http://localhost:12809/":

  RewriteCond %{HTTP:UPGRADE} ^WebSocket$ [NC]
  RewriteCond %{HTTP:CONNECTION} ^Upgrade$ [NC]
  RewriteRule .* ws://localhost:12809%{REQUEST_URI} [P]

shoulders:
i think this can be closed 😄 

HTTPS is the only type of proxying recommended now.

iateadonut:
This is a discussion about websockets.  Would the same issue would still exist even if you're using a secure websocket connection wss rather than ws?

If you point me to the direction in the source code for proxy connections, I'll check it out.

jcameron:
The relevant code is here : https://github.com/virtualmin/virtualmin-gpl/blob/master/balancer-lib.pl#L57

iliajie:
Jamie, I think we should also setup WebSocket proxy! We have a tested and perfectly working example described in [Webmin FAQ](https://webmin.com/faq/#can-i-run-webmin-or-usermin-behind-reverse-proxy). Please note that for `:80`  we'd want to setup `ws://` and for `:443` it should be `wss://`.

jcameron:
Good idea, I'll work on this..

jcameron:
So would these directives actually replace the ProxyPass that we're currently using?

```
RewriteCond %{HTTP:UPGRADE} ^WebSocket$ [NC]
RewriteCond %{HTTP:CONNECTION} ^Upgrade$ [NC]
RewriteRule .* ws://localhost:12809%{REQUEST_URI} [P]
```

iliajie:
> So would these directives actually replace the ProxyPass that we're currently using?

It won't replace the directives we're currently using. These will be extra.

iliajie:
Have this been implemented yet in the light of the latest commits to the Virtualmin repo?

jcameron:
Not yet, but it's getting there!

jcameron:
Is there a good way I can test websockets proxying from the shell, to validate that this feature works?

[...]

PR #822, opened 2024-06-06
Fix new WebSockets proxy API after extensive testings

iliajie:
Hey Jamie!

As it was [discussed here](https://github.com/virtualmin/virtualmin-gpl/issues/305#issuecomment-2151405458), I decided to run tests using this new feature and found and fixed a few crucial bugs!

1. We must avoid mixed content by using correct protocol for the WebSocket;
2. Using `%{REQUEST_URI}` inserts the full original request URI, including `$b->{'path'}` and everything after it, so it just won't work, and therefore we should use `RewriteRule` with substitution string;
3. If a balancer path originally ends with `/`, like `/webmin/` then it would break everything just by a simple form re-save (without making any changes)

The test case was a proxy to Webmin.

Apache config: 
  1. Path was set to `/webmin/`;
  2. Backend URL was set to `https://localhost:10000/`.

Webmin:
 1. `/etc/webmin/config` file:
    ```
    referers=127.0.0.1 ubuntu24-pro.virtualmin.dev
    webprefix=/webmin
    webprefixnoredir=1
    ```
 2. `/etc/webmin/miniserv.conf` file:
    ```
    cookiepath=/webmin
    ```

### DIFF
--- a/balancer-lib.pl
+++ b/balancer-lib.pl
@@ -41,7 +41,7 @@ foreach my $pp (&apache::find_directive("ProxyPass", $vconf)) {
 		       'none' => 1 };
 		}
 	if ($b) {
-		my ($rwr) = grep { /^\Q$b->{'path'}\E\s+ws:\/\// } @rwr;
+		my ($rwr) = grep { /^\Q^$b->{'path'}?(.*)\E\s+"ws?s:\/\// } @rwr;
 		$b->{'websockets'} = 1 if ($rwr);
 		push(@rv, $b);
 		}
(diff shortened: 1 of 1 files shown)


iliajie:
Besides, as Joe mentioned earlier, maybe we don't really need to set up a balancer at all? I don't even think a WebSockets proxy can work with a balancer at all?

jcameron:
Thanks for the fixes!

And yes we don't need a balancer unless proxying to multiple URLs.

iliajie:
> Thanks for the fixes!
> 
> And yes we don't need a balancer unless proxying to multiple URLs.

Will you please fix that?

jcameron:
> Will you please fix that?

Fix that though? We don't create a balancer block currently unless multiple URLs are given

iliajie:
> We don't create a balancer block currently unless multiple URLs are given

We actually do!

jcameron:
What command or UI action triggers this?

iliajie:
This creates a balancer:

image

iliajie:
This fixes it https://github.com/virtualmin/virtualmin-gpl/pull/824.

iateadonut:
Would this also work with the backend set to (insecure) http://localhost:PORT?

(Because the websockets server is accessed through the apache proxy, the security of the WebSocket connection will be handled by the HTTPS proxy, as long as the PORT is not exposed to the Internet.)

iliajie:
> Would this also work with the backend set to (insecure) http://localhost:PORT?

Yes, it will.

PR #823, opened 2024-06-06
Fix S3 bugs

iliajie:
Hey Jamie!

This PR fixes a few S3 related issues

  1. Fix to sent expected params
  2. Fix not to require S3 modules if AWS CLI is available

### DIFF
--- a/save_s3.cgi
+++ b/save_s3.cgi
@@ -54,7 +54,7 @@ else {
 		}
 
 	# Validate that it works
-	my $buckets = &s3_list_buckets($s3);
+	my $buckets = &s3_list_buckets($s3->{'access'}, $s3->{'secret'});
 	if (!ref($buckets)) {
 		&delete_s3_account($s3) if ($in{'new'});
 		&error(&text('s3_echeck', $buckets));
--- a/s3-lib.pl
+++ b/s3-lib.pl
@@ -58,6 +58,7 @@ return ($err, $warn);
 # Load Perl modules needed by S3 (which are included in Virtualmin)
 sub require_s3
 {
+return if (&has_aws_cmd());
 foreach my $m (@s3_perl_modules) {
 	eval "use $m";
 	die "$@" if ($@);


save_s3.cgi
```diff
@@ -54,7 +54,7 @@ else {
 		}
 
 	# Validate that it works
-	my $buckets = &s3_list_buckets($s3);
+	my $buckets = &s3_list_buckets($s3->{'access'}, $s3->{'secret'});
```
jcameron, line 57: Why is this needed? This function should take the s3 object has a valid parameter..

iliajie: > This function should take the s3 object has a valid parameter

It isn't. It expects single params, no?

https://github.com/virtualmin/virtualmin-gpl/commit/0de85d99515818ff35284f62260d93bb6b8830ff#diff-e59314a6bc9dc35ecdc4d6bcfbaf69f2c32042a111cd010aa217798974e96338R521

jcameron: No, the functions the params get passed down to can take an s3 object.

What bug does this change fix?

iliajie: > No, the functions the params get passed down to can take an s3 object.

Based on code and description the function accepts single params only!

image

Double check please!

jcameron: Trust me on this - all the functions call down to `get_s3_account` which accepts either an access key, or an s3 account object.

What goes wrong without this line change?

iliajie: > all the functions call down to `get_s3_account` which accepts either an access key

Well, it seem what you're expecting is right, except one case scenario in `save_s3.cgi` file (very end of the screenshot), e.g.:

image

`list_buckets.cgi` does it right too! However, `save_s3.cgi` sends a hash reference of `$s3`.

> What goes wrong without this line change?

`save_s3.cgi` when called on the new account, constructs `$s3` as a _hash reference_ (not array reference that you could send as `@$s3`) on line 30-34, and later the same hash reference sent to `s3_list_buckets`.

Here are the dumps:

#### `$s3` when sent from `save_s3.cgi`:

```
{
  'access' => 'X1XX0XXXXXXXXXXXXX',
  'desc' => 'S3 default',
  'id' => '171691669236956',
  'location' => 'eu-north-1',
  'secret' => '8ZFNZc8ZFN8ZFNkJAIiCXrmNCzAIiCXrmNCOnN'
}
```

#### `$akey` when received in `s3_list_buckets`:

```
{
  'access' => 'X1XX0XXXXXXXXXXXXX',
  'desc' => 'S3 default',
  'id' => '171691669236956',
  'location' => 'eu-north-1',
  'secret' => '8ZFNZc8ZFN8ZFNkJAIiCXrmNCzAIiCXrmNCOnN'
}
```

#### `$skey` when received in `s3_list_buckets`:

```
undef
```

jcameron: That all looks OK to me? What I mean is, what goes wrong from the user's point of view without this change?

iliajie: > That all looks OK to me?

Alright, then perhaps the comment for `s3_list_buckets` should be improved?

I don't understand why we pass different parameter types, i.e., sometimes a string and sometimes a hash. Why not always make `s3_list_buckets` receive a hash reference with `access` and `secret` keys? Why couldn't the caller just send `{ access => 'QWERTY', secret => 'ASDFG' }` instead of making it polymorphismic? 🙂 

> What goes wrong from the user's point of view without this change?

From the user's point of view, it just works fine!

iliajie: I have reverted the last patch. Though there is still one commit in this PR that's worth merging, I think?

jcameron: Yeah the comments really need to be fixed. This API went through a lot of iterations while I was adding multiple S3 provider support, and the docs weren't kept up to date.

iliajie: > This API went through a lot of iterations while I was adding multiple S3 provider support, and the docs weren't kept up to date.

Ah, then that explains it! Thanks!

PR #825, opened 2024-06-07
Add support for proxying additional backend types

iliajie:
Hey Jamie!

As Joe pointed out earlier, we need to support different backend types, particularly Unix sockets. This PR implements support for that.


jcameron:
This could be cool, but let's hold off on new features until we do a new release of Virtualmin.

Also, it seems dangerous to allow non-root users to create Unix socket proxies.

iliajie:
> This could be cool, but let's hold off on new features until we do a new release of Virtualmin.

This is a tiny change rather than a new big feature. Though wanted!
 
> Also, it seems dangerous to allow non-root users to create Unix socket proxies.

I'm not sure what you mean, but Unix sockets are much more secure, as they aren't available over a network and can only be used by processes with the right permissions.

So, why is this dangerous? What can a Unix socket proxy do that a TCP/IP socket proxy cannot?

jcameron:
What I mean is, do we really want a domain owner to be able to configured Apache to connect to unix sockets? And even if we do, I think that this would be a very niche feature that the typical domain owner can't understand anyway. So we should limit it to root for now..

iliajie:
> So we should limit it to root for now

I have fixed that, alongside with another bug.

Although, perhaps I could improve it further. I can parse the socket file path, check if it exists, and if it does, verify if it is owned by the domain owner user and/or located under the user's home directory?

iliajie:
> > So we should limit it to root for now
> 
> I have fixed that, alongside with another bug.
> 
> Although, perhaps I could improve it further. I can parse the socket file path, check if it exists, and if it does, verify if it is owned by the domain owner user and/or located under the user's home directory?

Check this patch https://github.com/virtualmin/virtualmin-gpl/pull/825/commits/89dab8348431e01f7379ec3cca0b2731562e8a23. It will restrict what the domain user can do with the Unix socket file.

jcameron:
Let's just limit this to root instead of adding permission checks ... I don't want to complicate the UI too much with a new socket type.

iliajie:
> Let's just limit this to root instead of adding permission checks ... I don't want to complicate the UI too much with a new socket type.

Done.

edit_balancer.cgi
```diff
@@ -53,20 +53,22 @@ if (($in{'new'} || !$b->{'balancer'}) &&
 		        ".forEach(function(radio) { radio.disabled = ".
 			    "this.checked; }, this);'");
 	}
-
+my $placeholder = "$text{'index_global_eg'} http://127.0.0.1:12345";
+$placeholder .= " $text{'or'} unix:/path/to/socket|http://"
```
jcameron, line 57: What's this http:// doing at the end of the socket path?

iliajie: > What's this http:// doing at the end of the socket path?

Thanks for pointing this out, as this part should be `http://127.0.0.1` instead!

Fixed now!

jcameron: Do we really need to require that the user enter this though? If it's needed for unix domain sockets, we should add it automatically.

iliajie: Require? No, we don't require it. This is a placeholder, a tip for the user what can be entered to the field potentially ..

jcameron: What I mean is, in Apache to proxy to a unix domain socket that `|http://127.0.0.1` suffix is needed in ProxyPass for some reason, even though that's not the address the proxying is to. We should make things easier for users by adding and removing it automatically.

iliajie: I don't think we can do this, as `http://127.0.0.1` in UDS isn't a static value.

jcameron: What is that URL used for then?

iliajie: > What is that URL used for then?

It's used to define the communication backend protocol, which in my view can be anything, like `|wss://12.33.44.55` or `|uwsgi://localhost`.

jcameron: Ah so the IP address in the URL doesn't matter, only the protocol like ws or uwsgi ?

iliajie: > Ah so the IP address in the URL doesn't matter, only the protocol like ws or uwsgi ?

The protocol and IP matters.

jcameron: So from looking at the docs, it seems like only the protocol and path matters : https://httpd.apache.org/docs/current/mod/mod_proxy.html#proxypass

What would the IP/host mean when the connection is always being made to a socket?

jcameron: What I'm really wondering is if there's a more user-friendly way to allow users to proxy to unix domain sockets than expecting them to know this rather confusing Apache syntax...

jcameron: Anyway, let's not hold up this PR on UI details ... we can address that later.

iliajie: > Anyway, let's not hold up this PR on UI details ... we can address that later.

Great!

PR #827, opened 2024-06-14
Fix to always try to determine the maximum username size

iliajie:
Hey Jamie,

I found a long-standing bug: if users skip extra steps in the post-installation wizard, the default MariaDB/MySQL username is limited to 16 characters, despite modern systems supporting 32-128 characters. Specifically:

- Ubuntu 24.04/Debian 12 supports up to 128 characters
- Alma/Rocky 8/9 supports up to 80 characters
- CentOS 7 is limited to 16 characters
- MySQL 8+ supports up to 32 characters.

This patch ensures the correct username length is set for each database system.

I discovered this issue during pre-release testing on Ubuntu 24.04, where skipping the extra steps in wizard resulted in the username being cut off at terrifying 16 characters.

### DIFF
--- a/feature-mysql.pl
+++ b/feature-mysql.pl
@@ -10,6 +10,23 @@ if (!$mysql::config{'login'}) {
 	}
 %mconfig = &foreign_config("mysql");
 $mysql_user_size = $config{'mysql_user_size'} || 16;
+# Try once to determine the maximum username size
+if (!$config{'mysql_user_size'} &&
+    !$config{'mysql_user_size_auto'}) {
+	&lock_file($module_config_file);
+	eval {
+		local $main::error_must_die = 1;
+		my @str = &mysql::table_structure($mysql::master_db, "user");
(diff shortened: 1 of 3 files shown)


feature-mysql.pl
```diff
@@ -10,6 +10,23 @@ sub require_mysql
 	}
 %mconfig = &foreign_config("mysql");
 $mysql_user_size = $config{'mysql_user_size'} || 16;
+# Try once to determine the maximum username size
```
jcameron, line 13: Do this in postinstall.pl rather than here, as that's the better place for these kinds of one-off updates.

iliajie: What if MySQL isn't installed? Should we check for it first?

Perhaps we could just leave it here, as this code will still only be called once?

jcameron: I just don't like to have a lot of things happening in require_mysql, because it can slow down Virtualmin. For example, what if MySQL is running but really slow to respond to queries? This code could hang the whole UI..

PR #829, opened 2024-06-14
Fix redirects to site root

iliajie:
Howdy Jamie!

As discussed earlier, when **Source URL path** is set to `/some-path` and **URL on this website** is set to `/`, it used to produce a false-positive error:

```
Failed to save redirect : Missing or invalid destination URL path
```

This PR fixes it.


jcameron:
Thanks for the fix!

iliajie:
Welcome, Jamie!

PR #831, opened 2024-06-15
Add API to show last login time

iliajie:
Hey Jamie!

This PR adds the last login time to the virtual server summary page and the list domains CLI. Some people evidently want to know when the last login happened for a virtual server and disable it if the last login was over a certain period of time.


virtual-server-lib-funcs.pl
```diff
@@ -20682,6 +20682,70 @@ sub list_script_plugins
...
+my @logins;
+foreach my $k (keys %acl::sessiondb) {
+	next if ($k =~ /^1111111/);
+	next if (!$acl::sessiondb{$k});
+	my ($user, $ltime, $lip) = split(/\s+/, $acl::sessiondb{$k});
+	next if ($user ne $d->{'user'});
+	push(@logins, $ltime);
+	}
+# Mail user logins
+foreach my $user (&list_domain_users($d, 1)) {
```
jcameron, line 20704: Why get the logins for all users if we're only going to show one?

Also, list_domain_users is a pretty slow function ... this would slow down the API and UI.

virtual-server-lib-funcs.pl
```diff
@@ -20682,6 +20682,70 @@ sub list_script_plugins
 return grep { &plugin_defined($_, "scripts_list") } @plugins;
 }
 
+# domain_last_login(&domain)
+# Returns last login for domain
+sub domain_last_login
+{
+my ($d) = @_;
+&foreign_require("acl");
+# Get domain owner user login
```
jcameron, line 20691: Seems like we should actually instead include webmin and usermin logins in the collection that feeds into get_last_login_time

virtual-server-lib-funcs.pl
```diff
@@ -20682,6 +20682,70 @@ sub list_script_plugins
...
+		push(@logins, $ll->{$k});
+		}
+	}
+@logins = sort { $b <=> $a } @logins;
+if (!@logins) {
+	return wantarray ?
+		($text{'users_ll_never'}, undef) :
+		 $text{'users_ll_never'};
+	}
+my $last = &make_date($logins[0], { '_' });
```
jcameron, line 20716: What does this `_` mean??

virtual-server-lib-funcs.pl
iliajie: > Why get the logins for all users if we're only going to show one?

We need to first check all users to later find out the latest login timestamp for a user associated with the domain.

> Also, list_domain_users is a pretty slow function ... this would slow down the API and UI.

Well, it will only be called on the domain summary page and when listing domains using the `--multiline` parameter. Is there a better way to list _only_ the _usernames_ associated with a virtual server?

virtual-server-lib-funcs.pl
iliajie: That would work too, although it won't provide accurate last login details.

virtual-server-lib-funcs.pl
jcameron: But surely what we want is to show the last login time for the domain owner, not *any* mailbox?

virtual-server-lib-funcs.pl
iliajie: Aside from a happy parameter smiley, it asks `make_date` to return a hash. :-)

virtual-server-lib-funcs.pl
iliajie: > But surely what we want is to show the last login time for the domain owner, not _any_ mailbox?

No, we want to show the last login time for _any_ user in the domain. This helps to understand the activity related to the domain.

jcameron: I don't think that information is critical enough to add complexity and slowness to the CLI and UI. And we have this visible in the List Users page already... better to just show on a per-user basis in the `list-users` API command.

iliajie: Well, we need to know the activity happening in the domain that can easily be checked in both the UI and CLI. Having it spread around the list of users has no real value in this context.

By the way, we don't need to get quotas and database users. This fix https://github.com/virtualmin/virtualmin-gpl/pull/831/commits/d6a83ab575e4de9b9172380f68ad15cb935f0398 will make it faster.

jcameron: Who is asking for this feature though? I just don't think a single date for the last time any user in the domain logged in is very useful..

virtual-server-lib-funcs.pl
jcameron: It would certainly be cleaner code though - having this code here look into the Webmin session DB doesn't seem right.

virtual-server-lib-funcs.pl
iliajie: I spoke with @marclaporte today, and he suggested that it would be nice to disable virtual servers on the development server that weren't used for a certain period of time.

Knowing the last activity would be pretty telling, in my opinion.

virtual-server-lib-funcs.pl
iliajie: I could do that. So, to be clear, you're suggesting that we cache the last logins for domain owners and store them to be used from `get_last_login_time`, right?

virtual-server-lib-funcs.pl
jcameron: I mean what's the difference between `{ '_' }` and an empty hash `{ }` ?

virtual-server-lib-funcs.pl
jcameron: I'm thinking that the cleaner thing to do would be to include Webmin and Usermin logins in `update_last_login_times` , which produces a file which can be read much faster.

virtual-server-lib-funcs.pl
iliajie: > I mean what's the difference between `{ '_' }` and an empty hash `{ }` ?

`%{$opts}` in `make_date` will never eval to _true_ unless something is set.

virtual-server-lib-funcs.pl
iliajie: Alright, using caching is better, I agree.

But can those functions involved return a hash or a list, or are they only designed to deal with strings?

virtual-server-lib-funcs.pl
jcameron: Then don't do that! Just check if $opts is not set to undef.

iliajie: Alright, fixed! However, that means this PR cannot be included in the upcoming Virtualmin release.

jcameron: Oh was there a separate PR to change the `make_date` behavior? If so, we can keep this current code until that's out..

virtual-server-lib-funcs.pl
jcameron: I don't get what you mean, sorry..

virtual-server-lib-funcs.pl
iliajie: No, I just fixed that last night. But okay, I can revert it until then.

virtual-server-lib-funcs.pl
iliajie: > I'm thinking that the cleaner thing to do would be to include Webmin and Usermin logins in `update_last_login_times` , which produces a file which can be read much faster.

Jamie, all finished! Please have another look.

virtual-server-lib-funcs.pl
```diff
@@ -20682,6 +20682,65 @@ sub list_script_plugins
...
+@logins = sort { $b <=> $a } @logins;
+if (!@logins) {
+	return wantarray ?
+		($text{'users_ll_never'}, undef) :
+		 $text{'users_ll_never'};
+	}
+my $last = &make_date($logins[0], { '_' }); # XXXX '_' can be removed later
+my $last_login;
+# Show ago for last login if within 24 hours
+if (ref($last) eq 'HASH' &&
```
jcameron, line 20714: Why would this not be a hash?

virtual-server-lib-funcs.pl
```diff
@@ -20682,6 +20682,65 @@ sub list_script_plugins
 return grep { &plugin_defined($_, "scripts_list") } @plugins;
 }
 
+# domain_last_login(&domain)
+# Returns last login for domain
+sub domain_last_login
```
jcameron, line 20687: Let's just limit this to the doamin owner, there's no need to scan through all the domain users for their last logins. Imagine a domain with thousands of mailboxes - this code would make summary_domain.cgi and list-domains.pl really slow.

feature-mail.pl
```diff
@@ -5602,6 +5602,43 @@ sub update_last_login_times
...
+	next if ($k =~ /^1111111/);
+	next if (!$acl::sessiondb{$k});
+	my ($user, $ltime, $lip) = split(/\s+/, $acl::sessiondb{$k});
+	$syslogins{$user} = $ltime
+		if ($ltime > $syslogins{$user} || !$syslogins{$user});
+	}
+# System last logins
+&foreign_require('useradmin');
+my @logins = &useradmin::list_last_logins();
+eval "use Time::Local;";
```
jcameron, line 5625: No need for the first ;

virtual-server-lib-funcs.pl
iliajie: It would never be a hash if `use DateTime` isn't available..

virtual-server-lib-funcs.pl
iliajie: > Let's just limit this to the domain owner, there's no need to scan through all the domain users for their last logins.

Well, I'm not sure about that. We need to know if any associated user logged in as an indication that the domain is active. Otherwise, adding this feature makes very little sense .. Ideally, we need to also track database logins to make it complete.

>  Imagine a domain with thousands of mailboxes - this code would make summary_domain.cgi and list-domains.pl really slow.

Why would it make it slow if we only use cached data from a file? Iterating over a simple `name=value` structure is blazing fast, even for 100,000 records, correct?

feature-mail.pl
iliajie: Thanks! Fixed now!

virtual-server-lib-funcs.pl
jcameron: It would maybe be OK if we didn't have to call list_domain_users. But even then, I think that we should also display the last login time by the domain owner separately.

virtual-server-lib-funcs.pl
jcameron: Hmmm ... I'm not sure I really like that API, as it's kind of confusing to callers if `make_date` can return different types depending on some internal implementation detail. But I suppose it's too late to change now :-(

Are those required Perl modules at least Webmin package dependencies?

iliajie: > I'm not sure I really like that API

Me neither!



> if `make_date` can return different types

It returns strings by default.




> But I suppose it's too late to change now :-(


This is because `make_date` existed long before it was implemented to support what it supports today.



> Are those required Perl modules at least Webmin package dependencies?

Yes, we do!

virtual-server-lib-funcs.pl
iliajie: > It would maybe be OK if we didn't have to call list_domain_users. But even then, I think that we should also display the last login time by the domain owner separately.

I don't understand the reason why! Initially, it was suggested to cache it, which was implemented. Iterating over key-value pairs is fast. What's so bad about knowing the domain's last login time?

I believe that either we implement the last login time for the domain comprehensively, considering all users including database users, or we don't do it at all.

jcameron: Has any customer asked for a way to see a single time which is the last login by *any* user in the domain? I can imagine why the `root` user would want to see the last time a domain owner logged in, and why a domain owner would want to see when each of their mailboxes logged in. But I can't imagine why a single last activity date by any user is useful.

But more importantly, showing this on the domain summary page will be slow because `list_domain_users` can be quite slow. It doesn't make sense to slow this page and `list-domains` down for a niche use case.

iliajie: > Has any customer asked for a way to see a single time which is the last login by _any_ user in the domain?

No, not in particular. Though it is a pretty obvious and potentially useful feature to have for a system that deals with domains.



> But I can't imagine why a single last activity date by any user is useful.

What if I have 1,000 domains created by developers and I want to know which domains haven't been used in the past 3 months, so I can delete them? This is a pretty straightforward need, and something @marclaporte has faced in his environment.



> But more importantly, showing this on the domain summary page will be slow because `list_domain_users` can be quite slow. It doesn't make sense to slow this page and `list-domains` down for a niche use case.

I agree with that. Alright, then perhaps we can factor it all out into a separate page, or just and only add a CLI API and be done with it?

iliajie: > > But more importantly, showing this on the domain summary page will be slow because `list_domain_users` can be quite slow. It doesn't make sense to slow this page and `list-domains` down for a niche use case.
> 
> I agree with that.

Or better yet, we just cache everything to the level where we write the last login time to a file and run `list_domain_users` in the background? This way, the summary page and list domains API would be as fast as before.

iliajie: .. I think this is exactly what I'll do tomorrow. I'll also add an additional column to the list virtual servers page in the UI to sort domains by last login date. Furthermore, I'll make it possible to disable last login collection in the Virtualmin configuration for whatever reason.

jcameron: Yeah I think if we collected this in the background as part of collectinfo.pl it would be acceptable..

iliajie: > Yeah I think if we collected this in the background as part of collectinfo.pl it would be acceptable

Jamie, it's been implemented! Please take a look.

iliajie: Currently, it shows all possible logins as follows.

image

I'm considering showing only the most recent login time and placing all details under a tooltip. At first glance, it seems a bit crowded, but it's also quite informative. What are your thoughts about this in particular?

jcameron: Maybe we could put those under a separate collapsible tab? We could even include the IP of the last login, if we collect that in future...

iliajie: I believe this functionality would be useful in the CLI, but it's too crowded for the UI already .. no?

virtual-server-lib-funcs.pl
```diff
@@ -20682,6 +20682,74 @@ sub list_script_plugins
...
+sub update_domains_last_login_times
+{
+foreach my $d (&list_domains()) {
+	next if ($d->{'alias'});
+	next if ($d->{'no_last_login'});
+	my $last_login_timestamp;
+	my $last_login;
+	my @logins;
+	# Get all users logins timestamps
+	foreach my $user (&list_domain_users($d, 0, 0, 1, 1)) {
```
jcameron, line 20696: You should set the 3rd param to 1 as well, so that list_domain_users doesn't include email virtusers

virtual-server-lib-funcs.pl
```diff
@@ -20682,6 +20682,74 @@ sub list_script_plugins
...
+		foreach my $k (sort { $a cmp $b } keys %$ll) {
+			push(@logins, $ll->{$k});
+			}
+		}
+	# No logins recorded yet
+	if (!@logins) {
+		$last_login_timestamp = undef;
+		$last_login = $text{'users_ll_never'};
+		}
+	# Logins found
```
jcameron, line 20707: Move this comment after the `else` and indent it

virtual-server-lib-funcs.pl
```diff
@@ -20682,6 +20682,74 @@ sub list_script_plugins
...
+	else {
+		# Sort logins and get last login
+		@logins = sort { $b <=> $a } @logins;
+		# Save timestamp
+		$last_login_timestamp = $logins[0];
+		# Show ago for last login if within 24 hours
+		my $last = &make_date($logins[0], { '_' }); # XXXX rm '_' later
+		if (ref($last) eq 'HASH' &&
+		$last->{'ago'} && !$last->{'ago'}->{'days'}) {
```
jcameron, line 20716: Indentation looks wrong

virtual-server-lib-funcs.pl
```diff
@@ -20682,6 +20682,74 @@ sub list_script_plugins
...
+					 $text{'summary_lastlogin_sec'} :
+					 $text{'summary_lastlogin_secs'});
+				}
+			}
+		else {
+			$last_login = &make_date($logins[0]);
+			}
+		}
+	&lock_domain($d);
+	$d->{'last_login'} = $last_login
```
jcameron, line 20744: Wouldn't it be cleaner to just store the last_login_timestamp, and convert to a human-readable string for display? Otherwise it may use the wrong language.

virtual-server-lib-funcs.pl
jcameron: Anyway I'm fine with how it current looks

virtual-server-lib-funcs.pl
iliajie: Done.

virtual-server-lib-funcs.pl
iliajie: Done too.

virtual-server-lib-funcs.pl
iliajie: If this function returned a value, it would. However, since we only save data to the domain config, saving just a timestamp means we'd have to reimplement the 'ago' logic for all callers or create a separate subroutine for it, both of which seem excessive.

virtual-server-lib-funcs.pl
iliajie: Yeah, let's just keep it that way.

virtual-server-lib-funcs.pl
jcameron: No, we should create a separate sub for it that is called when showing the last login time. Always separate presentation (what the user sees) from logic (what we're computing). The current code is clearly wrong, as the text for the last login will be in the wrong language if the root user has a a different language from a domain owner.

iliajie: > The current code is clearly wrong, as the text for the last login will be in the wrong language if the root user has a a different language from a domain owner.

Agreed! Fixed! Thanks for pointing this out!

PR #832, opened 2024-06-17
Add support to disable domain on schedule

iliajie:
Hello Jamie!

There have been a few requests in the past for the ability to disable a virtual server on a simple schedule. I have finally decided to implement this feature. The old Virtualmin Disable module is obsolete and overkill, requiring the installation of the `at` command. This PR implements the feature in a straightforward and simple way.

Please have a look:

image

image


jcameron:
This is an intersting feature, but I think it shouldn't go on the Edit Virtual Server page. Instead it should be either an extra menu item under Disable and Delete, or a new button on the Disable page..

iliajie:
> This is an intersting feature, but I think it shouldn't go on the Edit Virtual Server page. Instead it should be either an extra menu item under Disable and Delete, or a new button on the Disable page..

Well, the point was not to make it a separate link in the menu as it would involve a lot of additional complexity with access and limits.

Having it on the Disable Domain page makes sense, but it also makes it easy to forget when domains are created for development purposes, forcing to setup disable schedule with an additional step.

However, as I think about it now, it would also make sense to add an extra config option to set it automatically at server creation time. Then, having it factored out to the Disable Domain page would be easier, and won't require taking an additional step.

But still, what if a master admin wants to have different domains configured differently? Either way, it seems to add an additional step.

So, maybe having it under the Edit Virtual Server page is fine? It's a tiny and compact feature that won't add any complexity?

jcameron:
No I really disagree strongly that it should go on Edit Virtual Server. We already have a page specifically for disabling domains, and so any new feature to disable on schedule should go there.

iliajie:
Alright, I will move it to Disable Domain page..

iliajie:
This time, I approached the problem with proper design considerations in mind.

I moved the functionality to the **Disable Virtual Server** page and added an option to the templates. If an admin wants a domain to be automatically disabled after creation in a certain amount of time, they just need to create a template with this option enabled.

Here are some screenshots for clarity:

image

image

image

Jamie, please review!

marclaporte:
> If an admin wants a domain to be automatically disabled after creation in a certain amount of time, they just need to create a template with this option enabled

This is great!!

jcameron:
> If an admin wants a domain to be automatically disabled after creation in a certain amount of time, they just need to create a template with this option enabled

Fun fact - this kind of feature was responsible for this outage https://arstechnica.com/gadgets/2024/05/google-cloud-accidentally-nukes-customer-account-causes-two-weeks-of-downtime/

disable-domain.pl
```diff
@@ -75,29 +86,72 @@ package virtual_server;
 $d->{'disabled'} && &usage("Virtual server $domain is already disabled");
 @doms = ( $d );
 
-# If disabling sub-servers, find them too
-if ($subservers && !$d->{'parent'}) {
-	foreach my $sd (&get_domain_by("parent", $d->{'id'})) {
-		if (!$sd->{'disabled'}) {
-			push(@doms, $sd);
+if ($scheduled) {
+	if ($schedule eq "show") {
```
jcameron, line 90: There's no need for this functionality here, instead include the scheduled disable time in `list-domains` output

marclaporte:
> this kind of feature was responsible for this outage

We plan to stick to only disabling sites automatically. Any deletions will require a manual review.

marclaporte:
And I trust that Virtualmin remote backups are write-only? So even if we delete a Virtualmin Virtual Server, the backups persist on AWS, etc.?

disable-domain.pl
iliajie: > There's no need for this functionality here, instead include the scheduled disable time in `list-domains` output

Alright, that was removed and I have already added it to `list-domains`.

virtual-server-lib-funcs.pl
```diff
@@ -8924,6 +8930,32 @@ sub call_feature_delete
...
+		$_->{'disabled_auto'} <= time() } &list_domains()) {
+	next if ($d->{'protected'});
+	&push_all_print();
+	&set_all_null_print();
+	&lock_domain($d);
+	eval {
+		local $main::error_must_die = 1;
+		my $disabled_auto = &make_date($d->{'disabled_auto'});
+		&disable_virtual_server($d, 'schedule',
+			&text('disable_autodisabledone', $disabled_auto));
```
jcameron, line 8949: Shouldn't we clear disabled_auto once the disable is done?

virtual-server-lib-funcs.pl
```diff
@@ -8924,6 +8930,32 @@ sub call_feature_delete
 	}
 }
 
+# disable_scheduled_virtual_servers()
+# Disables all scheduled virtual servers that are due to be disabled
+sub disable_scheduled_virtual_servers
+{
+foreach my $d (grep {
+		!$_->{'disabled'} && $_->{'disabled_auto'} &&
+		$_->{'disabled_auto'} =~ /(\d{10,})/ &&
```
jcameron, line 8939: Why would this ever be set to something that's not a number?

disable_domain.cgi
```diff
@@ -8,88 +8,180 @@ $d = &get_domain($in{'dom'});
...
+		my @tabs = ( [ "disable", $text{'disable_domain'},
+				$prog."disable" ],
+			  [ "schedule", $text{'disable_domain_sched'},
+				$prog."schedule" ],
+			);
+		print &ui_tabs_start(\@tabs, "mode", $in{'mode'} || "disable", 1);
 
-&ui_print_footer(&domain_footer_link($d),
-	"", $text{'index_return'});
+		# Details of current cert
```
jcameron, line 113: cert?

virtual-server-lib-funcs.pl
iliajie: Yes, and we do, in `disable_virtual_server` and `enable_virtual_server` since those save the domain.

virtual-server-lib-funcs.pl
iliajie: This is a fail-safe measure to ensure that the value is a timestamp and nothing else, in case it was manually set to something else by the user.

disable_domain.cgi
iliajie: It accidentally ended up here as I copied and pasted the tabs API from the `cert_form.cgi` file.

virtual-server-lib-funcs.pl
jcameron: Where could the user set it to something else though? Presumably we check for a valid input in `disable-domain.pl`

I don't think there's a lot of value in doing these kinds of internal checks of Virtualmin's own data, because if we can't trust it then there are hundreds of places it could go wrong.

virtual-server-lib-funcs.pl
jcameron: Oh cool, thanks!

jcameron:
> And I trust that Virtualmin remote backups are write-only? So even if we delete a Virtualmin Virtual Server, the backups persist on AWS, etc.?

@marclaporte yes, deleting a domain does NOT delete it's backups! Then would only be removed if you setup purging of old backups, or over-write an existing one.

virtual-server-lib-funcs.pl
iliajie: > Where could the user set it to something else though? Presumably we check for a valid input in `disable-domain.pl`
> 
> I don't think there's a lot of value in doing these kinds of internal checks of Virtualmin's own data, because if we can't trust it then there are hundreds of places it could go wrong.

It would only be possible if a user sets it manually by editing domain config file. I don't see how it could be done otherwise (by Virtualmin).

Code like `$_->{'disabled_auto'} =~ /(\d{10,})/` emphasizes that this string is expected to be a timestamp. I don't think it's bad in any way; it just makes the code easier to understand.


> I don't think there's a lot of value in doing these kinds of internal checks of Virtualmin's own data, because if we can't trust it then there are hundreds of places it could go wrong.

Agreed!

jcameron: I mean it's not bad, but I find that unnecessary code is confusing because it makes future readers wonder why this check is needed..

iliajie: > I mean it's not bad, but I find that unnecessary code

This code isn't unnecessary _potentially_  speaking, as it could minimize the chance of having a bug, as mentioned in the Google article you linked earlier.  :--)

> it makes future readers wonder why this check is needed..

I understand your point &mdash; this code isn't there to cover some unexpected behavior on Virtualmin's side if this is your initial concern. On the contrary, having a check for a number of 10 or more digits explicitly clarifies what this code does.

jcameron: Seriously, I think it's best to just remove it. If I was a new programmer on the project and saw that code, I would tear my hair out wondering situation what this check is validating for.

iliajie: > If I was a new programmer on the project and saw that code, I would tear my hair out

LOL! They would have torn their hair out anyway!



> Seriously, I think it's best to just remove it. We certainly can't protect against all possible random edits to domains files!

I could, but I'd rather not. What if something goes wrong and the system's date and time resets to January 1, 1970, 12:00:00 &mdash; it's always best to do sanity check, in particular when we're going to list and then disable all domains.

A new user will Google it. This isn't the most complicated code around the project ..

Is this code really that big of a problem that we're spending so much time on it?

jcameron: Sure this line isn't that important on it's own, but I'm trying to set general principles for the overall Webmin codebase to keep it clean and maintainable.  The point I'm trying to make is that every line of code is a *liability* not an asset, and so we should never write code that is unnecessary. That means that checks for impossible conditions shouldn't be added unless the outcome is catastrophic..

iliajie: > Sure this line isn't that important on it's own, but I'm trying to set general principles for the overall Webmin codebase to keep it clean and maintainable.

This is very well understood, and I share the same principles!



> Sure this line isn't that important on it's own, but I'm trying to set general principles for the overall Webmin codebase to keep it clean and maintainable. The point I'm trying to make is that every line of code is a _liability_ not an asset, and so we should never write code that is unnecessary.

Yes, I perfectly agree!

> That means that checks for impossible conditions shouldn't be added unless the outcome is catastrophic..

This particular case represents a catastrophic outcome if things go wrong. I don't want to leave any chance for this to happen, which is why I added this check in the first place. As we don't use warnings for this library and don't need support for timestamps after November 20, 2286 (what if we do?), I have simplified the code a bit. However, there's no guarantee that we will make it past January 19, 2038 anyways ..  :--)

jcameron: What if we just checked that `disabled_auto` was non-empty? That would cover by far the most common failure case of the field not being set..

iliajie: > What if we just checked that `disabled_auto` was non-empty?

As I previously explained, that could be potentially fragile. I don't want it. I want the check to be explicitly performed for a timestamp.

jcameron: Can we at least extract this out into a function called `is_timestamp` then, which checks if it's a reasonable number? That will be easier to read.

iliajie: > Can we at least extract this out into a function called `is_timestamp` then, which checks if it's a reasonable number? That will be easier to read.

Yes, this is an excellent idea! Implemented here https://github.com/virtualmin/virtualmin-gpl/pull/832/commits/a17bbb9634ce1f12609b2ca2cb82c6b68197ecfc.

Thanks!

PR #834, opened 2024-06-20
Add ability to test command if it starts with the path

iliajie:
https://github.com/virtualmin/virtualmin-gpl/commit/a0f141955798a5e397576b4db5f3d1d04584eafa#commitcomment-143323102

### DIFF
--- a/scripts-lib.pl
+++ b/scripts-lib.pl
@@ -3350,7 +3350,8 @@ my ($ver) = @_;
 my $basever = substr($ver, 0, 1);
 my @opts = ( $config{'python_cmd'} );
 if (!$basever || $basever == 3) {
-	push(@opts, "python3.14", "python314",
+	push(@opts, "/etc/alternatives/python3",
+		    "python3.14", "python314",
 		    "python3.13", "python313",
 		    "python3.12", "python312",
 		    "python3.11", "python311",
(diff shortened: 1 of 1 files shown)


scripts-lib.pl
```diff
@@ -3362,13 +3363,22 @@ sub get_python_path
...
+		    "python2.7", "python27",
 		    "python2.6", "python26",
 		    "python2", "python20");
 	}
 push(@opts, "python");
 foreach my $o (@opts) {
-	my $p = &has_command($o);
+	my $p;
+	if ($o =~ /^\//) {
+		my ($path, $cmd) = ($o =~ m|^(.*)/([^/]+)$|);
```
jcameron, line 3375: I don't understand what this code is doing ... has_command will still work if given an absolute path

iliajie: > has_command will still work if given an absolute path

It didn't work for me in my initial tests.

> I don't understand what this code is doing

This is simple. I just split the path and command, then set the path only for this if block for `has_command` to test if `/etc/alternatives` directory exclusively has `python3` command ..

jcameron: There's no need to do that split though - if you pass has_command an absolute path, it will just check to see if it is executable.

iliajie: > There's no need to do that split though

You're right! I overlooked it last night.

PR #843, opened 2024-06-29
Fix to clear Dovecot config cache if creating and deleting on the same call

iliajie:
https://forum.virtualmin.com/t/re-lets-encrypt-apache-website-an-ipv6-dns-record-mydomain-org-with-address-1-exists-but-this-virtual-server-does-not-have-ipv6-enabled/125447/45?u=ilia

### DIFF
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -20447,6 +20447,7 @@ return &$err(join(" ", @warns)) if (@warns);
 # Create the server
 &push_all_print();
 &set_all_null_print();
+$main::setup_virtualmin_default_hostname_ssl_created++;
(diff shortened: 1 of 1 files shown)


jcameron:
No let's not do this, I'd prefer to actually fix this properly by making sure the cache is correct.

iliajie:
Yeah, that's what I though too, and as I mentioned on the thread I thought it to be rather a work-around ..

shoulders:
@jcameron and @iliajie 

I cannot work out if this should be closed because it was fixed another way instead of using the pull request

This is a couple of months old and maybe should be closed.

jcameron:
We don't need this anymore because I fixed the underlying issue...

PR #844, opened 2024-06-30
Fix to set error for initial LE creation if errors are shown

iliajie:
This PR will set an error to be accessible and shown outside of `create_initial_letsencrypt_cert` sub.

### DIFF
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -8556,8 +8556,10 @@ if ($valid) {
 				push(@estr, $e->{'desc'}." : ".
 					    &html_strip($e->{'error'}));
 				}
-			&$second_print(&text('letsencrypt_evalid',
-				&html_escape(join(", ", @estr))));
+			my $err = &html_escape(join(", ", @estr));
+			$virtual_server::create_initial_letsencrypt_cert::error =
+				$err;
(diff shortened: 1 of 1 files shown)


virtual-server-lib-funcs.pl
```diff
@@ -8556,8 +8556,10 @@ sub create_initial_letsencrypt_cert
 				push(@estr, $e->{'desc'}." : ".
 					    &html_strip($e->{'error'}));
 				}
-			&$second_print(&text('letsencrypt_evalid',
-				&html_escape(join(", ", @estr))));
+			my $err = &html_escape(join(", ", @estr));
+			$virtual_server::create_initial_letsencrypt_cert::error =
```
jcameron, line 8560: Don't use global variables like this to pass the error around - it's bad programming practice. Instead just have it returned from the `create_initial_letsencrypt_cert` function.

iliajie: I know. But It has to be returned by create domain function, which would involve more significant changes.

jcameron: Best to put it into a field in the domain hash then. Using global variables as a "secret" return value is never a good idea..

iliajie:
> Best to put it into a field in the domain hash then.

Agreed. Check this https://github.com/virtualmin/virtualmin-gpl/pull/847 PR instead!

PR #847, opened 2024-07-01
Fix to always store last LE error

iliajie:
Hey Jamie!

This PR suppresses #844 and always stores the LE error in the domain config file, making it accessible at all times on the **SSL Certificate** page to avoid obscurity.

To be clear, even though we save these errors, they are not displayed in the UI upon creation time unless enabled in the Virtualmin configuration page explicitly.

### DIFF
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -20474,7 +20474,10 @@ $succ = 2 if ($rs->{'ssl_same'});
 my $elelast;
 if ($config{'err_letsencrypt'}) {
 	$elelast = $rs->{'letsencrypt_last_err'};
-	$elelast =~ s/\t/\n/g, $elelast = " : $elelast" if ($elelast);
+	if ($elelast) {
+		$elelast =~ s/\t/\n/g;
+		$elelast = " : $elelast";
+		}
 	}
 my $succ_msg = $succ ? 
 	&text($succ == 2 ? 'check_defhost_sharedsucc' : 'check_defhost_succ', $system_host_name) :
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -8550,14 +8550,18 @@ if ($valid) {
 		}
 	my @errs = &validate_letsencrypt_config($d, $vcheck);
 	if (@errs) {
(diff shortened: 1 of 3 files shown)


virtual-server-lib-funcs.pl
```diff
@@ -20458,9 +20471,14 @@ sub setup_virtualmin_default_hostname_ssl
 my $succ = $rs->{'letsencrypt_last'} ? 1 : 0;
 # Perhaps shared SSL certificate was installed, trust it
 $succ = 2 if ($rs->{'ssl_same'});
+my $elelast;
+if ($config{'err_letsencrypt'}) {
+	$elelast = $rs->{'letsencrypt_last_err'};
+	$elelast =~ s/\t/\n/g, $elelast = " : $elelast" if ($elelast);
```
jcameron, line 20477: This line is confusing, please split it up

iliajie: Fixed.

virtual-server-lib-funcs.pl
```diff
@@ -20458,9 +20471,17 @@ sub setup_virtualmin_default_hostname_ssl
 my $succ = $rs->{'letsencrypt_last'} ? 1 : 0;
 # Perhaps shared SSL certificate was installed, trust it
 $succ = 2 if ($rs->{'ssl_same'});
+my $elelast;
+if ($config{'err_letsencrypt'}) {
```
jcameron, line 20475: What does this config option do, and where is it defined? Would we ever not want to show the LE error?

iliajie: > What does this config option do, and where is it defined?

It's defined on the **Virtualmin Configuration → SSL Settings** page.

> Would we ever not want to show the LE error?

Some time ago, we decided not to show irritating Let's Encrypt errors during the initial domain creation time to improve the user experience.

jcameron: Ah ok .. is it on or off by default?  If it's off by default, it seems likely users will never enable this before the default domain is created.

iliajie: Well, yes, but at least there will be a clearer way to debug it!

PR #852, opened 2024-07-08
Add ability to display system information for plugins

iliajie:
Related https://github.com/virtualmin/virtualmin-nginx/pull/72.

### DIFF
--- a/collect-lib.pl
+++ b/collect-lib.pl
@@ -172,7 +172,7 @@ foreach my $f ("virtualmin", @features) {
 			}
 		}
 	}
-foreach my $f (split(" ", $config{'plugins'})) {
+foreach my $f (@plugins) {
 	if (&plugin_defined($f, "feature_sysinfo")) {
 		$f =~ s/-/_/g;
 		my $ifunc = "${f}::feature_sysinfo";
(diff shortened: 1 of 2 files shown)


collect-lib.pl
```diff
@@ -172,6 +172,13 @@ sub collect_system_info
 			}
 		}
 	}
+foreach my $f (split(" ", $config{'plugins'})) {
```
jcameron, line 175: There's an @plugins global array for this, or @feature_plugins

collect-lib.pl
```diff
@@ -172,6 +172,13 @@ sub collect_system_info
...
+	if (&plugin_defined($f, "feature_sysinfo")) {
+		$f =~ s/-/_/g;
+		my $ifunc = "${f}::feature_sysinfo";
```
jcameron, line 178: There's a plugin_call function that does the module to package mapping

collect-lib.pl
iliajie: Thanks! Fixed now!

collect-lib.pl
iliajie: Oh, that's right! Great! Thanks! Fixed now!

PR #858, opened 2024-07-13
Fix not to leak personal details when crating default self-signed SSL certificate

iliajie:
Hey Jamie!

This patch ensures that private details won't be leaked when automatically generating a self-signed SSL certificate.

### DIFF
--- a/feature-ssl.pl
+++ b/feature-ssl.pl
@@ -411,10 +411,10 @@ if ($d->{'dom'} ne $oldd->{'dom'} && &self_signed_cert($d) &&
 		$info->{'c'},
 		$info->{'st'},
 		$info->{'l'},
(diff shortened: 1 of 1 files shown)


jcameron:
I'm not opposed to this PR, but wouldn't the same thing happen when generating an LE cert?

iliajie:
> I'm not opposed to this PR, but wouldn't the same thing happen when generating an LE cert?

What thing? Let's Encrypt doesn't add user email or any other potentially personal details. The description for the user can be private in nature.

jcameron:
Oh right .. in that case, this looks fine to me.

marclaporte:
Thank you @iliajie and @jcameron !

PR #887, opened 2024-08-17
Fix to update domain users shell and home accordingly

iliajie:
Hey Jamie!

This PR addresses the issue of domain users not being updated in jail, which is serious.

It has been thoroughly tested. I was considering making a direct commit, but I thought you might want to discuss a few details first.

### DIFF
--- a/jailkit-lib.pl
+++ b/jailkit-lib.pl
@@ -64,6 +64,9 @@ foreach $f (&mount::files_to_lock()) {
 	&unlock_file($f);
 	}
 
+# Create a fake /etc/passwd file in the jail
+&create_jailkit_passwd_file($d);
+
 # Modify the domain user's home dir and shell
 &require_useradmin();
 foreach my $uinfo (&list_domain_users($d, 0, 0, 1, 1, 0)) {
@@ -81,9 +84,6 @@ foreach my $uinfo (&list_domain_users($d, 0, 0, 1, 1, 0)) {
 	&foreign_call($usermodule, "modify_user", $olduinfo, $uinfo);
 	&foreign_call($usermodule, "made_changes");
 	}
-
-# Create a fake /etc/passwd file in the jail
-&create_jailkit_passwd_file($d);
 
 # Set chroot for all domains' PHP-FPM configs
 foreach my $pd ($d, &get_domain_by("parent", $d->{'id'})) {
@@ -239,17 +239,13 @@ foreach my $sd ($d, &get_domain_by("parent", $d->{'id'})) {
 # Write out chosen users to the jail passwd file
 my $pfile = $dir."/etc/passwd";
 my $sfile = $dir."/etc/shadow";
+my %jail_shell =
+	map { $_->{'user'} => $_->{'shell'} } &get_domain_jailed_users_shell($d);
 &open_lock_tempfile(PASSWD, ">$pfile");
 &open_lock_tempfile(SHADOW, ">$sfile");
 foreach my $u (@ucreate) {
 	my $shell = $u->{'shell'};
-	if ($shell =~ /\/jk_chrootsh$/) {
-		# Put back real shell
-		my $ushell = $d->{"unjailed_shell_$u->{'user'}"};
-		$shell = $u->{'domainowner'} ? 
-				$ushell || $d->{'unjailed_shell'} :
-				$ushell || "/bin/false";
-		}
+	$shell = $jail_shell{$u->{'user'}} if ($shell =~ /\/jk_chrootsh$/);
 	my $home = $u->{'home'};
 	$home =~ s/^\Q$dir\E\/\.//;
 	my @pline = ( $u->{'user'}, "x", $u->{'uid'}, $u->{'gid'},
@@ -273,37 +269,25 @@ foreach my $g (@gcreate) {
 &close_tempfile(GROUP);
 }
 
-# modify_jailkit_users(&domain, [username], [old-username])
-# Sync user real shell when user added or modified after jail creation
-sub modify_jailkit_users
+# modify_jailkit_user(&domain, username)
+# Fix up the shell for the actual user domain
+sub modify_jailkit_user
 {
-my ($d, $uname, $olduname) = @_;
-# Sync user real shell when user added or modified after jail creation
-if ($d->{'jail'} && $uname) {
-	my $dir = &domain_jailkit_dir($d);
-	my ($uinfo) = grep { $_->{'user'} eq $uname } &list_all_users();
-	if ($uinfo) {
-		my $ushell = $uinfo->{'shell'};
-		&lock_domain($d);
-		delete($d->{"unjailed_shell_$olduname"}) if ($olduname);
-		$d->{"unjailed_shell_$uname"} = $ushell
-			if ($ushell !~ /\/jk_chrootsh$/);
-		&save_domain($d);
-		&unlock_domain($d);
-		my $olduinfo = { %$uinfo };
-		if ($uinfo->{'shell'} !~ /\/jk_chrootsh$/) {
-			$uinfo->{'shell'} = &has_command("jk_chrootsh") ||
-					    "/usr/sbin/jk_chrootsh";
-			}
-		$uinfo->{'home'} =~ s/^\Q$dir\E\/\.//;
-		$uinfo->{'home'} = $dir."/.".$uinfo->{'home'};
-		&foreign_call($usermodule, "set_user_envs", $uinfo,
-			'MODIFY_USER', $plainpass, [], $olduinfo);
-		&foreign_call($usermodule, "making_changes");
-		&foreign_call($usermodule, "modify_user", $olduinfo, $uinfo);
-		&foreign_call($usermodule, "made_changes");
-		}
-	}
+my ($d, $user) = @_;
+return if (!$d->{'jail'});	# Jailing isn't enabled
+my $dir = &domain_jailkit_dir($d);
+return if (!-d $dir);		# Jailing isn't enabled
+return if (!-d "$dir/etc");	# Jail directory is invalid
+my $olduser = { %$user };
+$user->{'shell'} = &has_command("jk_chrootsh") ||
+			"/usr/sbin/jk_chrootsh";
+$user->{'home'} = $dir."/.".$user->{'home'}
+	if ($user->{'home'} !~ /^\Q$dir\E/);
+&foreign_call($usermodule, "set_user_envs", $user,
+	'MODIFY_USER', $plainpass, [], $olduser);
+&foreign_call($usermodule, "making_changes");
+&foreign_call($usermodule, "modify_user", $olduser, $user);
+&foreign_call($usermodule, "made_changes");
 }
 
 # copy_jailkit_files(&domain, [dir])
--- a/jailkit-lib.pl
+++ b/jailkit-lib.pl
@@ -66,22 +66,29 @@ foreach $f (&mount::files_to_lock()) {
 
 # Modify the domain user's home dir and shell
 &require_useradmin();
-my ($uinfo) = grep { $_->{'user'} eq $d->{'user'} } &list_all_users();
-if (!$uinfo) {
-	return &text('jailkit_euser', $d->{'user'});
+foreach my $duser (map { $_->{'user'} } &list_domain_users($d, 0, 0, 1, 1, 0)) {
+	my ($uinfo) = grep { $_->{'user'} eq $duser } &list_all_users();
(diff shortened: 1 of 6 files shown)


jcameron:
Looks reasonable but does it set the same shell that would be used if a jailed user is created on the Edit Users page?

iliajie:
> does it set the same shell that would be used if a jailed user is created on the Edit Users page?

Well, it didn't yet! But I dug in more and here is the fix https://github.com/virtualmin/virtualmin-gpl/pull/887/commits/6dd29389179de6b2f488b0891c1fe18cb1b25e5d.

However, there’s still one issue (separate) I just realized we have, which, in my opinion, is not user-friendly at all. We’re showing a "`Has shell /sbin/jk_chrootsh`" for all jailed users, which is technically correct but doesn’t make sense and has no practical value.

image

I suggest, instead of reading the default `/etc/passwd` for jailed users in domain, we should look into `/home/chroot ... /etc/passwd` instead, as inside the jail, the correct shell is already set.

That said, for jailed domains, we should not display in the UI the shell currently set in the main `/etc/passwd` file, including on the **Edit User** page. Instead, we should use the `passwd` file from the jailed environment.

Could you fix that after merging this PR?

jcameron:
Yes I agree that the shell inside the chroot should be shown, not the `jk_chrootsh` !

iliajie:
Thanks, Jamie!

Would you take care of it, or I could do it within this PR? Or you want me to open a new PR?

jailkit-lib.pl
```diff
@@ -66,22 +66,29 @@ sub enable_domain_jailkit
...
-if (!$uinfo) {
-	return &text('jailkit_euser', $d->{'user'});
-	}
-my $olduinfo = { %$uinfo };
-if ($uinfo->{'shell'} !~ /\/jk_chrootsh$/) {
-	$d->{'unjailed_shell'} = $uinfo->{'shell'};
-	$uinfo->{'shell'} = &has_command("jk_chrootsh") ||
-			    "/usr/sbin/jk_chrootsh";
+foreach my $duser (map { $_->{'user'} } &list_domain_users($d, 0, 0, 1, 1, 0)) {
+	my ($uinfo) = grep { $_->{'user'} eq $duser } &list_all_users();
```
jcameron, line 70: Why fetch the users in the outer loop, and then fetch all users again inside here? If you just remove `map { $_->{'user'} }` you'll have the `$uinfo` hash ref already

jailkit-lib.pl
```diff
@@ -153,25 +160,32 @@ sub disable_domain_jailkit
...
-	return &text('jailkit_euser', $d->{'user'});
-	}
-my $olduinfo = { %$uinfo };
-if ($uinfo->{'shell'} =~ /\/jk_chrootsh$/) {
-	my $tmpl = &get_template($d->{'template'});
-	my $defshell = $tmpl->{'ushell'};
-	if ($defshell eq 'none' || !$defshell) {
-		$defshell = &default_available_shell('owner');
+foreach my $duser (map { $_->{'user'} } &list_domain_users($d, 0, 0, 1, 1, 0)) {
+	my ($uinfo) = grep { $_->{'user'} eq $duser } &list_all_users();
```
jcameron, line 164: Same comment here as above

jailkit-lib.pl
iliajie: Because we only need _domain users_ not all users on the system. And we use `&list_all_users()` to get Unix user info into `$uinfo` .. right?

jcameron: But list_domain_users returns that same uinfo Unix user info

iliajie: Oh, it does!? Thanks! Okay, I'll take a closer look a bit later .. working on the Workbench now ...

iliajie: > But list_domain_users returns that same uinfo Unix user info

Got it! Fixed now! Thanks!

iliajie:
> Yes I agree that the shell inside the chroot should be shown, not the `jk_chrootsh` !
>> Would you take care of it, or I could do it within this PR? Or you want me to open a new PR?

No worries, Jamie! This has already been implemented in the https://github.com/virtualmin/virtualmin-gpl/pull/887/commits/cb19321822397c8dd058029e89b0fa9560dec5cf and https://github.com/virtualmin/virtualmin-gpl/pull/887/commits/a5b451e401cc395d80729950bd66af0cb187796e commits, both under the same PR.

iliajie:
Jamie, I added new fixes! Please check.

iliajie:
While running the final tests, I discovered a small bug that caused imprecision when enabling or disabling jail — the shell for the domain owner wasn’t restored correctly. Now it's impeccable!

jailkit-lib.pl
```diff
@@ -66,22 +66,30 @@ sub enable_domain_jailkit
...
-	return &text('jailkit_euser', $d->{'user'});
-	}
-my $olduinfo = { %$uinfo };
-if ($uinfo->{'shell'} !~ /\/jk_chrootsh$/) {
-	$d->{'unjailed_shell'} = $uinfo->{'shell'};
-	$uinfo->{'shell'} = &has_command("jk_chrootsh") ||
-			    "/usr/sbin/jk_chrootsh";
+foreach my $uinfo (&list_domain_users($d, 0, 0, 1, 1, 0)) {
+	my $duser = $uinfo->{'user'};
+	if (!$uinfo) {
```
jcameron, line 71: How can $uinfo ever not be set if we get it back from `list_domain_users` ?

jailkit-lib.pl
```diff
@@ -354,4 +397,22 @@ sub copy_all_domain_jailkit_files
...
+# Get the actual jailed users info in the domain
+sub get_domain_jailed_users_shell
+{
+my ($d) = @_;
+return () if (!$d->{'jail'});
+my $dir = &domain_jailkit_dir($d);
+my $pfile = $dir."/etc/passwd";
+return () if (!-r $pfile);
+&foreign_require("useradmin");
+local $useradmin::config{'passwd_file'} = $pfile;
```
jcameron, line 410: Messing with the cache and useradmin config like this seems dangerous. Even though it means repeating code, it would be cleaner to read /etc/passwd in the chroot directly and have this function just return a list of usernames and shells.

iliajie: I can fix that! Thanks!

Though, could you please explain why it's _exactly_ dangerous?

jcameron: Because it's going around the back of the API, and making assumptions about what config variables and global variables are used by the `list_users` method.

iliajie: Thanks! Fixed here https://github.com/virtualmin/virtualmin-gpl/pull/887/commits/20159155fc3a4d9a74c7a504eead2a292c9ef58a.

jailkit-lib.pl
iliajie: Yes, this can happen when re-enabling or disabling jail for the domain, as we deal with additional domain users.

jcameron: I don't understand - how can `list_domain_users` ever return `undef` ?

iliajie: > I don't understand - how can `list_domain_users` ever return `undef` ?

Well, could we rewind a bit and remember why it was there originally and what made it to be there in the first place?

jcameron: Well the original code is here I think : https://github.com/virtualmin/virtualmin-gpl/blob/master/jailkit-lib.pl#L69

In this case it just checks if the domain owner user exists or not, but in this loop since its going through all users there's no need to check.

iliajie: Okay, how about we just remove it https://github.com/virtualmin/virtualmin-gpl/pull/887/commits/0bc412758ebf8299db5b08f69cb50c4a3c6731fd?

iliajie:
All good or there is anything else?

iliajie:
Also, everything worked smoothly when renaming the domain, including changing the username and home directory ..

jailkit-lib.pl
```diff
@@ -275,6 +276,32 @@ sub create_jailkit_passwd_file
...
 	}
 &close_tempfile(GROUP);
+
+# Sync user real shell when user added or modified after jail creation
+if ($d->{'jail'} && $uname) {
+	my ($uinfo) = grep { $_->{'user'} eq $uname } &list_all_users();
+	if ($uinfo) {
+		my $ushell = $uinfo->{'shell'};
+		&lock_domain($d);
+		delete($d->{"unjailed_shell_$olduname"}) if ($olduname);
```
jcameron, line 286: Why do we need to store unjailed_shell for each user in the domain's config? Can't it just be read from the /etc/passwd inside the jail?

iliajie: Well, yes! The current design uses the saved entries from `/etc/passwd` to pass them to the new jail.

I don't think we should rework this as part of this change.

jcameron: This PR is mostly good, but I am confused about why this function is being used to sync all users, but also has special parameters for a specific user?? Why are these two use cases being mixed?

iliajie: Do you think we should do it separately?

jcameron: Yes absolutely! Can you explain in what cases this function is being called with a user as parameter, and why?

iliajie: > Yes absolutely!

Alrigthy!

Check this patch out https://github.com/virtualmin/virtualmin-gpl/pull/887/commits/6e51148e7f73fdc23f366bd00afd48a3362b9bf9.

> Can you explain in what cases this function is being called with a user as parameter, and why?

We're syncing the user's real shell when the user is added or modified after the jail was created.

jcameron: Oh I see .... but couldn't we just re-sync all users in `create_jailkit_passwd_file` each time a user is added, updated to deleted?

iliajie: We could... Though, is this something that wasn't happening before the last patch?

iliajie: Gosh, that was time consuming to fix!

Anyway, after digging in, I’ve removed the need to store and update real users' shells in the domain config. I also de-coupled the subs you mentioned earlier. Now, I understand what the problem was!

jcameron: Excellent thanks!

iliajie: You're very welcome, Jamie!

iliajie:
Thanks for merging it!

Though, one suggestion for making it completely flawless is to disable and then re-enable the jail before renaming the domain if the jail is initially enabled! Making it work properly for renamed users within the domain would be much harder!

jcameron:
What's the issue with renaming the domain? Can we just not re-generate the passwd file in the jail after the rename to populate it with the new usernames?

iliajie:
> What's the issue with renaming the domain?

We won't have the right shell when a user renamed.

> Can we just not re-generate the passwd file in the jail after the rename to populate it with the new usernames?

We could do various things but most simple and straight forward solution is to disable and re-enable jail, I think ..

jcameron:
Do you know why the shell is incorrect? Surely that doesn't change if the domain is renamed ...

iliajie:
> Do you know why the shell is incorrect? Surely that doesn't change if the domain is renamed ...

After the user is renamed, the mapping with the new username is off ..

jcameron:
Which mapping are you referring to?

iliajie:
Of users with the old usernames that are in jail?

jcameron:
I'm confused, where is this mapping stored? If the domain is renamed and the /etc/passwd in the jail is just re-generated, doesn't that solve the issue?

I can look into this more if you can give me the steps that trigger the problem..

iliajie:
> I'm confused, where is this mapping stored? If the domain is renamed and the /etc/passwd in the jail is just re-generated, doesn't that solve the issue?

Regenerated based on what? We have `jk_chrootsh` set as the shell in the main `/etc/passwd` file at this time, because jail is enabled. As I mentioned from the start, it's definitely doable, but would require more work than simply enabling and disabling the jail.

> I can look into this more if you can give me the steps that trigger the problem..

Well, thanks! But why not just disable and re-enable jail? To reproduce it you need to enable jail for domain and rename the domain ..

jcameron:
Oh I see the issue, the problem is that the real shell is stored in the jail, but when a user is renamed we can't look it up anymore because it's in the old entry in the jail's /etc/passwd. I'll fix that ...

iliajie:
Thanks!

jcameron:
Ok the latest commits should fix the rename issues

iliajie:
Jamie, thanks! This is immaculate now!

PR #893, opened 2024-09-03
Alphabetically sort by name the server templates list

jacob-js:


jacob-js:
> I'm not sure if we want this, as it would mean that the default templates no longer appear first in the list.

Oh, I see! Would it be better to strip off the default templates from the sorting operation so that they are still shown first in the list?

jcameron:
I'd be OK with a sort order that always puts the default templates first and then sorts everything else alphabetically..

jcameron:
Thanks for the patch!

PR #901, opened 2024-09-18
Add support to output JSON and XML in CLI

iliajie:
Hello Jamie!

This PR combines all the commits that were made to the master branch. I didn't expect there to be so many changes initially. I've reverted those and added them to this PR.

### DIFF
--- a/json-lib.pl
+++ b/json-lib.pl
@@ -213,7 +213,8 @@ return Dumper($data);
 
 # cli_convert_remote_format(format)
 # Catches and displays Virtualmin CLI standard listing
-# commands in JSON or XML format
+# commands in JSON or XML format. Returns 1 if the output
+# should be multiline, 0 if not.
 sub cli_convert_remote_format
 {
 my ($format) = @_;
@@ -249,6 +250,13 @@ END {
 		print &convert_remote_format($lines, 0, $program, \%in, $format);
 		}
 	}
+# Should we auto-enable multiline output?
+foreach my $arg (@ARGV) {
+	if ($arg =~ /--(name|id|user|home|file|ip)-only/) {
+		return 0;
+		}
+	}
+return 1;
 }
 
 # execute_webmin_script(command, module, &args, output-fh)
--- a/json-lib.pl
+++ b/json-lib.pl
@@ -23,11 +23,11 @@ else {
 return undef;
 }
 
-# convert_remote_format(&output, exit-status, command, &in, [format])
+# convert_remote_format(&output, exit-status, command, &in, [format], [config])
 # Converts output from some API command to JSON or XML format
 sub convert_remote_format
 {
-my ($out, $ex, $cmd, $in, $format) = @_;
+my ($out, $ex, $cmd, $in, $format, $config) = @_;
 
 # Parse into a data structure
 my $data = { 'command' => $cmd,
@@ -168,7 +168,13 @@ else {
 if ($format) {
 	# Call formatting function
 	my $ffunc = "create_".$format."_format";
-	return &$ffunc($data);
+	# JSON output can be minified or pretty
+	my $pretty;
+	if ($format eq 'json') {
+		$pretty = $in->{'json'} eq 'minified' ? 0 :
+			  ($config->{'json_pretty'} // 1);
+		}
+	return &$ffunc($data, $pretty);
 	}
 else {
 	# Just return perl hash (for internal use)
@@ -185,17 +191,14 @@ eval "use XML::Simple";
 return XMLout($data, RootName => 'api');
 }
 
-# create_json_format(&hash)
+# create_json_format(&hash, [pretty])
 # Convert a hash into JSON
 sub create_json_format
 {
-my ($data) = @_;
+my ($data, $pretty) = @_;
 eval "use JSON::PP";
-my $pretty = $config{'json_pretty'} // 1;
-if (defined($in{'json'})) {
-	$pretty = $in{'json'} eq 'minified' ? 0 : 1;
-	}
-my $coder = JSON::PP->new->pretty($pretty ? 1 : 0);
+$pretty //= 1;
+my $coder = JSON::PP->new->pretty($pretty);
 return $coder->encode($data)."\n";
 }
 
--- a/json-lib.pl
+++ b/json-lib.pl
@@ -235,11 +235,12 @@ END {
 		my $program = $0;
 		$program =~ s/.*\/|\.\w+$//g;
 		my %in;
-		while (my $arg = shift @ARGV_) {
+		local @ARGV = @ARGV;
+		while (my $arg = shift @ARGV) {
 			if ($arg =~ /^--(?!json|xml)([\w-]+)$/) {
 				$in{$1} = 1;
-				$in{$1} = (shift @ARGV_)
-					if (@ARGV_ && $ARGV_[0] !~ /^--/);
+				$in{$1} = (shift @ARGV)
+					if (@ARGV && $ARGV[0] !~ /^--/);
 				}
 			}
 		# Always force multiline format, as JSON and XML output make no
--- a/json-lib.pl
+++ b/json-lib.pl
@@ -191,7 +191,11 @@ sub create_json_format
 {
 my ($data) = @_;
 eval "use JSON::PP";
-my $coder = JSON::PP->new->pretty;
+my $pretty = $config{'json_pretty'} // 1;
+if (defined($in{'json'})) {
+	$pretty = $in{'json'} eq 'minified' ? 0 : 1;
+	}
+my $coder = JSON::PP->new->pretty($pretty ? 1 : 0);
 return $coder->encode($data)."\n";
 }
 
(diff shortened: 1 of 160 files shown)


json-lib.pl
```diff
@@ -191,7 +191,11 @@ sub create_json_format
 {
 my ($data) = @_;
 eval "use JSON::PP";
-my $coder = JSON::PP->new->pretty;
+my $pretty = $config{'json_pretty'} // 1;
+if (defined($in{'json'})) {
+	$pretty = $in{'json'} eq 'minified' ? 0 : 1;
```
jcameron, line 196: You shouldn't access `%in` here, instead pass in whether to minify or not as a parameter.

json-lib.pl
```diff
@@ -204,6 +208,45 @@ sub create_perl_format
...
+# commands in JSON or XML format
+sub cli_convert_remote_format
+{
+my ($format) = @_;
+my ($lines, $fh, $ofh);
+# Redirect STDOUT to a variable
+open ($fh, '>', \$lines) || return;
+# Save the original STDOUT
+$ofh = select($fh);
+END {
```
jcameron, line 222: Defining an END function inside this function is a very confusing way to process the output at the end of the API command. It would be much cleaner and more readable to create a different function that API commands call at the end to convert the saved output to JSON.

json-lib.pl
```diff
@@ -204,6 +208,45 @@ sub create_perl_format
...
+	# In case of error, print the output as is
+	if ($? != 0) {
+		print $lines;
+		}
+	# Otherwise, convert the output to JSON
+	else {
+		my $program = $0;
+		$program =~ s/.*\/|\.\w+$//g;
+		my %in;
+		while (my $arg = shift @ARGV_) {
```
jcameron, line 235: Why parse @ARGV when the format is already a parameter to the outer function?

virtual-server-lib.pl
```diff
@@ -3,7 +3,7 @@
 
 BEGIN { push(@INC, ".."); };
 use WebminCore;
-
+@ARGV_ = @ARGV if (@ARGV && !@ARGV_);
```
jcameron, line 6: Oh god this seems terrible, and unnecessary if you use the approach I suggested above.

json-lib.pl
iliajie: > You shouldn't access `%in` here, instead pass in whether to minify or not as a parameter.

I don’t understand this comment. Could you clarify?

json-lib.pl
iliajie: > Defining an END function inside this function is a very confusing way to process the output at the end of the API command. 

I _don’t_ want it to be implemented that way. I prefer to keep all related logic in one place, and have it called in a _**simple**_ way; changing it will just introduce a lot of unnecessary complexity!

The `END` block is tightly coupled with the function's purpose. Having inline `END` block is common for handling cleanup in Perl!

> Create a different function that API commands call at the end to convert the saved output to JSON.

It has to be called all the time, even on error (non-zero exit).

json-lib.pl
iliajie: Jamie, what do you mean? The `convert_remote_format` sub doesn't get `$format` using `%in`; it uses it to check for other parameters like `name-only` and `id-only`, which need to be tested.

virtual-server-lib.pl
iliajie: > Oh god this seems terrible

It’s not any worse than many other places where global variables are used.

> and unnecessary

It’s against best practices to modify globals like `@ARGV`. The right thing to do would be to keep it intact, so other parts of the script can access it and not find it empty! This is the root of the problem and the reason I had to set `@ARGV_`.

We need to preserve original `@ARGV` because it's being chomped by the program and we cannot access it later.

jcameron: No what I mean is, it shouldn't ever be necessary for the later code to access @ARGV.

json-lib.pl
jcameron: Hmmm .. I feel like there has to be a better way of doing this. Using `END` makes debugging difficult because the control flow will magically jump to this function without it being called.

json-lib.pl
jcameron: %in is not passed into this function as a parameter, so it shouldn't be used. It's bad practice to refer to variables defined in the outer scope inside a function.

virtual-server-lib.pl
iliajie: However, it seems necessary this time, right?

json-lib.pl
iliajie: Yes, it makes debugging more difficult; however, we almost never use the `END` block, and using it once in a blue moon won’t make debugging anyhow harder in practice. I'm not suggesting we start using it regularly. I used it here because I find it easier to handle both successful and failing logic in this specific case scenario.

Besides, there won’t be an issue with it, I don't think, as `cli_convert_remote_format` is only meant to be called _once_.

json-lib.pl
iliajie: Ah, yes! Thanks for the suggestion! The patch for this is here https://github.com/virtualmin/virtualmin-gpl/pull/901/commits/69138fd469e32e7829574e2c9f30c78a7f12095d.

json-lib.pl
jcameron: What I mean is, doesn't `cli_convert_remote_format` already get thge format passed as a parameter? If so, can't you pass this to `END` somehow rather than having it re-parse `@ARGV` ?

iliajie: > What I mean is, doesn't `cli_convert_remote_format` already get thge format passed as a parameter?

Yes, you're right that `format` is passed as a parameter. However, the actual issue is that we need to capture the parameters passed via the command line and pass them further to `convert_remote_format` using the `%in` parameter.

To do this, we need to use the original `@ARGV` array.

json-lib.pl
iliajie: > Why not instead create two functions, one that gets called at the end of successful output and one that gets called from usage (which is called for all errors)? Yes this would be more code that definingEND here as you'd also need to add this to each list-*.pl script, but the control flow would be much clearer.

As I mentioned earlier, I _**really**_ dislike it. It's too much redundant repetitive code for no good reason, as the Perl documentation doesn’t advise against using the `END` block, and there's nothing wrong with it. Unless my implementation introduces an issue, I’d prefer to keep it.

> and the format could be passed in as a parameter.

It has to be fixed, you're right! However, instead, we should follow best practices and avoid altering global variables. I'll work on fixing that so we can eliminate the messy `@ARGV_ = @ARGV if (@ARGV && !@ARGV_);` in `virtual-server-lib.pl`, as I agree it doesn’t look good!

virtual-server-lib.pl
jcameron: Excellent thanks!

json-lib.pl
jcameron: Thanks!

json-lib.pl
jcameron: But isn't there also an END sub defined in web-lib-funcs.pl? Which one will get called if we define two in the same script?

iliajie: Yes, we do! It isn't a problem though, as `END` blocks are executed in reverse order of their definition, i.e. the last one defined will execute first.

virtual-server-lib.pl
iliajie: You're very much welcome!

json-lib.pl
jcameron: Ok let's stick with this approach for now - in future I may re-factor the way CLI commands are defined.

Can you check that the scripts still work when called via remote.cgi in XML mode?

json-lib.pl
```diff
@@ -204,6 +211,54 @@ sub create_perl_format
...
+	# In case of error, print the output as is
+	if ($? != 0) {
+		print $lines;
+		}
+	# Otherwise, convert the output to JSON
+	else {
+		my $program = $0;
+		$program =~ s/.*\/|\.\w+$//g;
+		my %in;
+		local @ARGV = @ARGV;
```
jcameron, line 239: I don't understand why parsing @ARGV again is needed here when you have the format in $format above? 

Also, convert_remote_format could just have `%in` set to `{ 'multiline' => 1 }` since that's the only key it looks at. This way you could also do away with those `local @ARGV` lines in all the API commands.

iliajie: > ... since that's the only key it looks at.

Ah, now I see where the misunderstanding is coming from. `multiline` isn't the only key, as `convert_remote_format` sub also checks for `name-only` and `id-only`.

json-lib.pl
iliajie: > Ok let's stick with this approach for now

Alright, cool!


> Can you check that the scripts still work when called via remote.cgi in XML mode?

I had already done that, and just tried again — everything works flawlessly, at least with the ones I tried with. Do you have any specific scripts in mind, perhaps?

Though, I don’t see how it wouldn’t work, since we never double-pass `--json` and `--xml`, which are filtered out [here](https://github.com/virtualmin/virtualmin-gpl/blob/master/remote.cgi#L67). However, if we did, it could lead to some funny consequences. 🙂

By the way, the remote Virtualmin's XML and JSON outputs are really genius! Thanks for adding those in the first place! And thanks for pushing me towards having a unified JSON format output for CLI. I didn’t even realize we were already doing something similar (i.e., capturing printed output). I also didn’t know much about `socketpair` before seeing your code.

jcameron: Ok cool I was just checking that we wouldn't somehow set the same flag twice.

json-lib.pl
jcameron: Would users ever set those when requesting JSON or XML output though?

iliajie: > Would users ever set those when requesting JSON or XML output though?

Yes, why not? For example:

```
virtualmin list-domains --name-only --json
```

json-lib.pl
iliajie: > See my comments below - I really don't think this is needed, and even if it is re-parsing @argv seems wrong.

This is needed — see [this comment](https://github.com/virtualmin/virtualmin-gpl/pull/901#discussion_r1775083589) please. There's nothing wrong with re-checking what was passed as parameters in the script.

jcameron: It's a problem because we're repeating a lot of logic. I have an idea how to fix this, but I will do that in a follow-up change ... for now let's just merge this one.

iliajie: Thank you, Jamie! I’ll follow up on your change once it’s done!

Issue #906, opened 2024-09-27
Pro: My GPL was downgraded to an old version of Pro


shoulders:
| SYSTEM INFORMATION ||
|----------|----------|
|  OS type and version |  Ubuntu Linux 22.04.5                   |
|  Usermin version     |  2.102                                  |
|  Virtualmin version  |  7.20.2 Pro                             |
|  Theme version       |  21.20.7                                |
|  Apache version      |  2.4.52                                 |
|  Package updates     |  All installed packages are up to date  |

### background

I upgraded my Virtualmin GPL 7.20.2 to Pro

### The issue

When I upgraded from GPL to Pro I expected the same version number but the Pro version, however this did not happen, I got an old version, `7.9.0 Pro`

The upgrade path went as follows:

- started on `7.20.2 (GPL)`
- installed the Pro licence keys + ran upgrade
- I then was on  `7.9.0 Pro`
- there was an upgrade package watiting for virtualmin which i installed
- I am now on `7.20.2.pro-1`

jcameron:
How did you do the upgrade to Pro exactly?

shoulders:
- i went to the menu item `Upgrade to Virtualmin Pro`
- I entered the serial number and key
- I clicked upgrade

iliajie:
We will address this issue in this PR https://github.com/virtualmin/virtualmin-gpl/pull/910.

Did you restore this instance from an old snapshot?

shoulders:
@iliajie did you mean me? Both my `dev` and `live` servers have never been restored from any snapshots. I have also never done any non-standard stuff to them either, so they are vanilla.

iliajie:
Anyway, that PR should address your issue! You can apply that patch using `webmin patch` command and see if that works for you.

shoulders:
Thanks for the fix.

When I installed my Pro license key and serial, Virtualmin was upgraded to `7.9 Pro` with `7.20.2 Pro` waiting in the package updates which I obviously installed immediately and now I am on `7.20.2 Pro`.

So unfortunately I cannot use this patch to see if it works, I reported it because not many people would notice this to report it. 😄 

I assumed it was the update script was pulling an old version of Pro which was the issue. It is almost like it was hardcoded into the upgrade script rather than getting the latest version from the repo (or the equivalent Pro version).

Will this of caused any issues with my installs? I did immediately update to the latest version after converting.

iliajie:
[...]

PR #910, opened 2024-09-28
Fix to clear packages cache before upgrading

iliajie:
Hey Jamie!

This PR address the possible issue described [here](https://github.com/virtualmin/virtualmin-gpl/issues/906#issuecomment-2380460520).


jcameron:
Thanks for the patch!

PR #913, opened 2024-09-30
Enhanced command-line parsing with checking

iliajie:
Hello Jamie!

This PR adds two new subs, `extract_params_from_usage` and `parse_cli_args`, aimed at improving the command-line argument parsing of all Virtualmin sub-commands.

#### Description

1. #### `extract_params_from_usage` sub:
> Extracts a comprehensive hash of available command-line parameters directly from the `usage` sub’s output. This includes identifying parameter names, determining if they require values, and detecting if they can be specified multiple times.

* Benefits:
  * Automation (eliminates manual definitions of parameters, reducing the risk of inconsistencies);
  * Reusability (can be used across multiple sub-programs, to uniform parsing logic);
  * Maintainability (simplifies the process of adding or modifying parameters by relying on the usage text).
  
2. #### `parse_cli_args` sub:
> Parses the command-line arguments (`@ARGV`) based on the parameter definitions extracted by `extract_params_from_usage`. It handles flags, parameters with single or multiple values, and enforces validation by detecting unknown or invalid parameters.

* Features:
  * Dynamic parsing (adapts to changes in parameter definitions without requiring code modifications);
  * Error handling (provides clear error messages for unknown parameters or missing required values);
  * Structured output (returns a hash reference containing all parsed parameters and their associated values or flags, making it easy to access and further process within scripts);
  * Reusability (can be called multiple times).
  
#### Usage Example
```Perl
#!/usr/local/bin/perl

=head1 list-domains.pl

Lists all virtual servers.
# ...

# Extract parameters from usage for the current sub-command (list-domains)
my %params = extract_params_from_usage(\&usage);

# Parse command-line arguments
my $flags = parse_cli_args(\@ARGV, \%params);

# Next, we can either dynamically assign globals using:
${$_ =~ s/-//gr} = $flags->{$_} for keys %{$flags}; # which I don't really like but it would work for now

# Or better yet, we can test it properly, e.g.:
if (exists($flags->{'domain'})) {
    @domains = @{ $flags->{'domain'} };
}
```

#### Output Example
```Bash
~# virtualmin list-domains --domain domain1.com --user-only --multiline --domain domain2.com --domain domain3.com --id-only
```

```Perl
{
  'domain' => [
    'domain1.com',
    'domain2.com',
    'domain3.com'
  ],
  'id-only' => 1,
  'multiline' => 1,
  'user-only' => 1
}
```

You can even incorporate it into your new `parse_common_cli_flags` sub for now if you don't want to dive deeper just yet.

I hope you like it and find it useful.


jcameron:
This is the wrong approach, sorry. Instead of trying to use `usage` to define flags, we should create a data structure that defines all the flags and have both `usage` and some argv parser use it. Something like : 

```
$flags = [
    { 'name' => 'domain', 'multiple' => 1, 'save' => \@dnames },
    { 'name' => 'multiline', 'save' => \$multiline },
    etc..
];
&parse_cli_args(\@ARGV, $flags);
```

iliajie:
> This is the wrong approach, sorry. Instead of trying to use `usage` to define flags, we should create a data structure that defines all the flags and have both `usage` and some argv parser use it. Something like :
> 
> ```
> $flags = [
>     { 'name' => 'domain', 'multiple' => 1, 'save' => \@dnames },
>     { 'name' => 'multiline', 'save' => \$multiline },
>     etc..
> ];
> &parse_cli_args(\@ARGV, $flags);
> ```

Oh, that would be perfect! I thought about it, but it seemed like a lot of work. Now that I’m thinking about it again, I see a way to automate this! Let me figure out how we can batch fix it. I’ll add more commits to this PR...

jcameron:
BTW, while the `usage` sub is usually up to date, I wouldn't count on it to cover all the support args.

iliajie:
Yes, exactly! I’m currently working on extracting everything into an array of hash references and removing the usage sub. With the list of parameters and their descriptions, we can dynamically generate the usage, control, test, and set the parameters as needed.

I’m a bit stuck with parsing correctly the different logic of various usage outputs.

jcameron:
Yeah the function would maybe need to know if a flag has a value or not, or if can appear multiple times.

iliajie:
> This is the wrong approach, sorry. Instead of trying to use `usage` to define flags, we should create a data structure that defines all the flags and have both `usage` and some argv parser use it. Something like :
> 
> ```
> $flags = [
>     { 'name' => 'domain', 'multiple' => 1, 'save' => \@dnames },
>     { 'name' => 'multiline', 'save' => \$multiline },
>     etc..
> ];
> &parse_cli_args(\@ARGV, $flags);
> ```
>> Oh, that would be ideal! I thought about it, but it seemed like a lot of work. Now that I’m thinking about it again, I see a way to automate this! Let me figure out how we can batch fix it. I’ll add more commits to this PR...
>>> Yeah the function would maybe need to know if a flag has a value or not, or if can appear multiple times.

Yep, like this! Take a look please [0757065](https://github.com/virtualmin/virtualmin-gpl/pull/913/commits/0757065aa6c638dbb17e2975f4aa6056f3267135#diff-b5167041acf3279d3a6f2f26858a8cee5fc085346b515d67c6db47092d5d5b7b). This is just the first part! Even though I’ve automated it, it still takes a lot of time to manually review everything.

iliajie:
Jamie, let’s try to speed up our PRs discussions so we can keep things moving forward with more momentum.

jcameron:
Hang on, let's slow down on this! I'm worried that you are writing a ton of code before we have even agreed on the API or correct approach!!

I do like the idea of params being defined as a data structure, which can be used by both a generate `usage` function and a parse that populates variables based on this structure and ARGV. But I would strongly suggest doing this for just one API call first so we can get agreement on what a reasonable design looks like, then expand it.

Also, I strongly doubt that we can automate conversion from old-style `@ARGV` parsing to this new method if we want to do it right.

iliajie:
> But I would strongly suggest doing this for just one API call first so we can get agreement on what a reasonable design looks like, then expand it.

That’s a very valid point, and I agree.

In my previous comment, I added a link to the DNS change subcommand (give it a moment to load). Take a look and let me know if you like the format of the data structure. I like it because it clearly represents the parameters from the old usage sub. Everything should be self-explanatory, except for the nested `values` key – that’s for grouped _required_ alternate parameters, like `–domain` or `–all-domains`.

I have to head to bed now, but I’ll respond tomorrow morning.

Also, if you’re happy with everything, I’ll go ahead and wrap this up. But if you have any major suggestions or concerns, please let me know.



> Also, I strongly doubt that we can automate conversion from old-style `@ARGV` parsing to this new method if we want to do it right.

Most probably we could. Yet not 100% but close to that. To the level where it would be accomplishable in a day or two. 

Please check and share more of your thoughts over it.

Thank you!

jcameron:
I looked at your commit and the change to `modify-dns`, but I don't see how those args defined in `@usage` are actually used by the rest of the script? The goal should be to shift arg parsing AND usage generation to use the same array of params, and to have some function that can do basic validation and assign them to variables.

iliajie:
Jamie, I’ll always aim to respond in my evenings, which are your mornings, so we can keep the discussion rolling and make it more productive overall. Plus, my emails will land right at the top of your inbox stack! 😏

---

> The goal should be to shift arg parsing AND usage generation to use the same array of params, and to have some function that can do basic validation and assign them to variables.

Yes, I think that’s my goal as well!

>  I looked at your commit and the change to `modify-dns`, but I don't see how those args defined in `@usage` are actually used by the rest of the script? 

You’re right, I haven’t fully implemented it yet. Right now, I only have `@usage` constructed for each sub-command. Yet, there are a few minor issues in some automatically constructed `@usage` that I’ll have resolved by the time this PR is ready to merge.

As you might recall, in this PR, I initially added the `parse_cli_args` sub to handle parameter validation by comparing the passed `@ARGV` against the `@usage` structure. The unified usage generator hasn’t been implemented yet.

Would you prefer that I finish this first for `modify-dns` so we can go over it in detail, and you can see how it works before rolling it out to the other sub-commands?

jcameron:
Yes please get this working for one command first so we can review the API and how it is called.

Also, I think a better way to avoid the need for a lot of round-trip code reviews discussing fundamentals of the change is to have a design doc or plan first sketching out how it's going to work. Or at least a skeleton PR that demonstrates the approach, so we can get agreement before going ahead with a lot of code changes.

iliajie:
> Yes please get this working for one command first so we can review the API and how it is called.

Alright, will do later tomorrow! Sorry for the delay!

> Also, I think a better way to avoid the need for a lot of round-trip code reviews discussing fundamentals of the change is to have a design doc or plan first sketching out how it's going to work. Or at least a skeleton PR that demonstrates the approach, so we can get agreement before going ahead with a lot of code changes.

Yes, agreed! As in here #919 by the way.

jcameron:
Thanks! I will also take a look at that FCGI ticket..

jcameron:
Thanks, taking a look now! Is there an example of where `parse_cli_args` is called and the results used in this PR though?

iliajie:
> Is there an example of where `parse_cli_args` is called and the results used in this PR though?

Yes, in `modify-dns.pl`, right here https://github.com/virtualmin/virtualmin-gpl/pull/913/files#diff-b5167041acf3279d3a6f2f26858a8cee5fc085346b515d67c6db47092d5d5b7bR344 (give it time to load).

iliajie:
Updated link https://github.com/virtualmin/virtualmin-gpl/blob/ef1b8f8ebf6cf5168ed6ac264e09ab483139836f/modify-dns.pl#L344.

iliajie:
In fact, the entire `modify-dns.pl` script must be fully functional.

jcameron:
Ok I'll take a look. Maybe you should remove the other API .pl files from this PR, so it's easier to review?

iliajie:
> Maybe you should remove the other API .pl files from this PR, so it's easier to review?

I could just make them load all the time as you suggested earlier?

Don't review the whole PR though! Focus only on `modify-dns.pl` and `commands-lib.pl` for now.

modify-dns.pl
```diff
@@ -120,209 +336,226 @@ package virtual_server;
...
+	&usage($usage, "The BIND DNS server is not enabled for Virtualmin");
 &set_all_text_print();
 
-# Parse command-line args
-while(@ARGV > 0) {
-	local $a = shift(@ARGV);
-	if ($a eq "--domain") {
-		push(@dnames, shift(@ARGV));
+# Return passed parameters or throw an error
+my $flags = &parse_cli_args(\@ARGV, $usage);
```
jcameron, line 344: Ok this API seems like it makers the caller do too much work, because first they have to define that `$usage` struct and then extra the final values from `$flags`. Why not change the structure passed to `parse_cli_args` to include references to the final variables the caller wants the flags put into?  For example something like : 

```
  {
    param => "cloud-dns",
    value => "provider|'local'"
    dest => \$clouddns,
  },
```

iliajie: > Why not change the structure passed to `parse_cli_args` to include references to the final variables the caller wants the flags put into?

First of all, it can’t be automated. And, the `$usage` shouldn’t assign anything or provide a reference to any variable.

But most importantly, I want to separate the logic for parsing parameters passed in the command line, checking the validity of the arguments, and generating help usage.

The assignment of the global variables isn't really necessary, and we iterate over `foreach my $a (keys %{$flags}) {}` just for backward compatibility, as we could directly use `$flags->{'cloud-dns'}` or `@{$flags->{'domain'}}` instead of setting the global `$clouddns` and `@dnames` variable.

iliajie: > It's tempting I agree, but it's not the proper solution

Why not exactly? It’s only going to be generated once, upon transition, and then we can adjust it however we want. 

> validator => sub { my ($f) = @_; $f eq "local" || &is_cloud_dns($f) ? undef : "Invalid cloud DNS provider $f" },

I don’t think it’s realistic to implement without fully reworking over 100 sub-commands.

jcameron: Well we can do them gradually, but in this case it's worth doing right.

iliajie: Alright, no problem. I’ll leave it to you then. I’ll go ahead and close this PR. Thank you!

PR #917, opened 2024-10-06
Add support for creating a domain using a Virtualmin CLI-like way

iliajie:
Hello Jamie!

It was fairly difficult to implement, but I have created a clean way to set up a domain using a simple API, as discussed earlier.

Example of usage:

```Perl
foreign_require('virtual-server');
virtual_server::load_api('create-domain');
my $dom = virtual_server::create_domain_cli({
    domain => 'domain1.example.tld',
    pass => '123',
    unix => 1,
    dir => 1,
    dns => 1,
    web => 1,
    ssl => 1
});
```

This API is fully compatible with the Virtualmin CLI `create-domain` command.


iliajie:
And this is how we will use the new API https://github.com/virtualmin/virtualmin-pro/commit/ef769bfe06a6e9a5dffaa9144d4588789b21d20d.

iliajie:
Later on, as a logical continuation of this, we’ll be able to do more similar factoring, eventually droping all the logic from related sub-programs like `create-domain`. We can then just pass all the parsed parameters to the new sub, like this:

```perl
#!/usr/local/bin/perl

=head1 create-domain.pl

Create a virtual server
...

=cut

my $flags = &parse_cli_args();
eval { &create_domain_cli($flags); };
if ($@) {
  usage();
}
```

However, that will be a later and unrelated step, as I’m still working on PR https://github.com/virtualmin/virtualmin-gpl/pull/913 to move all flags from the existing `usage()` function into a unique list (hash) of flags for each sub-program. This will allow us to dynamically generate usage information and test flags.

jcameron:
I'm sorry but this seems like the wrong approach. It's like you took the existing `create-domain.pl` and then just turned all args into params to this new function. But that adds an unnecessary extra intermediate format that's not at all needed.

What I think would be much cleaner is an API like :

```
$overrides = {
  'owner' => 'Bob Smith',
  'parent' => $parent->{'id'},
};
create_simple_virtual_server($domain, $overrides);
```

The `create_simple_virtual_server` function would do all validation, selection of features, etc. But callers can override some domain attributes with the `$overrides` hash ref, for example to set features or choose the type of domain.

iliajie:
> I'm sorry but this seems like the wrong approach.

🙂 How can this be a wrong approach?

> It's like you took the existing `create-domain.pl` and then just turned all args into params to this new function.

Yes, exactly! The `create-domain.pl` implementation you made for domain setup is solid and already in place — no need to redo it. It’s been tested and works great. If any bugs pop up later, we’ll only need to fix this subroutine since it will cover both CLI and UI tasks. It also handles variable scoping and text abstraction, and runs in `strict` mode.

If a bug shows up in the future, this subroutine will be the only place we need to check.

> But that adds an unnecessary extra intermediate format that's not at all needed.

You probably missed my earlier comment, but this subroutine _can/will_ also be used by `create-domain.pl`.


> But callers can override some domain attributes with the $overrides hash ref, for example to set features or choose the type of domain.

Yes. I’m already doing that, but I’m not passing the domain name in a separate variable — I’ve just fixed it!


> What I think would be much cleaner is an API like :
> 
> ```
> $overrides = {
>   'owner' => 'Bob Smith',
>   'parent' => $parent->{'id'},
> };
> create_simple_virtual_server($domain, $overrides);
> ```

Yes, I have just fixed it to pass the domain name as a separate parameter, e.g.:

```
my $overrides = {
    pass => '123',
    unix => 1,
    dir => 1,
    dns => 1,
    web => 1,
    ssl => 1
};
my $dom = create_domain_cli('domain1.example.tld', $overrides);
```

It will either return an error or a reference to the newly created domain object.

The last improvement could be adjusting `create_domain_cli` so it works without any overrides at all. If no plan ID is given, it could automatically enable the default minimal necessary features like `unix`, `dir`, `web`, and `ssl`. I’ll leave that up to you if you’d like to enhance it further, making it so calling `create_domain_cli('domain1.example.tld', { pass => '1234' });` would be enough to create a new domain. Even as it stands, it’s much simpler and works perfectly for what we need.

virtual-server-lib-funcs.pl
```diff
@@ -21100,6 +21100,22 @@ sub validation_select_features
 return @fopts;
 }
 
+# load_api(lib)
+# Load given API
+sub load_api
```
jcameron, line 21105: This is nice! But please factor it out into a separate PR

lang/en
```diff
@@ -8448,4 +8448,61 @@ monitor_nodoms=No domains to check found!
 monitor_nofeats=No features to check found!
 monitor_done=Successfully checked $2 features in $1 domains
 
+cli_create_domain_missing_domain_name=Missing domain name
```
jcameron, line 8451: The whole "cli_create_domain_" prefix is too long, and also misleading since this isn't necesarily only used by CLI commands. Maybe replace it with "newdomain_" instead?

iliajie: Yes, agreed! Will do!

virtual-server-lib-funcs.pl
iliajie: But why? This change introduced the feature and is directly connected to it...

lang/en
iliajie: Fixed now. I used even shorter `api_ndom_`. Here is the patch https://github.com/virtualmin/virtualmin-gpl/pull/917/commits/4cd1e6defb951c4e57331884da1e7b2f3449bfad.

jcameron: You could even drop the `api_` from the string!

create-domain-api.pl
```diff
@@ -0,0 +1,1026 @@
...
+
+if (defined($opts->{'email'})) {
+	$email = $opts->{'email'};
+	&extract_address_parts($email) ||
+		&error($text{'api_ndom_invalid_email_address'});
+	}
+
+my $user  = defined($opts->{'user'})  ? lc($opts->{'user'})  : undef;
+my $group = defined($opts->{'group'}) ? lc($opts->{'group'}) : undef;
+my $pass;
```
jcameron, line 48: Why not just assign to `$pass` without the `if` statement?

create-domain-api.pl
```diff
@@ -0,0 +1,1026 @@
...
+	$pass = $opts->{'pass'};
+	}
+
+if (defined($opts->{'passfile'})) {
+	$pass = &read_file_contents($opts->{'passfile'});
+	$pass =~ s/\r|\n//g;
+	}
+
+my ($mysqlpass, $postgrespass, $hashpass);
+$mysqlpass    = $opts->{'mysql-pass'}    if (defined($opts->{'mysql-pass'}));
```
jcameron, line 59: Same here, no need for `if defined`

create-domain-api.pl
```diff
@@ -0,0 +1,1026 @@
...
+	$uquota = 0 if ($uquota =~ /unlimited/i);
+	$anylimits = 1;
+	}
+
+# Set features
+my %feature = ();
+foreach my $f (@features) {
+	if ($opts->{$f}) {
+		$config{$f} ||
+		    &error(&text('api_ndom_feature_not_enabled', $f));
```
jcameron, line 81: Wouldn't it be better if this API returned an error message instead of calling `error` ?

virtual-server-lib-funcs.pl
jcameron: Where is this `load_api` function called?

jcameron:
In general, you need to remove most of those un-necessary uses of `if defined($opts->{'x'})` since it's simpler to just assign to the other variable regardless.

create-domain-api.pl
iliajie: Yes, agreed! I have just fixed that!

create-domain-api.pl
iliajie: Not really, because this subroutine (i.e. `create_domain_cli`) is meant to be called in both the UI and CLI. In the CLI, we want to locally override the standard error function, which is intended for UI, by using:
```perl
{
    local *error = sub {
        my $message = shift;
        # Convert HTML to plain text
        $message = html_tags_to_text($message);
        print "Error: $message\n";
        };
    create_domain_cli('example.tld', { 'pass' => 1234, unix => 1, dir => 1, web => 1 });
}
```

virtual-server-lib-funcs.pl
iliajie: > Where is this `load_api` function called?

It’s called when we need to use this API. In particular here https://github.com/virtualmin/virtualmin-pro/blob/master/scripts/workbench-lib.pl#L1040/

jcameron: Oh .. but that's inconsistent with the rest of how Virtualmin's separate `*-lib.pl` files work. Why not just load this all the time, like we do here : https://github.com/virtualmin/virtualmin-gpl/blob/master/virtual-server-lib-funcs.pl#L19

create-domain-api.pl
jcameron: No no, that's not the right approach. API functions like this should always return either an error message, or maybe a hash ref with more details on the result.

iliajie: > No no, that's not the right approach. API functions like this should always return either an error message, or maybe a hash ref with more details on the result.

Oh, you’re absolutely right! My bad! It’s definitely better to return an error message rather than die with an HTML message in APIs. Though, on success, we return a hash for the newly created domain.

Fixed here https://github.com/virtualmin/virtualmin-gpl/pull/917/commits/ba1d4f9c0deff6c550a71be5adc10bb36bae0fa7.

virtual-server-lib-funcs.pl
iliajie: You’re right, it doesn’t match how the other lib files are loaded in Virtualmin. And, I knew that (how other libs are loaded in Virtualmin), but I made the choice on purpose. I think this API should only load when it’s needed since it’s 1000 lines long and there’s no point having it around all the time. If we did the same with other functions (like `create-user-api.pl` and others), loading them all unconditionally would just add a huge unnecessary overhead.

jcameron: Right, but the advantage of loading them in https://github.com/virtualmin/virtualmin-gpl/blob/master/virtual-server-lib-funcs.pl#L19 is that they are cached in memory after Webmin is started..

iliajie: Yes, caching is an advantage for always-needed APIs. For an API that’s only needed for specific tasks, I think lazy loading is a better approach.

lang/en
iliajie: > You could even drop the `api_` from the string!

Nah, I prefer to keep the `api_` prefix. I think it’s important, and it’s short too...

jcameron: ok fine!

virtual-server-lib-funcs.pl
jcameron: Yeah but it's inconsistent with the rest of the code. Consistency makes things easier to understand, even if it's a tiny bit less efficient.

iliajie: I completely agree about consistency, but I feel like 1000 lines of code is more than a “tiny bit.” If this is being referenced across 10 or 20 similar subroutines, would we still be seeing it as just a "small inefficiency"?

jcameron: What efficiency are you concerned about though? The RAM consumed by the code when it's loaded? The cost of parsing it is a one-off at startup time.

iliajie: Well, sure, we can do as you suggest, but for no better reason, we don’t load all the Perl modules in vendor_perl upfront — we load them on demand. So why not do it here, or conversely, why not load all the modules at once?

Is caching the only reason?

iliajie: > Oh .. but that's inconsistent with the rest of how Virtualmin's separate `*-lib.pl` files work.

Fixed here https://github.com/virtualmin/virtualmin-gpl/pull/917/commits/8f101760dc6552bd6c0042e925e4f09ae3d096e5.

jcameron: We do actually load several of the important Perl modiules up-front like POSIX and Socket, via miniserv.pl. In fact I'd be open to pre-loading more, as one of my big concerns in Webmin is page load latency due to the time spent parsing Perl..


Thanks for the fix!

iliajie: > In fact I'd be open to pre-loading more, as one of my big concerns in Webmin is page load latency due to the time spent parsing Perl

We could, but the load depends directly on the available computing power, network speed, and latency.

I remember trying to preload theme files a couple of years ago, but something didn’t work as expected.

> Thanks for the fix!

You’re welcome! Is there anything else we need to fix here before merging this PR?

jcameron: I still had a few open comments on this PR, like about the use of `if (defined($x)) {`

api-create-domain-lib.pl
```diff
@@ -0,0 +1,1006 @@
...
+	&check_ipaddress($dns_ip) ||
+		return $text{'api_ndom_invalid_dns_ip'};
+	}
+
+if ($opts->{'no-dns-ip'}) {
+	$dns_ip = "";
+	}
+
+my $mailboxlimit;
+if (defined($opts->{'mailboxlimit'}) || defined($opts->{'max-mailboxes'})) {
```
jcameron, line 178: Why two different opts that do the same thing?

api-create-domain-lib.pl
```diff
@@ -0,0 +1,1006 @@
...
+my $letsencrypt;
+if ($opts->{'letsencrypt'} || $opts->{'acme'}) {
+	$letsencrypt = 1;
+	}
+
+if ($opts->{'letsencrypt-always'} || $opts->{'acme-always'}) {
+	$letsencrypt = 2;
+	}
+
+my $jail;
```
jcameron, line 304: A ternerary expression would be more readable here

api-create-domain-lib.pl
```diff
@@ -0,0 +1,1006 @@
...
+my $noslaves;
+$noslaves       = 1 if $opts->{'no-slaves'};
+my $nosecondaries;
+$nosecondaries  = 1 if $opts->{'no-secondaries'};
+
+my $precommand = $opts->{'pre-command'};
+my $postcommand = $opts->{'post-command'};
+
+my $letsencrypt;
+if ($opts->{'letsencrypt'} || $opts->{'acme'}) {
```
jcameron, line 296: Again I don't see why two different opts are needed?

virtual-server-lib-funcs.pl
jcameron: I'm still seeing a lot of duplicated opts and use of `defined` that doesn't seem necessary. The flags to create-domain.pl shouldn't be just passed through to this new function.

iliajie: > I'm still seeing a lot of duplicated opts and use of `defined` that doesn't seem necessary.

All of those left are necessary. If you think any of them aren’t, please let me know which in particular.

api-create-domain-lib.pl
iliajie: > Why two different opts that do the same thing?

This was in the original : https://github.com/virtualmin/virtualmin-gpl/blob/master/create-domain.pl#L262

api-create-domain-lib.pl
iliajie: Alright, fixed here https://github.com/virtualmin/virtualmin-gpl/pull/917/commits/43da2202bf5da3e4204201019a07542092a37f3c.

api-create-domain-lib.pl
iliajie: > Again I don't see why two different opts are needed?

Again, because it was in the original https://github.com/virtualmin/virtualmin-gpl/commit/a6dee79f8f61731760521a6710bdc6e77537a2a9#diff-2bac7886b5474f1bb1ec134dda768ff7aec92a80564c5f2a6eb12ba1f47f3a6aR358.

jcameron: Right, but those are command line flags - because I wanted to rename the `--letsencrypt` flag to `--acme` , but wanted to keep compatibility with the old flag to not break other users of the API.  But since in this case you're creating a new API, there's no need to keep compatability.

iliajie: Ah, yes, right! I fixed it here https://github.com/virtualmin/virtualmin-gpl/pull/917/commits/c45c7eb1dc5330505e9683009d899676746a84f8.

api-create-domain-lib.pl
```diff
@@ -0,0 +1,999 @@
...
+
+my ($mysqlpass, $postgrespass, $hashpass);
+$mysqlpass    = $opts->{'mysql-pass'};
+$postgrespass = $opts->{'postgres-pass'};
+$hashpass     = 1 if ($opts->{'hashpass'});
+
+my ($quota, $uquota);
+if (defined($opts->{'quota'})) {
+	$quota = $opts->{'quota'};
+	$quota = 0 if ($quota =~ /unlimited/i);
```
jcameron, line 63: Why not just expect the caller to pass in `0` meaning unlimited? It's an internal API, so no need to support human-friendly strings like `unlimited`

api-create-domain-lib.pl
```diff
@@ -0,0 +1,999 @@
...
+	$anylimits = 1;
+	}
+
+if (defined($opts->{'uquota'})) {
+	$uquota = $opts->{'uquota'};
+	$uquota = 0 if ($uquota =~ /unlimited/i);
```
jcameron, line 69: Same here

api-create-domain-lib.pl
```diff
@@ -0,0 +1,999 @@
...
+	$ip = "allocate";    # will be done later
+	$virt = 1;
+	$name = 0;
+	}
+
+my $virtalready;
+$virtalready = 1 if ($opts->{'ip-already'});
+
+my $sharedip;
+if (defined($opts->{'shared-ip'})) {
```
jcameron, line 115: No need for `defined`

api-create-domain-lib.pl
```diff
@@ -0,0 +1,999 @@
...
+	# IPv6 on shared address
+	$ip6 = $opts->{'shared-ip6'};
+	$virt6 = 0;
+	$name6 = 1;
+	&indexof($ip6, &list_shared_ip6s()) >= 0 ||
+		return &text('api_ndom_ip6_not_in_shared_list', $ip6);
+	}
+
+my $dns_ip;
+if (defined($opts->{'dns-ip'})) {
```
jcameron, line 167: No need for `defined`

api-create-domain-lib.pl
jcameron: Right, but again this is an internal API so there's no need to support legacy flag names

api-create-domain-lib.pl
```diff
@@ -0,0 +1,999 @@
...
+	}
+
+my $mailboxlimit;
+if (defined($opts->{'mailboxlimit'}) || defined($opts->{'max-mailboxes'})) {
+	$mailboxlimit = $opts->{'mailboxlimit'} || $opts->{'max-mailboxes'};
+	$anylimits = 1;
+	}
+
+my $dbslimit;
+if (defined($opts->{'dbslimit'}) || defined($opts->{'max-dbs'})) {
```
jcameron, line 184: Same here for the multiple option names

api-create-domain-lib.pl
```diff
@@ -0,0 +1,999 @@
...
+	$dbslimit = $opts->{'dbslimit'} || $opts->{'max-dbs'};
+	$anylimits = 1;
+}
+
+my $domslimit;
+if (defined($opts->{'domslimit'}) || defined($opts->{'max-doms'})) {
```
jcameron, line 190: Same here for the multiple option names

api-create-domain-lib.pl
```diff
@@ -0,0 +1,999 @@
...
+	$domslimit = $opts->{'domslimit'} || $opts->{'max-doms'};
+	$anylimits = 1;
+	}
+
+my $aliaslimit;
+if (defined($opts->{'aliaslimit'}) || defined($opts->{'max-aliases'})) {
```
jcameron, line 196: Same here for the multiple option names

api-create-domain-lib.pl
```diff
@@ -0,0 +1,999 @@
...
+	$aliaslimit = $opts->{'aliaslimit'} || $opts->{'max-aliases'};
+	$anylimits = 1;
+	}
+
+my $aliasdomslimit;
+if (defined($opts->{'aliasdomslimit'}) || defined($opts->{'max-aliasdoms'})) {
```
jcameron, line 202: Same here for the multiple option names

api-create-domain-lib.pl
```diff
@@ -0,0 +1,999 @@
...
+	$aliasdomslimit = $opts->{'aliasdomslimit'} || $opts->{'max-aliasdoms'};
+	$anylimits = 1;
+	}
+
+my $realdomslimit;
+if (defined($opts->{'realdomslimit'}) || defined($opts->{'max-realdoms'})) {
```
jcameron, line 208: Same here for the multiple option names

api-create-domain-lib.pl
```diff
@@ -0,0 +1,999 @@
...
+	$anylimits = 1;
+	}
+
+my $tlimit;
+$tlimit = 1 if ($opts->{'limits-from-template'} || $opts->{'limits-from-plan'});
+
+my $prefix = $opts->{'prefix'} || $opts->{'suffix'};
+
+my $db;
+if (defined($opts->{'db'})) {
```
jcameron, line 249: Is `defined` really needed?

api-create-domain-lib.pl
```diff
@@ -0,0 +1,999 @@
...
+	$db = $opts->{'db'};
+	$db =~ /^[a-z0-9\-\_]+$/i || return $text{'invalid_database_name'};
+	}
+
+my $fwdto;
+if (defined($opts->{'fwdto'})) {
```
jcameron, line 255: Is `defined` really needed?

api-create-domain-lib.pl
iliajie: We could, it’s not a problem. However, I’d like this function to be a drop-in replacement for the current code in `create-domain.pl`. If I change that now, I might forget later when building the `%opts` parameter for this new subroutine. Let's just support both, `unlimited` and `0`...

api-create-domain-lib.pl
iliajie: For this and all other "Same here for the multiple option names," as I mentioned [here](https://github.com/virtualmin/virtualmin-gpl/pull/917#discussion_r1805434186), I really want this new sub to be a _simple_ drop-in replacement for the existing code in `create-domain.pl` so we don’t end up supporting both.

api-create-domain-lib.pl
jcameron: I don't think that's good API design though. Internal APIs like this don't need to support human-friendly values, since we can just do the conversion from `unlimited` to `0` in `create-domain.pl`

api-create-domain-lib.pl
jcameron: It can't be a drop-in replacement because it's an internal API, but `create-domain` is an external API. It makes no sense to keep support for legacy flags in `create-domain` in a new internal API.

iliajie: > It makes no sense to keep support for legacy flags in create-domain in a new internal API.

Agreed! Fixed on the latest commit!

jcameron:
Ok we're looking a lot better now! Only major issue I see is the support for the "unlimited" quota value, which you could just move up to the caller.

iliajie:
> Ok we're looking a lot better now! Only major issue I see is the support for the "unlimited" quota value, which you could just move up to the caller.

Alright! Fixed this too here https://github.com/virtualmin/virtualmin-gpl/pull/917/commits/cb29f69c1a6e40dd4dfe24d9b5d78d695f63b348.

Thanks!

virtual-server-lib-funcs.pl
jcameron, line 21105: I guess we can get ride of this `load_api` function then?

iliajie: Ah, yes! Sorry about that! I thought I had, but I’ve just done it now.

jcameron:
Thanks for the PR!

iliajie:
> Thanks for the PR!

Thank _you_ for reviewing it and sharing your experience!

PR #921, opened 2024-10-14
Migrate TikiWiki installer to Virtualmin GPL

iliajie:
Hello Jamie!

As I previously discussed with Marc, we’ve decided to move the TikiWiki installer to Virtualmin GPL and fully support the new Virtualmin Workbench API.

Currently, the Virtualmin TikiWiki Manager runs several commands as the root user, which we want to avoid. I’ll be adding support for most (and more) features currently available in Virtualmin Tiki Manager in the Tiki Workbench, and improving the overall Tiki installer experience.

The old Virtualmin TikiWiki module will only handle broader tasks, like managing multiple Tiki instances across different domains, where root access is needed.

There’s not much else to do here — just merge it and remove the TikiWiki installer from the Virtualmin Pro scripts updates (if that’s even needs to be done). I’ll remove the `tikiwiki.pl` file from `pro/scripts` after merging it.

Thanks!

### DIFF
--- a/scripts/tikiwiki.pl
+++ b/scripts/tikiwiki.pl
@@ -1,0 +1,312 @@
+
+# script_tikiwiki_desc()
+sub script_tikiwiki_desc
+{
+return "TikiWiki";
+}
+
+sub script_tikiwiki_uses
+{
+return ( "php" );
+}
+
+sub script_tikiwiki_longdesc
+{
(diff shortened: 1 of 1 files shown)


jcameron:
Thanks! Please send me the PR to remove it from the `pro/scripts` directory

iliajie:
Thanks, Jamie!

PR #938, opened 2024-10-21
Allow setting status collection interval to days

willnode:
Even though the manual say we can disable status collection, it is impractical since it handles more than just dashboard status.

I disable status collection on my machine since I rarely visits Virtualmin dashboard and it takes a lot of CPU. But disabling it also disables Let's Encrypt renewals, which is bad.

So it can't be disabled, but I also can't extend it more than 60 minutes.

This PR allows to extend it for up to 32 days (once a month) so Let's Encrypt renewals still works.

### DIFF
--- a/collect-lib.pl
+++ b/collect-lib.pl
@@ -714,18 +714,46 @@ sub setup_collectinfo_job
 # Work out correct steps
 local $step = $config{'collect_interval'};
 $step = 5 if (!$step || $step eq 'none');
-$step = 60 if ($step > 60);
+# max step is 32 days (46080)
+$step = 46080 if ($step > 46080);
 local $offset = int(rand()*$step);
 local @mins;
-for(my $i=$offset; $i<60; $i+= $step) {
-	push(@mins, $i);
+local @hours;
+local @days;
+
+if ($step < 120) {
+	# 1 - 60 minutes step 
+	for (my $i = $offset; $i < 60; $i += $step) {
+		push(@mins, $i);
 	}
+	@hours = ('*');
+	@days = ('*');
+} elsif ($step < 2880) {
+	# 2 - 24 hours step
+	$hstep = int($step / 60);
+	$offset = int($offset / 60);
+	for (my $i = $offset; $i < 24; $i += $hstep) {
+		push(@hours, $i);
+	}
+	@mins = ($offset % 60);
+	@days = ('*');
+} else {
+	# 2 - 31 days step
+	$dstep = int($step / 1440);
+	$offset = int($offset / 1440);
+	for (my $i = $offset; $i < 32; $i += $dstep) {
+		push(@days, $i);
+	}
+	@mins = ($offset % 60);
+	@hours = (int($offset / 60) % 60);
+}
+
 local $job = &find_cron_script($collect_cron_cmd);
 if (!$job && $config{'collect_interval'} ne 'none') {
 	# Create, and run for the first time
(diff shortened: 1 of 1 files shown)


iliajie:
Hello,

Thanks for your PR!

> Virtualmin dashboard and it takes a lot of CPU.

All the data on the Virtualmin dashboard is cached, so there’s no need to disable status collection. It's not the right approach. Sorry.

> So it can't be disabled, but I also can't extend it more than 60 minutes.

That’s rather a bit too complicated. I also don’t think it’s right to stop Let’s Encrypt renewals when status collection is off. We need to separate the logic for that. What do you think, Jamie?

jcameron:
Seems like we need to look at the cause of CPU consumption if running this only once an hour is too expensive! How long does it run for on your system?

willnode:
With host about 350 sites it's 40 seconds first time, then subsequent run is about 20 seconds.

willnode:
Well I have slower system, it runs for 2 min 40 seconds first time, then 1 min 20 seconds subsequently

willnode:
I'm suspect that most lengthy operation about this cron job is checking SSL renew. It's much expensive than you think.

I have an occurrence where it retries many domains at a time and kept failing due to customer has broken SSL setup somewhere along the way. Image this running every hour!

Even i have to code [my own scheduler](https://github.com/domcloud/bridge/blob/main/sudocleanssl.js) to disable renew if domains kept failing to renew for >24 hours.

jcameron:
Wow that is pretty slow! Is it actually re-trying to renew the Let's Encrypt on every attempt?

willnode:
I read some code here last time, it seems has some back-off mechanism to prevent hitting rate limit. 

Nonetheless, yes it keep retrying. I had couple occurences when the whole server is hitting LE rate limit in IP level until I code my own scheduler and SSL domain sharing.

jcameron:
Yeah the rate limit is per domain, so if you had a large number of domains that are behind on their LE renewals it could cause problems. Does the LE service have a per-IP limit as well as per-domain?

willnode:
Yes, 300 New Orders per account (basically it's whole server) per 3 hours. https://letsencrypt.org/docs/rate-limits/

willnode:
I offer a kind of free hosting, if you wonder why.

jcameron:
Odd, there's already code in Virtualmin which attempts to stay below 150 renewals per hour for this reason.

willnode:
The limit inside `apply_letsencrypt_cert_renewals`? I think it is because it only count renewals inside scheduler, not counting new SSL issues and manually triggered renewals.

Anyway, this IP level limit is only happened rarely and it's been long time ago; i'm happy with my current setup.

I could better execute this PR by splitting the collectinfo.pl cronjob but that's a lot of changes.

jcameron:
One thing you could try to lower the load from LE renewals is to reduce the number here : https://github.com/virtualmin/virtualmin-gpl/blob/master/feature-ssl.pl#L2917

Say down to 50 from the default of 300.

willnode:
I think that will do. Thanks!

How about this PR? No longer valid? Or reduce to hourly cap?

jcameron:
Let me know how it goes for you with a lower cap, and if that solves the issue maybe we can drop this PR.

willnode:
Hello, after few days I think I can safely say i'm quite comfortable with 60 minutes interval. I think we can drop this.

Please separate the cronjob for status collection and SSL renewal if it possible, thank you!

PR #947, opened 2024-10-30
Fix the incorrect temporary default redirect code from 301 to 302

ReallyVirtual:
Redirect defaults to temporary redirect [R] with the code 302. 301 is a permanent redirect.


iliajie:
Thanks for the patch!

PR #950, opened 2024-11-01
Update feature-dir.pl to add SELinux setting for /tmp directory

verne-work:
If the SELinux utility chcon is available, its likely SELinux is enabled on the system.  If so, set the tmp directory to the specific security context of  "tmp_t".

### DIFF
--- a/feature-dir.pl
+++ b/feature-dir.pl
@@ -1047,7 +1047,12 @@ if ($d->{'dir'}) {
 	local $tmp = "$d->{'home'}/tmp";
 	if (!-d $tmp) {
 		&make_dir_as_domain_user($d, $tmp, 0750, 1);
-		}
+		if (&has_command("chcon")) {
+			&execute_command("chcon -t tmp_t ".
+				quotemeta("$tmp").
+				">/dev/null 2>&1");
+			}
+  }
 	return $tmp;
 	}
 else {


jcameron:
Thanks for the patch!

feature-dir.pl
```diff
@@ -1047,7 +1047,12 @@ sub create_server_tmp
 	local $tmp = "$d->{'home'}/tmp";
 	if (!-d $tmp) {
 		&make_dir_as_domain_user($d, $tmp, 0750, 1);
-		}
+		if (&has_command("chcon")) {
```
iliajie, line 1050: @verne-work Do we need to execute this if SELinux is disabled? Also, using `chcon` won’t persist through reboots or relabeling — shouldn't you be using `semanage` instead?

@jcameron I don’t think this patch is complete enough yet to be merged.

verne-work: This is my initial stab at submitting a number of SELinux related local custom patches I have.  I did look for existing code in Virtualmin that uses chcon and just copied its format and general syntax  😄    But it does make sense if SELinux is disabled, that no files at all will have any context settings so this also would not be needed.

I believe as the change is to the metadata on disk for the file, it does survive a reboot.

But a restorecon wants to change it ("restore it") to user_home_t which frankly I did not know was even there, so I have not tested that setting.  In fact I have been using my custom patch for this adjustment for awhile now and offhand do not remember exactly what operations/functions made SELinux mad when it was set to home_root_t.

I did toy with the semanage concept (I use it for some other settings), but was not sure if Virtualmin wanted to get into any of that; does Virtualmin have any core or addon modules to manage SELinux settings/items like semanage entries?  Also I worry about edge cases, if any even exist, where a semanage custom setting is somehow ignored.

iliajie: The problem is that it won’t be re-labeled with the correct SELinux type after a reboot.

> I did toy with the semanage concept (I use it for some other settings), but was not sure if Virtualmin wanted to get into any of that;

I think if we do this, we should do it right. Imagine the confusion if it works initially after domain was created, but then your PHP app starts failing magically after a reboot, because SELinux will prevent writing to the `~/tmp` dir.

And, I'm sorry, as I had to revert this PR, I suggest you run more tests and then create a new PR so we can discuss it.

Thank you very much for your time and effort!

verne-work: while I disagree with the reboot comment, I will update/enhance my code and do more tests, and research to substantiate everything.

and a big THANK YOU for even considering anything related to SELinux !!

iliajie: > while I disagree with the reboot comment

Why do you disagree though? `if (!-d $tmp) { }` runs only once.

> and a big THANK YOU for even considering anything related to SELinux !!

I’ve used SELinux for years on one of my production servers. It’s a real PITA

verne-work: the few Redhat 8 systems I'm running keep that tmp_t setting across reboots, from what I am assuming is because the security context is part of the file metadata ON DISK.

if a relabeling operation is done, of course that would change it -- do some environments to a global relabeling at every reboot?

or I am not understanding at all what you are telling me, SORRY  😦 

and YES its a PITA -- but my boss thinks its really important -- what's a guy to do??

PR #953, opened 2024-11-02
Fix warnings and messages

iliajie:
Hello Jamie,

This is a related PR that we should discuss here: https://github.com/virtualmin/virtualmin-pro/pull/47.


virtual-server-lib-funcs.pl
```diff
@@ -4132,26 +4132,58 @@ sub bottom_scroll_js
...
 # Print functions for HTML output
-sub first_html_print { print_and_capture("<span data-first-print>@_</span>","<br>\n");
-		       print bottom_scroll_js(); }
-sub second_html_print { print_and_capture("<span data-second-print>@_</span>","<br><br data-x-br>\n");
-		        print bottom_scroll_js(); }
-sub indent_html_print { print_and_capture("<ul data-indent-print>\n");
-		        print bottom_scroll_js(); }
-sub outdent_html_print { print_and_capture("</ul>\n");
-		         print bottom_scroll_js(); }
+sub first_html_print
```
jcameron, line 4135: What is going on here??

virtualmin-licence.pl
```diff
@@ -25,17 +25,177 @@ sub licence_scheduled
...
-	(3, $1, "The licence for this server expired on $1", $2, $3, $4) :
-       $out =~ /^ERR\s+(.*)/ ?
-	(2, undef, "An error occurred checking the licence : $1", undef) :
-       $out =~ /^OK\s+(\S+)\s+(\S+)\s+(\S+)\s+(\S+)\s+(\d+)/ ?
-	(0, $1, undef, $2, $3, $4, $5) :	# Auto-renewal flag
-       $out =~ /^OK\s+(\S+)\s+(\S+)\s+(\S+)\s+(\S+)/ ?
-	(0, $1, undef, $2, $3, $4) :
-	(1, undef, "No valid licence was found for your host ID $hostid and ".
-		   "serial number $serial{'SerialNumber'}", undef);
+return (2, undef, "$text{'licence_efailed'} : $error") if ($error);
```
jcameron, line 28: Are we expecting a change in the format returned by the licence manager?

virtualmin-licence.pl
```diff
@@ -25,17 +25,177 @@ sub licence_scheduled
...
+       $out =~ /^ERR\s+(?<err>.*)/ && ($regerr = $+{err}) &&
+       	       $regerr !~ /invalid\s+host\s+or\s+serial\s+number/i ?
+	(2, undef, "$text{'licence_echk'} : $regerr", undef) :
+       $out =~ /^OK\s+(\S+)\s+(\S+)\s+(\S+)\s+(\S+)(?:\s+(\d+)(?:\s+([\w])(?:\s+(\d+))?)?)?/ ?
+	(0, $1, undef, $2, $3, $4, $5, $6, $7) :
+	(1, undef, &text("licence_evalid",
+		   	 "<tt data-evalid>$serial{'SerialNumber'}</tt>"));
+}
+
+# Change license with a new serial and key
```
jcameron, line 40: Why is this code even in the GPL version?

virtual-server-lib.pl
```diff
@@ -463,5 +466,37 @@ sub help_pre_load
 return $htext;
 }
 
+# check_first_print_lock([html])
+# Returns a warning message if set, but only the first time it prints
+our $check_first_print_lock = sub
+{
+my $html = shift;
+# Read-only mode warning
+if ($main::readonly_mode_cache) {
```
jcameron, line 475: I'm not sure I like this check being done in `first_print`, it seems like the wrong place for it

virtual-server-lib-funcs.pl
iliajie: See my explanations [here](https://github.com/virtualmin/virtualmin-pro/pull/47#issuecomment-2453419239).

virtualmin-licence.pl
iliajie: > Why is this code even in the GPL version?

Who knows why `virtualmin-licence.pl` and `change-licence.pl` ended up in a GPL repo? But hey, let’s save that chat for another day!

virtual-server-lib.pl
iliajie: Also, see my explanations [here](https://github.com/virtualmin/virtualmin-pro/pull/47#issuecomment-2453419239).

virtualmin-licence.pl
jcameron: Ok in that case we'll have to roll this out carefully, because we cannot assume that all users will upgrade on any reasonable schedule, so we can't make breaking changes on the licence manager.

virtualmin-licence.pl
jcameron: Ok we can leave it here then

virtualmin-licence.pl
iliajie: There are no and will not be any breaking changes...

virtualmin-licence.pl
iliajie: Sure

virtualmin-licence.pl
jcameron: Even if an older Virtualmin gets the new-format message from the licence manager?

iliajie: > Even if an older Virtualmin gets the new-format message from the licence manager?

Yes, that’s right! In fact, they’re already getting it. The original regex didn’t have a `$` at the end, so there’s nothing to worry about.

jcameron: Excellent!

licence-info.pl
```diff
@@ -44,6 +45,7 @@ package virtual_server;
 &read_file($licence_status, \%lstatus);
 print "Serial number: $vserial{'SerialNumber'}\n";
 print "License key: $vserial{'LicenseKey'}\n";
+print "Host ID: @{[&get_licence_hostid()]}\n";
```
jcameron, line 48: Why not just : 

`print "Host ID: ",&get_licence_hostid(),"\n";
`

licence.cgi
```diff
@@ -1,27 +1,27 @@
 #!/usr/local/bin/perl
 # Force a re-check of the licence
-
+$main::readonly_mode = 0;
 require './virtual-server-lib.pl';
 &can_recheck_licence() || &error($text{'licence_ecannot'});
-&ui_print_unbuffered_header(undef, $text{'licence_title'}, "");
+&read_env_file($virtualmin_license_file, \%serial);
+my $serial = $serial{'SerialNumber'};
+($serial =~ /$virtualmin_gpl/i || $serial eq '1234567') &&
```
jcameron, line 8: Another use of this magic serial number

licence.cgi
```diff
@@ -33,29 +33,22 @@ if ($status == 0) {
...
-		print &text('licence_goterr',
-			&ui_text_color($err, 'danger')),"<p>\n";
-		}
+	$err = lcfirst($err);
+	$err =~ s/\s*\.$//;
+	print &text('licence_goterr', lcfirst($err)),"<p>\n";
 	}
 
-&ui_print_footer("", $text{'index_return'});
+&ui_print_footer("pro/licence.cgi", lc($text{'licence_manager'}));
```
jcameron, line 53: Why the pro path when this `licence.cgi` file is in the GPL directory?

virtual-server-lib-funcs.pl
```diff
@@ -11838,6 +11870,24 @@ sub require_licence
...
+{
+return 0 if (!$virtualmin_pro);
+return 0 if (!-r $licence_status);
+my %licence_status;
+&read_file($licence_status, \%licence_status);
+my ($bind, $time) = ($licence_status{'bind'}, $licence_status{'time'});
+my $scale = 1;
+$scale = 3 if ($licence_status{'status'} == 3);
+if ($bind && !$time) {
+	$bind = int(($bind-time())/86400)+$virtualmin_pro/$scale;
```
jcameron, line 11885: Why is $virtualmin_pro being used in a mathematically expression here??

virtual-server-lib-funcs.pl
```diff
@@ -12034,19 +12092,40 @@ sub list_warning_messages
...
-		if ($alert_text !~ /$virtualmin_renewal_url/);
-	$alert_text =~ s/>\Q$virtualmin_renewal_url\E</>$text{'license_shop_name'}</;
+	$alert_text .= $err;
+	$alert_text = "$alert_text. " if ($err !~ /\.$/);
+	if ($bind) {
+		my $multi = $bind > 1 ? 'm' : '';
+		$alert_text .= " $text{'licence_maxwarn'}";
+		$alert_text .= " ".&text("licence_renew3$multi", $bind); 
+		}
+	elsif ($alert_text !~ /$virtualmin_renewal_url/) {
```
jcameron, line 12113: Seems very odd to be checking if the alert message already contains the URL when we're actually building up the alert message in this same function?

virtual-server-lib-funcs.pl
```diff
@@ -12034,19 +12092,40 @@ sub list_warning_messages
...
+					 $virtualmin_renewal_url,
+					 $text{'license_shop_name'})
+		}
+	else {
+		$alert_text .= " $text{'licence_renew4'}" if (defined($bind));
+		}
 	if (&can_recheck_licence()) {
-		$alert_text .= &ui_form_start("@{[&get_webprefix_safe()]}/$module_name/licence.cgi");
-		$alert_text .= &ui_submit($text{'licence_recheck'});
+		$alert_text .= &ui_form_start("@{[&get_webprefix_safe()]}/$module_name/pro/licence.cgi");
```
jcameron, line 12124: Change this to : 

`&get_webprefix_safe() . "/$module_name/pro/licence.cgi"
`

virtual-server-lib.pl
```diff
@@ -66,8 +66,9 @@
 	}
 
 # For the GPL version, force some features off.
+$virtualmin_gpl = 'GPL';
 $virtualmin_pro = $module_info{'version'} =~ /pro/i ||
-		  -d "$module_root_directory/pro" ? 1 : 0;
+		  -d "$module_root_directory/pro" ? 20 + 1 : 0;
```
jcameron, line 71: What is this number 20 here?

virtualmin-licence.pl
```diff
@@ -25,17 +25,177 @@ sub licence_scheduled
...
+sub change_licence
```
jcameron, line 41: Is this code exactly the same as what was extracted from change-licence.pl ? Knowing that will make it easier for me to review

licence-info.pl
iliajie: Sure, why not. Fixed now.

licence.cgi
iliajie: Duh, thanks! Fixed now!

licence.cgi
iliajie: > Why the pro path when this `licence.cgi` file is in the GPL directory?

No, this is correctly referenced `licence.cgi` file in `pro` directory.

virtual-server-lib-funcs.pl
iliajie: > Why is $virtualmin_pro being used in a mathematically expression here??

I explained this here https://github.com/virtualmin/virtualmin-pro/pull/47#issuecomment-2453611662.

virtual-server-lib-funcs.pl
iliajie: Yes, it is. This is the old code though. Not part of this PR. There was a reason why I added it in the past. Not sure what it was now. So, I just kept it because it's not really relevant...

virtual-server-lib-funcs.pl
iliajie: Okay, fixed!

virtual-server-lib.pl
iliajie: > What is this number 20 here?

Explained here https://github.com/virtualmin/virtualmin-pro/pull/47#issuecomment-2453611662.

virtualmin-licence.pl
iliajie: Yes, it is. Though, it has now only the language strings loaded from `lang/en`...

jcameron:
Let me know when this is ready for re-review based on our VC discussion today ..

virtual-server-lib-funcs.pl
```diff
@@ -5041,6 +5060,7 @@ sub reset_domain_envs
 # pre-change command
 sub making_changes
 {
+return &pre_making_changes() if (defined(&pre_making_changes));
```
jcameron, line 5063: This will cause the pre-command to never be run!

virtual-server-lib-funcs.pl
```diff
@@ -11838,6 +11858,44 @@ sub require_licence
...
+&read_file($licence_status, \%licence_status);
+my ($bind, $time) = ($licence_status{'bind'}, $licence_status{'time'});
+my $scale = 1;
+$scale = 3 if ($licence_status{'status'} == 3);
+if ($main::webmin_script_type ne 'cron' && !$time && $bind &&
+    int(($bind-time())/86400)+($virtualmin_pro/$scale) <= 0) {
+	my $title = $text{'readonly_mode_warning'};
+	my $body = &text("license_manager_readonly", &get_webprefix_safe().
+				"/$module_name/pro/licence.cgi");
+	*pre_making_changes = sub {
```
jcameron, line 11877: What on earth is this * format?

virtual-server-lib.pl
jcameron: I think use a different variable for that, it's confusing to change this simple boolean into a number of days

virtual-server-lib-funcs.pl
jcameron: I'd recommend instead just always defining it, and checking if it has returned anything and if so return that from this function. That's a lot cleaner than conditionally defining it with that `*pre_making_changes` code.

virtual-server-lib-funcs.pl
jcameron: Oh I see what you're doing here. But please do it differently rather than dynamically creating a sub!

virtual-server-lib.pl
```diff
@@ -369,6 +371,32 @@
...
 $acme_providers_dir = "$module_config_directory/acme";
 
+# $making_changes->(module)
+# Injects code into the original making_changes subroutine of a module
+$making_changes = sub
+{
+my ($module) = @_;
+my $module_making_changes = "${module}::making_changes";
+# Backup the original subroutine
+*{"${module}_making_changes"} = \&{$module_making_changes};
```
jcameron, line 381: Wow this is really complex! Why not just change what `making_changes` sub does? And don't try to override the sub in the Users and Groups module, we can do that a much cleaner way.

virtual-server-lib-funcs.pl
iliajie: > I'd recommend instead just always defining it, and checking if it has returned anything and if so return that from this function. That's a lot cleaner than conditionally defining it with that `*pre_making_changes` code.

I’m not sure I see the benefits of this. Defining it all the time would just make the code more complex, and I don’t think I really like that idea...

virtual-server-lib-funcs.pl
iliajie: > Oh I see what you're doing here. But please do it differently rather than dynamically creating a sub!

See [this](https://github.com/virtualmin/virtualmin-gpl/pull/953#discussion_r1831276335). I don’t really understand why not. We check if the sub is defined and then call it if it is, multiple times throughout Virtualmin...

virtual-server-lib.pl
iliajie: > Wow this is really complex! Why not just change what making_changes sub does? 

`making_changes` is specific to a module. We shouldn't change it, as we only want to add a pre-hook.

> And don't try to override the sub in the Users and Groups module, we can do that a much cleaner way.

Why is this not good? We just extend it with the code we want, without actually changing the original function.

Anyway, I have changed it to a level where it can be used with any module's function. Please have another look https://github.com/virtualmin/virtualmin-gpl/pull/953/commits/a289549e7f4141b92e53711d8864d7511936b1dd.

virtual-server-lib-funcs.pl
jcameron: Yes but we do that only when it may or may not exist depending on Pro vs GPL. In this case it seems like the function is being defined only when the licence has expired

virtual-server-lib.pl
iliajie: Yeah, right, will do it differently in the next final commit...

virtual-server-lib-funcs.pl
```diff
@@ -5057,6 +5057,43 @@ sub making_changes
...
+# Licence related checks
+if ($virtualmin_pro && -r $licence_status) {
+	my %licence_status;
+	&read_file($licence_status, \%licence_status);
+	my ($bind, $time) = ($licence_status{'bind'}, $licence_status{'time'});
+	my $scale = 1;
+	$scale = 3 if ($licence_status{'status'} == 3);
+	if ($main::webmin_script_type ne 'cron' && !$time && $bind &&
+	    int(($bind-time())/86400)+(sqrt(441)/$scale) <= 0) {
+		my $title = $text{'licence'.'_'.'expired'};
```
jcameron, line 5073: Why not just make this the string 'licence_expired' ?

virtual-server-lib-funcs.pl
```diff
@@ -12089,7 +12088,7 @@ sub list_warning_messages
 	my $alert_text;
 	# Not valid .. show message
 	if ($bind) {
-		my $prd = $virtualmin_pro/$scale;
+		my $prd = sqrt(441)/$scale;
```
jcameron, line 12091: What on earth is the square root of 441 doing here?

virtual-server-lib-funcs.pl
iliajie: For apparent reason, not to make it easily searchable.

virtual-server-lib-funcs.pl
```diff
@@ -5057,6 +5057,43 @@ sub making_changes
...
+if ($virtualmin_pro && -r $licence_status) {
+	my %licence_status;
+	&read_file($licence_status, \%licence_status);
+	my ($bind, $time) = ($licence_status{'bind'}, $licence_status{'time'});
+	my $scale = 1;
+	$scale = 3 if ($licence_status{'status'} == 3);
+	if ($main::webmin_script_type ne 'cron' && !$time && $bind &&
+	    int(($bind-time())/86400)+(sqrt(441)/$scale) <= 0) {
+		my $title = $text{'licence'.'_'.'expired'};
+		my $body = &text('licence'.'_'.'expired'.'_'.'desc',
```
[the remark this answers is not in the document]
iliajie: For the same reason.

virtual-server-lib-funcs.pl
iliajie: It's number 21, not very apparent for the same reason. It's instead of using it with `$virtualmin_pro` variable. Also will make it for us to conveniently search and replace.

iliajie:
Jamie, let's also discuss this https://github.com/virtualmin/virtualmin-gpl/pull/953/commits/a802def57571a12a447f83e543e63b79dba7fe7d change as potentially acceptable.

I find it simpler and more straightforward because we just exit cleanly without potentially running anything. It also requires no wrapper for `useradmin`, as we discussed earlier.

iliajie:
I just realized that we already test modifier pages [here](https://github.com/virtualmin/virtualmin-gpl/blob/dev/liccheck/cgi_args.pl#L55).

virtual-server-lib-funcs.pl
jcameron: Let's not try to be clever about preventing people from changing the code. The reality is that this is impossible to prevent, and we're only going to make our own code harder to understand for no benefit.

create-admin.pl
```diff
@@ -45,6 +45,7 @@ package virtual_server;
 	$0 = "$pwd/create-admin.pl";
 	require './virtual-server-lib.pl';
 	$< == 0 || die "create-admin.pl must be run as root";
+	&licence_status();
```
jcameron, line 48: In this case and all the others, put this `licence_status` call just this `if` block

iliajie: Fixed https://github.com/virtualmin/virtualmin-gpl/pull/953/commits/5f05f12ae8c9bd82878476c8e33554b4cad49c16.

virtual-server-lib-funcs.pl
iliajie: Got it! Fixed now https://github.com/virtualmin/virtualmin-gpl/pull/953/commits/c05b8995c6a059031388856ed23a772e938cee68.

virtual-server-lib-funcs.pl
```diff
@@ -11912,7 +11911,11 @@ sub licence_status
 			exit(99);
 			}
 		elsif ($main::webmin_script_type eq 'web') {
-			&error("$titlecgi : $body");
+			&header($text{'error'}, "");
```
jcameron, line 11914: Why not ui_print_header ?

virtual-server-lib-funcs.pl
```diff
@@ -11878,6 +11880,48 @@ sub setup_licence_cron
...
 
+# licence_status()
+# Checks license status
+sub licence_status
+{
+# Licence related checks
+if ($virtualmin_pro && -r $licence_status) {
+	my %licence_status;
+	&read_file($licence_status, \%licence_status);
+	my ($bind, $time) = ($licence_status{'bind'}, $licence_status{'time'});
```
jcameron, line 11891: What is the meaning of this `bind` field? The name doesn't make it clear

virtual-server-lib-funcs.pl
jcameron: If the number 21 is repeated in a lot of place, maybe just put it in a global variable instead. I assume this is the number of days in the licence grace period?

virtual-server-lib-funcs.pl
jcameron: I see this has changed per our discussion

virtual-server-lib-funcs.pl
iliajie: > Why not ui_print_header ?

It's what we use in `error();`. I think it's just fine.

virtual-server-lib-funcs.pl
iliajie: > What is the meaning of this `bind` field? The name doesn't make it clear

https://github.com/virtualmin/virtualmin-pro/pull/47#issuecomment-2462982848

virtual-server-lib-funcs.pl
iliajie: I already fixed it by just using a number.

virtual-server-lib-funcs.pl
jcameron: That's only because there are some super-rare cases where the theme isn't loaded. In this case you can use ui_print_header

virtual-server-lib-funcs.pl
jcameron: got it thanks!

virtual-server-lib-funcs.pl
iliajie: > That's only because there are some super-rare cases where the theme isn't loaded. In this case you can use ui_print_header

Alright, fixed!

virtual-server-lib.pl
jcameron: Resolving this as it's no longer needed

PR #955, opened 2024-11-07
Use static value for pm.max_spare_servers

### LINKED ISSUE #954: `max_spare_servers` set too high dynamically. Running out of memory.
Ref: https://forum.virtualmin.com/t/max-spare-servers-set-too-high-dynamically/130261

Our vhosts are filling the memory on the server with old php-fpm processes, we believe this is due to the dynamic `max_spare_servers` setting.

With this set to 8 (as it seems to be dynamically with our setup of `max_children` = 16), we are seeing very large consistent memory usage (~50GB) across as relatively small set of vhosts (~60).

Example pool file:

```
[164448877221794]
user = test
group = test
listen.owner = test
listen.group = test
listen.mode = 0660
listen = 127.0.0.1:8064
pm = dynamic
pm.max_children = 16
pm.start_servers = 2
pm.min_spare_servers = 1
pm.max_spare_servers = 8 ; <- Far too high
php_value[upload_tmp_dir] = /home/test/tmp
php_value[session.save_path] = /home/test/tmp
php_value[log_errors] = On
php_admin_value[upload_max_filesize] = 40M
php_admin_value[post_max_size] = 40M
```

These 8x child processes per vhost really add up and in most situations do not need to be there. Unless you are expecting very high levels of traffic you never really need 8x processes on standby, 1 or 2 is nearly always sufficient and more will be spun up as needed.

From the PHP-FPM doc comments:

```
;             pm.max_spare_servers - the maximum number of children in 'idle'
;                                    state (waiting to process). If the number
;                                    of 'idle' processes is greater than this
;                                    number then some children will be killed
```

`pm.max_spare_servers` is the number of 'standby' processes that it will leave idle for each pool. If you have 8 of these using ~100mb each for every pool file then it quickly causes a memory issue.

In my opinion it should not be tied to `pm.max_children` at all, it would be far better suited to a static value, many guides advise setting this to the same as `pm.start_servers`.

P.S. One option is to reduce the `pm.max_children` setting for every vhost, however we've found that 16 is the sweet spot (for us) to allow for traffic bursts while protecting the server from too much load overall.

Muffinman:
Setting pm.max_children to a high value causes a lot of excess memory usage as vhost count grows. Use a static value to stop that.

Fixes #954

### DIFF
--- a/php-lib.pl
+++ b/php-lib.pl
@@ -2839,8 +2839,7 @@ return int($max);
 sub get_php_max_spare_servers
 {
 my ($defchildren) = @_;
-my $defmaxspare = $defchildren <= 1 ? $defchildren :
-        $defchildren >= 4 ? int($defchildren / 2) : 2;
+my $defmaxspare = $defchildren <= 1 ? $defchildren : 2;
 return int($defmaxspare);
 }
 


iliajie:
Thanks! 

See this comment https://github.com/virtualmin/virtualmin-gpl/issues/954#issuecomment-2461896279.

PR #974, opened 2024-12-02
Add basic support for IPv6-only systems

iliajie:
G'day Jamie!

This PR introduces basic support for IPv6-only systems, making sure it continues to work smoothly without failing when IPv4 is _not_ available.

There are a few more complicated patches that needs to be done in `add_name_virtual` and/or `add_listen` sub to make sure `VirtualHost` record isn't created as:

```
<VirtualHost :80 [fdb2:2c26:f4e4:0:21c:2222:1111:0000]:80>
```

To be clear, `/cgi-bin/ip.cgi` already returns IPv6 address correctly.

Please have a look!

### DIFF
--- a/dynip-lib.pl
+++ b/dynip-lib.pl
@@ -63,10 +63,14 @@ my $ip = sub {
 	return undef;
 	};
 my $now = time();
-if (!$nocache && $config{'external_ip_cache'} &&
-    $now - $config{'external_ip_cache_time'} < 24*60*60) {
+my $cache_optname = $type == 4 ?
+	'external_ip_cache' : 'external_ipv6_cache';
+my $cache_time_optname = $type == 4 ?
+	'external_ip_cache_time' : 'external_ipv6_cache_time';
+if (!$nocache && $config{$cache_optname} &&
+    $now - $config{$cache_time_optname} < 24*60*60) {
 	# Can use last cached value
-	return $ip->($config{'external_ip_cache'});
+	return $ip->($config{$cache_optname});
 	}
 my $url = "http://software.virtualmin.com/cgi-bin/ip.cgi";
 my ($host, $port, $page, $ssl) = &parse_http_url($url);
@@ -77,12 +81,12 @@ $out =~ s/\r|\n//g;
 $out = $ip->($out);
 if ($error) {
 	# Fall back to last cached value
-	return $ip->($config{'external_ip_cache'});
+	return $ip->($config{$cache_optname});
 	}
 # Cache it for future calls
 &lock_file($module_config_file);
-$config{'external_ip_cache'} = $out;
-$config{'external_ip_cache_time'} = $now;
+$config{$cache_optname} = $out;
+$config{$cache_time_optname} = $now;
 &save_module_config();
 &unlock_file($module_config_file);
 return $out;
--- a/dynip-lib.pl
+++ b/dynip-lib.pl
@@ -70,8 +70,7 @@ my $ip = sub {
 my $now = time();
 my $cache_optname = $type == 4 ?
 	'external_ip_cache' : 'external_ipv6_cache';
-my $cache_time_optname = $type == 4 ?
-	'external_ip_cache_time' : 'external_ipv6_cache_time';
+my $cache_time_optname = $cache_optname.'_time';
 if (!$nocache && $config{$cache_optname} &&
     $now - $config{$cache_time_optname} < 24*60*60) {
 	# Can use last cached value
--- a/dynip-lib.pl
+++ b/dynip-lib.pl
@@ -92,6 +92,17 @@ $config{$cache_time_optname} = $now;
 return $out;
 }
 
+# get_any_external_ip_address([no-cache], [prefer-ip-type])
+# Returns the IP address of this system, as seen by other hosts on the Internet,
+# either IPv4 or IPv6, preferring IPv4 by default.
+sub get_any_external_ip_address
+{
+my ($nocache, $prefer) = shift;
+my $ip4 = &get_external_ip_address($nocache, 4) if (!$prefer || $prefer != 6);
+my $ip6 = &get_external_ip_address($nocache, 6) if (!$prefer || $prefer != 4);
+return $prefer == 4 ? $ip4 : $ip6 || $ip4;
+}
+
 # update_dynip_service(new-ip, old-ip)
 # Talk to the configured dynamic DNS service, and return the set IP address
 # and an error message (if any)
--- a/dynip-lib.pl
+++ b/dynip-lib.pl
@@ -49,17 +49,22 @@ else {
 # Returns the IP address of this system, as seen by other hosts on the Internet.
 # If the no-cache flag is set, the IP is always fetched from the network. If the
 # type is set to 6 ("ipv6"), an IPv6 address is returned. Default is 4 (IPv4).
+# For DNS resolver to work, the config option "dns_resolver" must be set to a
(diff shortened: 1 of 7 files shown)


virtual-server-lib-funcs.pl
```diff
@@ -16574,7 +16583,8 @@ sub check_virtual_server_config
 else {
 	my $ext_ip = &get_external_ip_address(1);
 	if ($ext_ip) {
-		&$second_print(&text('check_dnsip3', $ext_ip));
+		my $ipv6 = $ext_ip =~ /:/ ? "v6" : "";
```
jcameron, line 16586: Cleaner just to write : 

```
my $msg = $ext_ip =~ /:/ ? 'check_dnsip3v6' : 'check_dnsip3';
&$second_print(&text($msg, $ext_ip));
```

virtual-server-lib-funcs.pl
```diff
@@ -16574,7 +16583,8 @@ sub check_virtual_server_config
 else {
 	my $ext_ip = &get_external_ip_address(1);
```
jcameron, line 16584: Does this return the external IPv4 or v4 address? Seems like a system might have both..

iliajie: > Does this return the external IPv4 or v4 address?

No, not really. It can also return an IPv6 address, as `software.virtualmin.com` now supports the IPv6 stack. If the remote system doesn’t have IPv4 configured, our server will fall back to displaying the user’s public IPv6 address.

jcameron: Maybe we need separate functions to get the external IPv4 and v6 addresses then? Or at least be clear which address type this function will prefer if both exist.

iliajie: > Maybe we need separate functions to get the external IPv4 and v6 addresses then? Or at least be clear which address type this function will prefer if both exist.

Yeah, I was also thinking about the possible consequences of `ip.cgi` not returning different types of IP addresses.

Though, do we really care which type of IP address is returned? We wouldn’t add IPv6 address to A records, right?

jcameron: Yes we do care because `get_external_ip_address` is used in a few different places, and existing code likely assumes it's always IPv4. To preserve existing behavior it should return `undef` if no external IPv4 address can be determined (like on a v6-only system), and there should be a separate `get_external_ip6_address` function.

iliajie: Jamie, that makes perfect sense!

Do you want this to be part of this PR, or would you prefer to merge this one and handle it separately?

jcameron: I think it would make sense as part of this PR.

iliajie: > Maybe we need separate functions to get the external IPv4 and v6 addresses then? Or at least be clear which address type this function will prefer if both exist.

Implemented here https://github.com/virtualmin/virtualmin-gpl/pull/974/commits/d686fef6cd6b3e13fc4ad2d208d2388caa0fb666

iliajie: I've checked in additional fixes. It is not possible to distinctly get IP address of type ore return undef, e.g.:

```perl
# Get IPv4, or return empty if not available
my $ip = &get_external_ip_address();

# Cached
my $ip4 = &get_external_ip_address(undef, 4); # get IPv4 or empty
my $ip6 = &get_external_ip_address(undef, 6); # get IPv6 or empty

# No cache
my $ip4 = &get_external_ip_address(1, 4); # re-check and get IPv4 or empty
my $ip6 = &get_external_ip_address(1, 6); # re-check and get IPv4 or empty
```

Or get _any_ available address using:

```perl
# Cached, prefer IPv4 by default or return IPv6 otherwise
my $ip = &get_any_external_ip_address();

# No cache, prefer IPv6, otherwise return IPv4
my $ip = &get_any_external_ip_address(1, 6);
```

The only things left to be fixed is Apache not creating broken like `<VirtualHost :80 [fdb2:2c26:f4e4:0::]:80>` record and DNS like `domain.tld.      IN      A       fdb2:2c26:f4e4:0::` — could you have a look a this?

To be clear, API is working just fine. But, the use of `get_external_ip_address` and `get_any_external_ip_address` is up to you!

iliajie:
Jamie, are we good with this one too?

dynip-lib.pl
```diff
@@ -45,36 +45,85 @@ sub set_last_dynip_update
...
+	return $ipaddr if ($type == 4 && &check_ipaddress($ipaddr));
+	return $ipaddr if ($type == 6 && &check_ip6address($ipaddr));
+	return undef;
+	};
 my $now = time();
-if (!$nocache && $config{'external_ip_cache'} &&
-    $now - $config{'external_ip_cache_time'} < 24*60*60) {
+my $cache_optname = $type == 4 ?
+	'external_ip_cache' : 'external_ipv6_cache';
+my $cache_time_optname = $type == 4 ?
```
jcameron, line 73: You could just write : 

```
my $cache_time_optname = $cache_optname.'_time';
```

dynip-lib.pl
```diff
@@ -45,36 +45,85 @@ sub set_last_dynip_update
...
+	'external_ip_cache_time' : 'external_ipv6_cache_time';
+if (!$nocache && $config{$cache_optname} &&
+    $now - $config{$cache_time_optname} < 24*60*60) {
 	# Can use last cached value
-	return $config{'external_ip_cache'};
+	return $ip->($config{$cache_optname});
```
jcameron, line 78: Is there a need to pass the cached value through the `$ip` sub when presumably we know it's valid already?

jcameron:
Looks pretty good, I just added a couple more comments

dynip-lib.pl
iliajie: Yes, validation itself isn’t an issue. The problem might come from the old value storing an IPv6 address, which could lead to unexpected behavior. I’d leave it as it is—it doesn’t hurt having it.

dynip-lib.pl
iliajie: Indeed, thanks! Fixed now!

dynip-lib.pl
jcameron: But if the cache is invalid, then shouldn't we not return the value and instead fall back to re-looking it up? With the current code, if there is a value in the cache but it doesn't pass the check, we'll always return `undef`

iliajie: > But if the cache is invalid, then shouldn't we not return the value and instead fall back to re-looking it up?

So, you think we should re-check it in case it’s coming from the cache? I was just concerned that the old code might return an IPv6 address for `$config{'external_ip_cache'}`...

iliajie: > But how would `external_ip_cache` get set to an IPv6 address in the old code though?

I think, fairly easily—when `ip.cgi` returns IPv6 address on IPv6-only networks?

> My feeling is that either the check should be complete to cover all situations, or dropped.

We should not drop the check—I'm open to improving it though.

jcameron: Ok in that case, if we want to keep the check, we should change the code to : 

```
my $c = $ip->($config{$cache_optname});
return $c if ($c);
```

iliajie: > Ok in that case, if we want to keep the check, we should change the code to :
> 
> ```
> my $c = $ip->($config{$cache_optname});
> return $c if ($c);
> ```

Agreed! Fixed now!

jcameron:
Thanks for the patch!

PR #978, opened 2024-12-05
Add option to choose DNS method for detecting external IP

iliajie:
This PR adds the ability to choose how the external IP is detected. We discussed this a while ago, and Joe mentioned he liked this DNS-based method.

However, it shouldn’t be enabled by default.

We can also make the default HTTP URL for retrieving the IP configurable.


iliajie:
I'll close it, because of the dirty tree and re-submit a new one!

PR #979, opened 2024-12-05
Add option to choose DNS method for detecting external IP

iliajie:
This PR adds the ability to choose how the external IP is detected. We discussed this a while ago, and Joe mentioned he liked this DNS-based method.

However, it shouldn’t be enabled by default.

We can also make the default HTTP URL for retrieving the IP configurable.

### DIFF
--- a/dynip-lib.pl
+++ b/dynip-lib.pl
@@ -85,7 +85,12 @@ if ((my $dig = &has_command("dig")) &&
 	}
 # Fetch IP using http
 if ($error || !$out) {
-	my $url = "http://software.virtualmin.com/cgi-bin/ip.cgi";
+	my $url = $config{'ip_lookup_url'} || 
+		"http://software.virtualmin.com/cgi-bin/ip.cgi";
+	my $url4 = $url;
+	my $url6 = $url;
+	($url4, $url6) = split(/ /, $url) if ($url =~ / /);
+	$url = $type == 4 ? $url4 : $url6;
 	my ($host, $port, $page, $ssl) = &parse_http_url($url);
 	&http_download($host, $port, $page, \$out, \$error, undef, $ssl,
 		undef, undef, $timeout, 0, 1);
--- a/dynip-lib.pl
+++ b/dynip-lib.pl
@@ -63,8 +63,8 @@ my $ip = sub {
 	my $ipaddr = shift;
 	return undef if (!$ipaddr);
 	$ipaddr =~ s/\r|\n//g;
-	return $ipaddr if ($type == 4 && $ipaddr =~ /^(\d+\.\d+\.\d+\.\d+)$/);
-	return $ipaddr if ($type == 6 && $ipaddr =~ /^([0-9a-fA-F:]+)$/);
+	return $ipaddr if ($type == 4 && &check_ipaddress($ipaddr));
+	return $ipaddr if ($type == 6 && &check_ip6address($ipaddr));
 	return undef;
 	};
 my $now = time();
--- a/dynip-lib.pl
+++ b/dynip-lib.pl
@@ -45,26 +45,55 @@ else {
 	}
 }
 
-# get_external_ip_address([no-cache])
+# get_external_ip_address([no-cache], [type])
 # Returns the IP address of this system, as seen by other hosts on the Internet.
+# If the no-cache flag is set, the IP is always fetched from the network. If the
+# type is set to 6 ("ipv6"), an IPv6 address is returned. Default is 4 (IPv4).
+# For DNS resolver to work, the config option "dns_resolver" must be set to a
+# string like "myip.opendns.com resolver1.opendns.com".
 sub get_external_ip_address
 {
-my ($nocache) = @_;
+my ($nocache, $type) = @_;
+my ($out, $error);
+my $timeout = 5;
+# Validate type
+$type = 4 if (!$type || ($type != 4 && $type != 6));
+# Internal sub to validate an IP address of the correct type
+my $ip = sub {
+	my $ipaddr = shift;
+	return undef if (!$ipaddr);
+	$ipaddr =~ s/\r|\n//g;
+	return $ipaddr if ($type == 4 && $ipaddr =~ /^(\d+\.\d+\.\d+\.\d+)$/);
+	return $ipaddr if ($type == 6 && $ipaddr =~ /^([0-9a-fA-F:]+)$/);
+	return undef;
+	};
 my $now = time();
 if (!$nocache && $config{'external_ip_cache'} &&
     $now - $config{'external_ip_cache_time'} < 24*60*60) {
 	# Can use last cached value
-	return $config{'external_ip_cache'};
+	return $ip->($config{'external_ip_cache'});
 	}
-my $url = "http://software.virtualmin.com/cgi-bin/ip.cgi";
-my ($host, $port, $page, $ssl) = &parse_http_url($url);
-my ($out, $error);
-&http_download($host, $port, $page, \$out, \$error, undef, $ssl,
-	       undef, undef, 5, 0, 1);
-$out =~ s/\r|\n//g;
+# Fetch IP using DNS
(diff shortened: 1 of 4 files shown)


iliajie:
> We can also make the default HTTP URL for retrieving the IP configurable.

Also implemented https://github.com/virtualmin/virtualmin-gpl/pull/979/commits/0a38f14e3faac9f70a25c547b08ade14fc656faa.

dynip-lib.pl
```diff
@@ -45,26 +45,60 @@ sub set_last_dynip_update
...
+my ($out, $error);
+my $timeout = 5;
+# Validate type
+$type = 4 if (!$type || ($type != 4 && $type != 6));
+# Internal sub to validate an IP address of the correct type
+my $ip = sub {
+	my $ipaddr = shift;
+	return undef if (!$ipaddr);
+	$ipaddr =~ s/\r|\n//g;
+	return $ipaddr if ($type == 4 && $ipaddr =~ /^(\d+\.\d+\.\d+\.\d+)$/);
```
jcameron, line 66: We have check_ipaddress and check_ip6address functions you can use instead of your own regexp

dynip-lib.pl
```diff
@@ -45,26 +45,60 @@ sub set_last_dynip_update
...
 if (!$nocache && $config{'external_ip_cache'} &&
     $now - $config{'external_ip_cache_time'} < 24*60*60) {
 	# Can use last cached value
-	return $config{'external_ip_cache'};
+	return $ip->($config{'external_ip_cache'});
+	}
+# Fetch IP using DNS
+if ((my $dig = &has_command("dig")) &&
+    $config{'dns_resolver'} =~ /^(?<qname>\S+)\s+(?<nserv>\S+)$/) {
+	my $qname = $+{qname};
```
jcameron, line 79: What does this `$+{qname}` syntax mean?

dynip-lib.pl
```diff
@@ -45,26 +45,60 @@ sub set_last_dynip_update
...
+	&execute_command($dig_cmd, undef, \$out, \$error);
+	$out = $ip->($out);
+	}
+# Fetch IP using http
+if ($error || !$out) {
+	my $url = $config{'ip_lookup_url'} || 
+		"http://software.virtualmin.com/cgi-bin/ip.cgi";
+	my $url4 = $url;
+	my $url6 = $url;
+	($url4, $url6) = split(/ /, $url) if ($url =~ / /);
```
jcameron, line 92: Why split this on a space if the URL above doesn't have a space in it?

dynip-lib.pl
```diff
@@ -45,26 +45,60 @@ sub set_last_dynip_update
...
 	# Can use last cached value
-	return $config{'external_ip_cache'};
+	return $ip->($config{'external_ip_cache'});
+	}
+# Fetch IP using DNS
+if ((my $dig = &has_command("dig")) &&
+    $config{'dns_resolver'} =~ /^(?<qname>\S+)\s+(?<nserv>\S+)$/) {
+	my $qname = $+{qname};
+	my $nserv    = $+{nserv};
+	my $dig_cmd = "$dig +time=$timeout +short -".($type == 6 ? "6" : "4").
```
jcameron, line 81: Is this doing a DNS lookup of the system's own hostname?

dynip-lib.pl
iliajie: That’s right! Forgot about those! I’ll fix it!

dynip-lib.pl
iliajie: Those are name groups to distinctly assign from regex instead of using `$1`, `$2` and etc.

dynip-lib.pl
iliajie: For more details see the HTML help file in this PR, which explains it.

dynip-lib.pl
iliajie: No, it’s using external resolver to get IP address of the system.

dynip-lib.pl
jcameron: Oh I see ... so wouldn't it be better to have two different config options, one for IPv4 and one for IPv6 ?

dynip-lib.pl
jcameron: oh right of course!

dynip-lib.pl
iliajie: > Oh I see ... so wouldn't it be better to have two different config options, one for IPv4 and one for IPv6 ?

Having two separate fields most likely is going to be a UI clutter. Since we provide clear details in the tooltip with explanations and examples, I don’t think separate fields are necessary.

jcameron: No I disagree, using spaces to separate two different values in a text box is just confusing and unfriendly to users.

iliajie: Well, I would agree if Webmin config parser supported multiple text fields for a single radio option. I don' think it is, and please correctly me if I'm wrong.

Currently it looks pretty neat, e.g.:

image

It must be filled only once, based on the documentation tooltip we provide, e.g.:

image

... where filled in field would look as:

image

... and that is neat and clear.

I don’t see an ideal way (based on your suggestion) to make it work with separate options—it would just end up being messy...

jcameron: I still don't think it's very user friendly to expect users to enter space-separated URLs like this, but given than most users won't touch it I'm not going to press the matter..

dynip-lib.pl
jcameron: Wait I don't see the fix yet?

dynip-lib.pl
iliajie: Thanks!

dynip-lib.pl
iliajie: Sorry, just fixed now!

dynip-lib.pl
```diff
@@ -45,26 +45,60 @@ sub set_last_dynip_update
...
+my $ip = sub {
+	my $ipaddr = shift;
+	return undef if (!$ipaddr);
+	$ipaddr =~ s/\r|\n//g;
+	return $ipaddr if ($type == 4 && &check_ipaddress($ipaddr));
+	return $ipaddr if ($type == 6 && &check_ip6address($ipaddr));
+	return undef;
+	};
 my $now = time();
 if (!$nocache && $config{'external_ip_cache'} &&
```
jcameron, line 71: Shouldn't the cache be separate for IPv4 and IPv6 ?

dynip-lib.pl
```diff
@@ -45,26 +45,60 @@ sub set_last_dynip_update
...
 	# Can use last cached value
-	return $config{'external_ip_cache'};
+	return $ip->($config{'external_ip_cache'});
+	}
+# Fetch IP using DNS
+if ((my $dig = &has_command("dig")) &&
+    $config{'dns_resolver'} =~ /^(?<qname>\S+)\s+(?<nserv>\S+)$/) {
+	my $qname = $+{qname};
+	my $nserv    = $+{nserv};
+	my $dig_cmd = "$dig +time=$timeout +short -".($type == 6 ? "6" : "4").
```
jcameron, line 81: Use quotemeta for `$timeout` and `$qname` below

dynip-lib.pl
iliajie: Yes, it will be separated after merge, as in this conflicting PR https://github.com/virtualmin/virtualmin-gpl/pull/974/commits/c1668847b49a18daa6825cc6f04ade07ec4ca89b.

dynip-lib.pl
iliajie: No need to use it for `$timeout`, as we have control over it. `$qname` - yes, agreed.

dynip-lib.pl
iliajie: Here https://github.com/virtualmin/virtualmin-gpl/commit/99c4dcd2aa6c9c3adf8d03003e022de874c4be1c.

Issue #981, opened 2024-12-09
[ALERT] Big bug since the last Virtualmin / PHP-FPM update


SIP-Online:
Hello,
I'm alerting you to a bug that caused a hell of a mess on version **7.30.2**.
All my spaces (Domains / Subdomains) are changed to PHP 8.2, after searching in (**Webmin Actions Log**), this is linked to the virtual-server/run-postinstalls.pl step:
```
> lastsave_script=/usr/share/webmin/run-postinstalls.pl
138c138
< php_fpm_version=8.0
---
> php_fpm_version=8.2
```

On another:
```
116c116
< php_fpm_version=7.4
---
> php_fpm_version=8.2
```
.....................................................
..................................
..............

This on all my web spaces in PHP 7.4, 8.0, 8.1...

On the tree side, everything is however still in the right places: **/etc/php/\*/fpm/pool.d/\***

This is clearly a bug that deserves to stop this update, because it will quickly degenerate Forums side.

I don't know how you're going to be able to catch up on this?
In the meantime, I've backed up the "**/etc/php**" folder

SIP-Online:
I have corrected it manually for the moment, using the following folders: **/etc/php/\*/fpm/pool.d/***
Here are the corrected files (**php_fpm_version**): **/etc/webmin/virtual-server/domains/\***
And also this one: **/etc/webmin/phpini/\***

**Restarted webmin, and also emptied the Template cache.**

Fortunately there were few web spaces on this server, but I am eager to know when it will be corrected, so that I can continue updating my other servers.

I think a batch script can auto-correct this bug.

iliajie:
Hello,

Thanks for the heads up!

> All my spaces (Domains / Subdomains) are changed to PHP 8.2, after searching in (**Webmin Actions Log**), this is linked to the virtual-server/run-postinstalls.pl step:

What is set in "System Settings ⇾ Server Templates" for top level and sub domain for "PHP options" sub-page? Can you provide screenshots for it?


> This is clearly a bug that deserves to stop this update, because it will quickly degenerate Forums side.

Yet, I could not reproduce this bug. Could you provide more details? Which version of Virtualmin did you have installed before upgrading to Virtualmin 7.30.2? What changes, if any, were made to the PHP configuration after the package upgrade?

> I have corrected it manually for the moment, using the following folders: **/etc/php/*/fpm/pool.d/***
[...]

PR #983, opened 2024-12-10
Fix the logic when fixing domain FPM config

iliajie:
Hello Jamie,

As discussed in [this thread](https://github.com/virtualmin/virtualmin-gpl/issues/981#issuecomment-2530400523), you mentioned that this part of the code might be the culprit behind the original issue mentioned in the ticket. I started digging deeper into the code and found something that needed attention.

However, I’m not entirely sure if it will fix the original issue—it might, though, as there is code that expects the domain’s `php_fpm_version` option to always be set, such as `list_domain_php_directories`.

Have a look please!

### DIFF
--- a/php-lib.pl
+++ b/php-lib.pl
@@ -2038,7 +2038,7 @@ if ($file =~ m{/php/(\d+)\.(\d+)/fpm} ||     # Debian and derivatives
 		return $version;
 		}
 	}
-return $avail[0] if @avail;
+return $avail[0] if (@avail && @avail == 1); # Return the only available version
 return undef;
 }
 
--- a/php-lib.pl
+++ b/php-lib.pl
@@ -2016,7 +2016,6 @@ my @rv;
 foreach my $fpmconf (@fpmconfs) {
 	my $file = $fpmconf->{'dir'}."/".$d->{'id'}.".conf";
 	if (-r $file) {
-		$fpmconf->{'file'} = $file;
 		push(@rv, $fpmconf);
 		}
 	}
--- a/php-lib.pl
+++ b/php-lib.pl
@@ -2006,6 +2006,42 @@ if (($ver = &get_domain_php_version($d, $mode)) &&
 	}
 }
 
+# find_php_fpm_config(&domain)
+# Find PHP-FPM config pool file for a domain
+sub find_php_fpm_config
+{
+my ($d) = @_;
+my @fpmconfs = &list_php_fpm_configs();
+my $fpmconf_file;
+foreach my $fpmconf (@fpmconfs) {
+	my $file = $fpmconf->{'dir'}."/".$d->{'id'}.".conf";
+	if (-r $file) {
+		# More than one file found, return undef (error)
+		return undef if ($fpmconf_file);
+		$fpmconf_file = $file;
+		}
+	}
+return $fpmconf_file;
+}
+
+# detect_php_fpm_version(&domain, [file])
+# Returns the PHP-FPM version used by a domain, based on the config file name
+sub detect_php_fpm_version
+{
+my ($d, $file) = @_;
+$file ||= &find_php_fpm_config($d);
+my @avail = map { $_->[0] } &list_available_php_versions($d, 'fpm');
+if ($file =~ m{/php/(\d+)\.(\d+)/fpm} ||     # Debian and derivatives
+    $file =~ m{/php(\d{1,2})(\d)/.*-fpm}) {  # RHEL and derivatives
+	my $version = "$1.$2";
+	if (&indexof($version, @avail) >= 0) {
+		return $version;
+		}
+	}
+return $avail[0] if @avail;
+return undef;
+}
+
 # get_php_fpm_config([version|&domain])
 # Returns the first valid FPM config for a domain
 sub get_php_fpm_config
--- a/php-lib.pl
+++ b/php-lib.pl
@@ -2007,38 +2007,36 @@ if (($ver = &get_domain_php_version($d, $mode)) &&
 }
 
 # find_php_fpm_config(&domain)
-# Find PHP-FPM config pool file for a domain
+# Find known PHP-FPM pools for a domain
 sub find_php_fpm_config
 {
 my ($d) = @_;
 my @fpmconfs = &list_php_fpm_configs();
-my $fpmconf_file;
+my @rv;
 foreach my $fpmconf (@fpmconfs) {
 	my $file = $fpmconf->{'dir'}."/".$d->{'id'}.".conf";
 	if (-r $file) {
-		# More than one file found, return undef (error)
-		return undef if ($fpmconf_file);
-		$fpmconf_file = $file;
+		$fpmconf->{'file'} = $file;
+		push(@rv, $fpmconf);
 		}
 	}
-return $fpmconf_file;
+return @rv;
 }
 
-# detect_php_fpm_version(&domain, [file])
-# Returns the PHP-FPM version used by a domain, based on the config file name
+# detect_php_fpm_version(&domain)
+# Returns PHP-FPM version used by domain or undef if not found or on error
 sub detect_php_fpm_version
-{
-my ($d, $file) = @_;
-$file ||= &find_php_fpm_config($d);
-my @avail = map { $_->[0] } &list_available_php_versions($d, 'fpm');
-if ($file =~ m{/php/(\d+)\.(\d+)/fpm} ||     # Debian and derivatives
-    $file =~ m{/php(\d{1,2})(\d)/.*-fpm}) {  # RHEL and derivatives
-	my $version = "$1.$2";
-	if (&indexof($version, @avail) >= 0) {
-		return $version;
+{ 
+my ($d) = @_;
+my @dom_vers = map { $_->{'shortversion'} } &find_php_fpm_config($d);
+return undef if (@dom_vers > 1); # Return error if more than one version found
+my @vers = map { $_->[0] } &list_available_php_versions($d, 'fpm');
+foreach $dom_ver (@dom_vers) {
+	if (&indexof($dom_ver, @vers) >= 0) {
+		return $dom_ver;
 		}
 	}
-return $avail[0] if (@avail && @avail == 1); # Return the only available version
+return $vers[0] if (@vers && @vers == 1); # Return the only available version
 return undef;
 }
 
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -16091,27 +16091,9 @@ if (&domain_has_website()) {
 
 			# Try to detect actually configured version
 			$dv = &detect_php_fpm_version($d);
-			# If still none exists, check template if anything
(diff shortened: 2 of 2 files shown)


jcameron:
Does this really solve the OP's problem of the `php_fpm_version` being changed incorrectly? The issue wasn't that it was not set..

IMHO if it's not set all, we shouldn't even mark the domain as being in FPM mode.

iliajie:
> Does this really solve the OP's problem of the php_fpm_version being changed incorrectly? The issue wasn't that it was not set

This can indirectly impact such changes because `list_domain_php_directories` might return an empty value for the `version` key, which could cause issues.

> IMHO if it's not set all, we shouldn't even mark the domain as being in FPM mode.

If we can detect that the mode is FPM, which we already do, shouldn’t we apply the logic I suggested in this PR? I think we do, as solely relying on the presence of `php_fpm_version` is potentially risky.

jcameron:
No I don't think we can safely apply this logic, because we have no idea what the version actually is. 

Maybe another safe option is to check ALL the pool files and see which ones are connected to which domains, and correct php_fpm_version based on that?

iliajie:
> No I don't think we can safely apply this logic, because we have no idea what the version actually is.
> 
> Maybe another safe option is to check ALL the pool files and see which ones are connected to which domains, and correct php_fpm_version based on that?

In fact, we are not fixing the running PHP version itself; we are just fixing the association in the Virtualmin domain config file. Perhaps we should just do (leave as is) what you have already done in the latest related patches? I mean showing nothing in the select dropdown in the UI if `php_fpm_version` is missing? The only suggestion I'd make is to show an error if no PHP version was selected (empty select) at the time of form submission, to make sure an accidental version won't be configured for the domain?

> Maybe another safe option is to check ALL the pool files and see which ones are connected to which domains, and correct php_fpm_version based on that?

What is the benefit and real issue here? We are not automatically changing the actual PHP version for the domain - we're just changing an association in domain config file. So why are we going down this rabbit hole?

Besides, if we don't update an actual PHP version, how is the original issue on its own even possible?

jcameron:
The reason this needs to be correct is that if the FPM version in the domain config is wrong, Virtualmin cannot find the PHP FPM pool .conf file, as it needs to know the version to find the correct base directory. So just showing an error message isn't great, as it means the domain can be in a broken state until the version is fixed.

iliajie:
> The reason this needs to be correct is that if the FPM version in the domain config is wrong, Virtualmin cannot find the PHP FPM pool .conf file, as it needs to know the version to find the correct base directory.

In this case you’re right—we’re _probably_ not handling this correctly. But, the FPM pool file name is unique to the domain ID, so if `php_fpm_version` is missing, we could easily figure out actual PHP version based on FPM pool file path.

_**However**_, it’s not clear how that would happen in the first place!? I’d like to know how exactly the [issue discussed](https://github.com/virtualmin/virtualmin-gpl/issues/981#issuecomment-2530400523) could be actually reproduced, because right now it feels like we’re addressing a hypothetical, unrealistic bug fixing which might ultimately cause more harm than good.

That said, I don’t see how the post-install script could cause this!? The config check might potentially set `php_fpm_version` to something unexpected, but actually changing the PHP version would require manual action on the "PHP Options" page or using the reset feature—correct me if I'm wrong!!??

iliajie:
But, okay! I see the bug. However, this issue can’t occur on its own—a user has to click the “Save” button on the “PHP Options” page or use the reset feature for it to happen.

The issue is that Virtualmin doesn’t know which PHP version it’s dealing with (fixing) and ends up restarting the wrong PHP-FPM server when `php_fpm_version` is missing. However, it’s not a normal or critical problem because for `php_fpm_version` to be missing or misaligned, it would require something unusual, like removing all PHP version-related packages that are in use.

And, to be clear, reset feature is _always_ broken, despite of `php_fpm_version` presence on the domain config.

That said, we should definitely fix it. But I don't see anything fundamentally urgent in it. Again, please correct me if I'm wrong.

jcameron:
> And, to be clear, reset feature is always broken, despite of php_fpm_version presence on the domain config.

Do you mean it's broken because `php_fpm_version` is being cleared when it shouldn't be?

iliajie:
> Do you mean it's broken because `php_fpm_version` is being cleared when it shouldn't be?

No, it's always broken despite. For Apache it ends up broken and for Nginx it doesn't even work.

iliajie:
> Maybe another safe option is to check ALL the pool files and see which ones are connected to which domains, and correct php_fpm_version based on that?

Implemented here https://github.com/virtualmin/virtualmin-gpl/pull/983/commits/852053e0e829cd95b7f6ef19332a049ae3078354.

iliajie:
> > Maybe another safe option is to check ALL the pool files and see which ones are connected to which domains, and correct php_fpm_version based on that?
> 
> Implemented here [852053e](https://github.com/virtualmin/virtualmin-gpl/commit/852053e0e829cd95b7f6ef19332a049ae3078354).

Jamie, please have a look at this patch [above.](https://github.com/virtualmin/virtualmin-gpl/commit/852053e0e829cd95b7f6ef19332a049ae3078354).

jcameron:
I need to take some time to see if this solution makes sense and isn't over-complicated.

IMHO the original bug was a rare case that we've never seen before despite this FPM fix code being in place for many releases. So we don't need to rush out a fix.

iliajie:
> I need to take some time to see if this solution makes sense and isn't over-complicated.

Thanks!

> IMHO the original bug was a rare case that we've never seen before despite this FPM fix code being in place for many releases. So we don't need to rush out a fix.

Great, I agree!

Then we can technically proceed with tagging 7.30.3. Unless you want to cherry pick more patches into it?

jcameron:
Yes I think we can proceed with 7.30.3

php-lib.pl
```diff
@@ -2006,6 +2006,42 @@ sub create_php_bin_links
 	}
 }
 
+# find_php_fpm_config(&domain)
+# Find PHP-FPM config pool file for a domain
+sub find_php_fpm_config
```
jcameron, line 2011: Wouldn't it be cleaner if this function returned the hash ref from `list_php_fpm_configs` instead of the file path? 

That way when it's called, `detect_php_fpm_version` wouldn't have to re-compute the FPM version because it's already in the hash.

php-lib.pl
```diff
@@ -2006,6 +2006,42 @@ sub create_php_bin_links
...
+{
+my ($d) = @_;
+my @fpmconfs = &list_php_fpm_configs();
+my $fpmconf_file;
+foreach my $fpmconf (@fpmconfs) {
+	my $file = $fpmconf->{'dir'}."/".$d->{'id'}.".conf";
+	if (-r $file) {
+		# More than one file found, return undef (error)
+		return undef if ($fpmconf_file);
```
jcameron, line 2020: Better to return multiple hashes if for some reason a single domain is in multiple FPM directories.

iliajie: It’s intentional—if the config is in multiple FPM directories, it means we can’t reliably detect it, and something is definitely wrong.

jcameron: Oh yeah for sure, and that's something we should reported on as part of validation! But for this internal function, it's more useful if it returns all the FPM configs found.

php-lib.pl
jcameron: My point is that in `detect_php_fpm_version` is uses the file path to figure out the version, but the hashes returned by `list_php_fpm_configs` already contain the version! So there's no need to duplicate that logic when we already have the information we want.

iliajie: Ah, understood now! Fixed!

php-lib.pl
iliajie: Also fixed!

php-lib.pl
```diff
@@ -2006,6 +2006,40 @@ sub create_php_bin_links
...
+# Find known PHP-FPM pools for a domain
+sub find_php_fpm_config
+{
+my ($d) = @_;
+my @fpmconfs = &list_php_fpm_configs();
+my @rv;
+foreach my $fpmconf (@fpmconfs) {
+	my $file = $fpmconf->{'dir'}."/".$d->{'id'}.".conf";
+	if (-r $file) {
+		$fpmconf->{'file'} = $file;
```
jcameron, line 2019: It's not safe to do this, because `list_php_fpm_configs` is a global function that returns hashes not related to any particular domain. Instead callers can just compute this path given a domain ID.

php-lib.pl
jcameron: Thanks!

virtual-server-lib-funcs.pl
```diff
@@ -16074,25 +16074,46 @@ sub check_virtual_server_config
 		# Check for invalid FPM versions, in case one has been
 		# upgraded to a new release
 		my @fpmfixed;
+		# Sort to have highest version first
 		@fpms = sort { &compare_versions(
-				$a->{'version'}, $b->{'version'}) } @fpms;
+				$b->{'version'}, $a->{'version'}) } @fpms;
```
jcameron, line 16079: Why change the sort order?

php-lib.pl
iliajie: Got it! Fixed now!

virtual-server-lib-funcs.pl
iliajie: We want to return highest if we ever reach this part of the code https://github.com/virtualmin/virtualmin-gpl/pull/983/files#diff-a8b0a88367118c8f13666eed07287b13ecc217c8fdfaf6a34d51b226dc196f92R16107-R16111.

virtual-server-lib-funcs.pl
```diff
@@ -16074,25 +16074,46 @@ sub check_virtual_server_config
...
 
-			# Find the existing version just above the one that
-			# was stored, or alternately the highest available
-			my ($nf) = grep { &compare_versions($_->{'shortversion'}, $dv) > 0 } @fpms;
-			$nf ||= $fpms[$#fpms];
+			# Try to detect actually configured version
+			$dv = &detect_php_fpm_version($d);
+			# If still none exists, check template if anything
+			# is preferred otherwise fall back to highest available
+			if (!$dv) {
```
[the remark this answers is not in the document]
iliajie: I agree! Since we can now determine the actual PHP versions used based on the pool file, there's no need to guess or fallback. Fixed on the latest commit!

virtual-server-lib-funcs.pl
jcameron: I don't think we even need this array anymore

iliajie: That's right! Dropped in the latest commit.

jcameron:
Thanks for this patch!

iliajie:
Sure, I’ll cherry-pick it to 7.30.3 branch tomorrow.

Is there anything else we’re missing for the new release?

Do you think my commits with added file locking will solve the issue with random configs corruption?

jcameron:
I think we're good for a new release. And I hope that locking fix resolves the issue!

PR #986, opened 2024-12-16
Fix to return error when downgraded from Pro to GPL

iliajie:
Hello Jamie!

This small patch address the [issue](https://forum.virtualmin.com/t/cloudflare-dns-issue-on-gpl-downgrade/129712?u=ilia) and failure when Virtualmin was downgraded from Pro to GPL.

### DIFF
--- a/feature-dns.pl
+++ b/feature-dns.pl
@@ -3776,6 +3776,10 @@ if ($d->{'dns_cloud'}) {
 	# actual file.
 	my $ctype = $d->{'dns_cloud'};
 	my $gfunc = "dnscloud_".$ctype."_get_records";
+	if (!defined(&$gfunc)) {
+		return ("This cloud DNS provider isn't supported in ".
+		        "Virtualmin GPL");
+		}
 	my $info = { 'domain' => $d->{'dom'},
 		     'id' => $d->{'dns_cloud_id'},
 		     'location' => $d->{'dns_cloud_location'} };


jcameron:
Thanks for the patch!

iliajie:
Thanks for merging it! However, with this patch, when visiting the “DNS Options” page, it silently displays the page with local DNS selected, even though the domain config has cloud DNS configured. On the other hand, visiting the “DNS Records” page correctly shows the error as it should.

jcameron:
Maybe the DNS Options page should show an error indicating that the current provider is no longer valid?

iliajie:
Yeah, that would be best. However, the error is already being returned in the patch in this PR, but it isn’t displayed. Perhaps you mean to show a warning alert _after_ the “DNS Options” page is opened? Or would it be better to just display an error message without loading the page at all, similar to “DNS Records” page. I think latter is better...

jcameron:
I've checked in a fix that will cause the DNS Options page to show an error (and nothing else) if the selected DNS provider is not valid : https://github.com/virtualmin/virtualmin-gpl/commit/d2657d71808549be3668af037084156dcebeba08

iliajie:
> I've checked in a fix that will cause the DNS Options page to show an error (and nothing else) if the selected DNS provider is not valid : [d2657d7](https://github.com/virtualmin/virtualmin-gpl/commit/d2657d71808549be3668af037084156dcebeba08)

Thanks! I've also added a small patch to display error messages consistently across pages https://github.com/virtualmin/virtualmin-gpl/commit/6d6c1e785f38201b45f1f800bfecfc132616aba1.

PR #989, opened 2024-12-19
Add more file locking

iliajie:
Hello Jamie,

I looked through all code that can be called from scheduled functions and found several places where file locking was missing, and wasn't set by the caller.

Please have a look!


jcameron:
Thanks for the patch!

iliajie:
You're welcome!

I will add it to 7.30.3, update change log and we're good to go.

PR #990, opened 2024-12-23
Add support to return a list of files for which permissions were fixed

iliajie:
Hello Jamie!

This PR adds support for returning a list of files for which permissions were fixed.

It took longer to implement than I initially expected because of the additional work needed to update themes to support the new layout.

Example:

https://github.com/user-attachments/assets/066d0143-9bea-49ae-9e9b-448564ffa191

### DIFF
--- a/feature-dir.pl
+++ b/feature-dir.pl
@@ -956,7 +956,7 @@ foreach my $sd ($d, &get_domain_by("parent", $d->{'id'})) {
 
 # Track changed files
 my @changed_files;
-my $change_files = sub {
+my $check_file_change = sub {
 	my ($f, $uid, $gid, $changed_files) = @_;
 	my ($curr_uid, $curr_gid) = (stat($f))[4,5];
 	my $curr_user = getpwuid($curr_uid) || $curr_uid;
@@ -988,11 +988,11 @@ LOOP: while(my $f = <FIND>) {
 	foreach my $s (@subhomes) {
 		next LOOP if ($f =~ /^\Q$s\E/);
 		}
-	$change_files->($f, $d->{'uid'}, $gid, \@changed_files);
+	$check_file_change->($f, $d->{'uid'}, $gid, \@changed_files);
 	&set_ownership_permissions($d->{'uid'}, $gid, undef, $f);
 	}
 close(FIND);
-$change_files->($d->{'home'}."/".$hd, $d->{'uid'}, $gid, \@changed_files);
+$check_file_change->($d->{'home'}."/".$hd, $d->{'uid'}, $gid, \@changed_files);
 &set_ownership_permissions($d->{'uid'}, $gid, undef, $d->{'home'}."/".$hd);
 foreach my $dir (&virtual_server_directories($d)) {
 	&set_ownership_permissions(undef, undef, oct($dir->[1]),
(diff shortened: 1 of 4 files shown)


fix-domain-permissions.pl
```diff
@@ -81,12 +81,23 @@ package virtual_server;
...
 	else {
-		$err = &set_home_ownership($d);
-		if ($err) {
-			&$second_print(".. failed : $err");
+		my (@changed_files) = &set_home_ownership($d);
+		if (@changed_files) {
+			my $details_body;
+			foreach my $changed (@changed_files) {
+				$changed->[0] =~ s/^$d->{'home'}\///;
+				$details_body .= "     ".
```
jcameron, line 89: You can use the function `&$indent_print();` to start an indented block rather than manually adding spaces

feature-dir.pl
```diff
@@ -954,6 +954,21 @@ sub set_home_ownership
 		}
 	}
 
+# Track changed files
+my @changed_files;
+my $change_files = sub {
```
jcameron, line 959: Better to name this like `find_change_files` as the current name implies that the function is actually doing the changes.

feature-dir.pl
```diff
@@ -954,6 +954,21 @@ sub set_home_ownership
...
+	my ($f, $uid, $gid, $changed_files) = @_;
+	my ($curr_uid, $curr_gid) = (stat($f))[4,5];
+	my $curr_user = getpwuid($curr_uid) || $curr_uid;
+	my $curr_group = getgrgid($curr_gid) || $curr_gid;
+	my $new_user = getpwuid($uid) || $uid;
+	my $new_group = getgrgid($gid) || $gid;
+	if ($curr_uid != $uid || $curr_gid != $gid) {
+		push(@$changed_files, 
+			[$f, "$curr_user:$curr_group", "$new_user:$new_group"]);
+		}
```
jcameron, line 969: How does this function handle sub-directories?

feature-dir.pl
iliajie: That's right! Thanks! Though, I renamed it to `check_file_change` because this function doesn't search for anything.

feature-dir.pl
iliajie: This function works with a single file passed to it, and it doesn't care if it's a file or a directory.

fix-domain-permissions.pl
```diff
@@ -81,12 +81,23 @@ package virtual_server;
...
+				}
+			&$second_print(
+				$text{'fixperms_oklist2'}. "\n".
+				$details_body);
 			}
 		else {
-			&$second_print(".. done");
+			&$second_print($text{'setup_done'});
 			}
 		}
```
iliajie, line 102: Thanks, Jamie! I know, but sadly, it won't work nicely. It won't print text grouped in a way we want it to be. It will mess up spacing, no matter of the combination of the first and second print we use...

feature-dir.pl
iliajie: We pass and modify array we track by reference with `\@changed_files`...

jcameron: I get that - but is the point of this sub to return all the files in directories and sub-directories under `$f` ? Because the `chown -R` call later will recursively process files in sub-directories.

iliajie: Ah, right! I understand what you mean now. I hadn't considered this scenario. The only way to do it cleanly is to iterate `@users` using the `find` command, as in the block above...

jcameron: Actually why not just write your own function to do the work of `chown` at the same time as tracking which files were changed?

iliajie: Because we shouldn't shell out to recursively change permissions or find files, it isn't best option. The fix to this particular issue we are discussing is here https://github.com/virtualmin/virtualmin-gpl/pull/990/commits/0f5f997ceae2aeb473e883d9e409259402d19911.

So, I believe we should replace the use of the `find` command with Perl's native `File::Find`, which is always available to us, in a Webmin core function. This function will locate files and accept the parameters like `$dir`, `$func_reference_to_work_on_results`, and `$hash_of_opts_to_file_find`. So, if you don't mind, I will submit a new PR to the Webmin repo to address this...

jcameron: Yes factoring out a separate recursive search function would be a good idea in future!

iliajie: > Yes factoring out a separate recursive search function would be a good idea in future!

Agreed!

PR #1012, opened 2025-02-25
Allow link cert with different owner while modifying domain in CLI

xxorax:
Linking cert is perfectly working while creating a domain, however it was not allowed to do so while modifying it.

This change allows linking a certificate when modifying a domain, based on the domain's `link_cert` configuration or falling back to the main `nolink_certs` main configuration.

This may required an additional option `--always-link-ssl-cert` similar to what we have in create-domain, but I'm not sure how to handle the domain config : `--link-cert`in modify-web does not originally save `link_cert` configuration on the domain and this config is mostly never saved unless you had specified `--break-ssl-cert` or `--link-ssl-cert` in the create-domain CLI only,

Anyway, this change brings an easy way to switch to a existing wildcard certificate owned by an other domain.

Thanks

### DIFF
--- a/modify-web.pl
+++ b/modify-web.pl
@@ -894,7 +894,11 @@ foreach $d (@doms) {
 			my @sames = &find_matching_certificate_domain($d);
 
 			# Determine if different owner linking is allowed
-			my $diff_owner_allowed = $d->{'link_certs'} == 2;
+			my $diff_owner_allowed = 2 == (
+				defined $d->{'link_certs'} ?
+				$d->{'link_certs'} :
+				$config{'nolink_certs'}
+				);
 
 			# Initialize priority stack
 			my @priorities = (
--- a/modify-web.pl
+++ b/modify-web.pl
@@ -892,21 +892,38 @@ foreach $d (@doms) {
 		else {
 			# Find a cert to link with, ideally a parent with the same owner
 			my @sames = &find_matching_certificate_domain($d);
-			my ($same) = grep { $_->{'user'} eq $d->{'user'} &&
-					    !$_->{'parent'} } @sames;
-			if (!$same) {
-				($same) = grep { $_->{'user'} eq $d->{'user'} } @sames;
+
+			# Determine if different owner linking is allowed
+			my $diff_owner_allowed = $d->{'link_certs'} == 2;
+
+			# Initialize priority stack
+			my @priorities = (
+				sub { grep { $_->{'user'} eq $d->{'user'} && $_->{'parent'} } @sames },
+				sub { grep { $_->{'user'} eq $d->{'user'} } @sames },
+			);
+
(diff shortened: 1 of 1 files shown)


modify-web.pl
```diff
@@ -892,21 +892,42 @@ package virtual_server;
...
+			# Determine if different owner linking is allowed
+			my $diff_owner_allowed = 2 == (
+				defined $d->{'link_certs'} ?
+				$d->{'link_certs'} :
+				$config{'nolink_certs'}
+				);
+
+			# Initialize priority stack
+			my @priorities = (
+				sub { grep { $_->{'user'} eq $d->{'user'} && $_->{'parent'} } @sames },
```
jcameron, line 905: Why is `sub` used here?

xxorax: I think it's clearer to define filters first and then run them as sub.
If we don't use them, the code will contain many ... = ... grep ... if ... statements, making it harder to read and understand.

By structuring it this way, the code remains readable, allows easy changes in priority, and makes it simple to add new cases (e.g., tomorrow retrieving a certificate from a remote server) without affecting performance (mostly).
Of course there is always better (pre-grouping by user and parent etc...) but I think it's overkill.

modify-web.pl
```diff
@@ -892,21 +892,42 @@ package virtual_server;
 		else {
 			# Find a cert to link with, ideally a parent with the same owner
 			my @sames = &find_matching_certificate_domain($d);
-			my ($same) = grep { $_->{'user'} eq $d->{'user'} &&
-					    !$_->{'parent'} } @sames;
-			if (!$same) {
-				($same) = grep { $_->{'user'} eq $d->{'user'} } @sames;
+
+			# Determine if different owner linking is allowed
+			my $diff_owner_allowed = 2 == (
```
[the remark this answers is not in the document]
jcameron: Oh I see, makes sense

modify-web.pl
jcameron: But isn't : 

`sub { grep { $_->{'user'} eq $d->{'user'} } @sames },`

just the same as : 

`(grep { $_->{'user'} eq $d->{'user'} } @sames),`

xxorax: Using sub does not evaluate the expression all at once. Instead, it processes each sub one by one until it finds a matching certificate. Without sub, all priorities would be evaluated first, and then a match would be found.
It is easier to maintain and evolve, and the performance gain can be significant when dealing with a large number of certificates.

jcameron: Honestly I would prefer more readable code over a tiny increase in efficiency in a code path that is rarely executed

xxorax: So I removed the subs
Thx

jcameron:
Thanks for the patch!

PR #1014, opened 2025-03-03
Add flush and sync before closing

iliajie:
G'day Jamie!

This is a similar PR we added to Webmin yesterday.

### DIFF
--- a/security-lib.pl
+++ b/security-lib.pl
@@ -1,6 +1,7 @@
 
 use feature 'state';
 use Fcntl ':mode';
+use IO::Handle;
 
 # Functions for accessing files and running commands as a domain owner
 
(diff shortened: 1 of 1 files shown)


security-lib.pl
```diff
@@ -291,6 +292,16 @@ sub open_tempfile_as_domain_user
 			exit(2);
 			}
 		}
+	# Add flush and sync before closing
+	unless (IO::Handle::flush(*FILE)) {
```
jcameron, line 296: Why is this flush done differently to below where we just use `$fh->flush();` ?

iliajie: The difference is that `$fh->flush()` works for filehandle variables, while bareword filehandles like `FILE` require the static method call `IO::Handle::flush(*FILE)` with typeglob notation...

jcameron: Ah I see ... so maybe actually this code should use a proper file handle variable instead of `FILE` ?

iliajie: Nah, I think it’s fine, as there are dozen of places where barewords are used—changing it here won’t benefit anything.

jcameron: Ok let's stick with this then

PR #1016, opened 2025-03-07
Update SSL certificate renewal to use 7 days instead of 30

iliajie:
Hello Jamie!

Here is the PR for the change we were discussing [here](https://github.com/virtualmin/virtualmin-gpl/commit/8b9dd0230cd9e253e07c0f30020e3766297e196b#r153400386).


jcameron:
Thanks for the fix!

Issue #1018, opened 2025-03-10
DKIM configuration is destroyed by virtualmin post install if the domain is not configured for dns


aqueos:
hi,

 I recently had a webmin update/

var/log/apt/history.log:Upgrade: webmin:amd64 (2.202, 2.302)

 And we noticed again that our dkim configuration was modified. Extra domains disappeared from the configuration.

 If in https://xxxxxxxx:10000/virtual-server/dkim.cgi?xnavigation=1 i have a bunch of "Extra domains to sign for" 
but those domains do not have the DNS activated for them in virtualmin (normal none use the virtualmin dns feature that is disabled completly). 

The postinstall script when you upgrade the package on debian do

```
		# Replace the list of extra DKIM domains with a new field, as
		# long as they are Virtualmin domains
		my @newextra;
		foreach my $e (@{$dkim->{'extra'}}) {
			my $d = &get_domain_by("dom", $e);
			if ($d) {
				&lock_domain($d);
				$d->{'dkim_enabled'} = 1;
				&save_domain($d);
				&unlock_domain($d);
				}
			else {
				push(@newextra, $e);
				}
			}
		$config{'dkim_extra'} = join(' ', @newextra);
		}
	&save_module_config();
```

 but this remove the domains as they are with the

```
DKIM private key in PEM format
	
  Use global default key 
```

option with no DNS feature enable for the domain.

 the

` $d->{'dkim_enabled'} = 1; &save_domain($d);  `

seems to do nothing if the DNS is not enabled on the domain. Therefor this does not do 

`$config{'dkim_extra'} = join(' ', @newextra); `

So the change add a configuration that seems to do nothing if DNS is not enabled AND remove the domain from dkim_extra.

Therefor it seems the configuration for all those domains is removed from dkim_extra and not added anywhere else and emails are not signed anymore.

best regards,
Ghislain. 
AQUEOS.

jcameron:
Thanks for the bug report and fix! So does your patch resolve the issue for you?

iliaross:
Reopened, as this was closed by an automated cleanup run.

⚠️ If the issue is resolved, please take a moment to close it to help keep the project organized.

PR #1019, opened 2025-03-10
Postinstall: Remove domain from dkim_extra only if domain exist and h…

lnicolas83:
Postinstall: Remove domain from dkim_extra only if domain exist and have dns feature


lnicolas83:
#1018

PR #1020, opened 2025-03-10
Add support for the new navigation menu layout

iliajie:
Hello Jamie!

As we discussed a while back in the chat, here’s the PR that adds support for the new menu layout. This update also addresses issues with labels for virtual servers, sub-servers, and alias servers.

The mail feature for alias domains is now called “Mail for alias” and has its own help tip, since there’s no way to edit mail settings for the domain—I wanted to keep things clear.

Also, let’s discuss whether a home directory is really necessary for an alias with its own mail, and if auto-replies work as intended or are needed at all?

Below are some examples of the reworked menu layout (supported in current Webmin 2.302), based on Joe’s suggestions and numerous user suggestions:

**Create top-level domain menu and form**

image

**Create sub-server domain menu and form**

image

**Create alias in top-level domain menu and form**

image

**Edit sub-domain menu and form**

image

**Edit alias server menu and form**

image

### DIFF
--- a/domain_form.cgi
+++ b/domain_form.cgi
@@ -81,8 +81,7 @@ if ($in{'subdom'}) {
 	$subdom = &get_domain($in{'subdom'});
 	$subdom || &error(&text('form_esubdom', $in{'subdom'}));
 	}
-my $subhead = $aliasdom || $parentdom;
-$subhead = &domain_in($aliasdom || $parentdom) if ($subhead);
+my $subhead = $aliasdom || $parentdom ? &domain_in($aliasdom || $parentdom) : "";
 &ui_print_header($subhead, $aliasdom ? $text{'form_title3'} :
 			   $subdom ? $text{'form_title4'} :
 			   $parentdom ? $text{'form_title2'} :
--- a/domain_form.cgi
+++ b/domain_form.cgi
@@ -50,57 +50,6 @@ elsif ($in{'enable'}) {
 &can_create_master_servers() || &can_create_sub_servers() ||
 	&error($text{'form_ecannot'});
 
-# If we are in generic mode, work out all possible modes for the current user
-if ($in{'generic'}) {
-	$gparent = &get_domain($in{'gparent'}) if ($in{'gparent'});
-	($tdleft, $tdreason, $tdmax) = &count_domains("topdoms");
-	if (&can_create_master_servers()) {
-		# Top-level server
-		if ($tdleft) {
-			push(@generics, [ $text{'form_generic_master'}, '' ]);
-			}
-		}
-	if (&can_create_sub_servers() && $gparent) {
-		# Sub-server under parent's user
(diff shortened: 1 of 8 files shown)


domain_form.cgi
```diff
@@ -132,45 +81,28 @@ if ($in{'subdom'}) {
...
-&ui_print_header(undef, $aliasdom ? $text{'form_title3'} :
-			$subdom ? $text{'form_title4'} :
-			$parentdom ? $text{'form_title2'} :
-				     $text{'form_title'}, "",
-			$aliasdom ? "create_alias" :
-			$subdom ? "create_subdom" :
-			$parentdom ? "create_subserver" :
-				  "create_form");
+my $subhead = $aliasdom || $parentdom;
+$subhead = &domain_in($aliasdom || $parentdom) if ($subhead);
```
jcameron, line 85: This is a bit confusing because $subhead is first a domain and then a string, which in a typed language would be illegal. Change it to : 

`my $subhead = $aliasdom || $parentdom ? &domain_in($aliasdom || $parentdom) : ""
`

jcameron:
Thanks! Added one comment, but overall this looks good!

iliajie:
> Thanks! Added one comment, but overall this looks good!

Thank you—fixed now.

jcameron:
Two more small things : 

1 - We shouldn't show "Virtual Server Summary" if "Edit Virtual Server" is also shown, because they basically show the same thing.

2 - Can the "Create Virtual Server" button at the top be red or something? This would be best done in the theme

iliajie:
> Two more small things :
> 
> 1 - We shouldn't show "Virtual Server Summary" if "Edit Virtual Server" is also shown, because they basically show the same thing.
> 
> 2 - Can the "Create Virtual Server" button at the top be red or something? This would be best done in the theme

Yeah, both are controlled by the theme and are unrelated to this patch.

1.	The "Virtual Server Summary" link has been shown for ages—users are accustomed to it, and it’s useful since it offers a quick overview of the virtual server, with details not available on the "Edit Virtual Server" page.

2.	As for the button, I can experiment with it further. It’s not yet clear how to improve its look and feel.

jcameron:
For the button color, I feel like it would be consistent with other web apps where the "create" or "compose" button is highlighted. Just a suggestion though ..

iliajie:
> For the button color, I feel like it would be consistent with other web apps where the "create" or "compose" button is highlighted. Just a suggestion though ..

I'll consider it! Thank you!

iliajie:
I have given it a fresh look, and while isolated screenshots look fine, they don't seem to work well when combined on the page (with right side). This is because red buttons usually convey a completely opposite, destructive meaning, such as delete or remove. Additionally, there are numerous other left menu palettes, and using red on them won't work either. Therefore, I believe the best solution is to use a highlight or contrast color, like this:

imageimageimage

PR #1029, opened 2025-03-18
Add the ability to differentiate WordPress clones

iliajie:
Hey Jamie!

This PR addresses the issue of correctly differentiating WordPress from its clones.

Details can be found here:

https://forum.virtualmin.com/t/wp-workbench-version-1-0-2-released/132611/6?u=ilia

### DIFF
--- a/scripts/classicpress.pl
+++ b/scripts/classicpress.pl
@@ -570,10 +570,8 @@ foreach my $wpconfig (@$files) {
 	my $wppath = $wpdir;
 	$wppath =~ s/^\Q$phd\E//;
 	$wppath ||= "/";
-	my @version = &find_recursive_files($wpdir, 'version.php');
-	next if (!@version);
 	my $cpok = 0;
-	foreach my $l (@{&read_file_lines("$wpdir/$version[0]", 1)}) {
+	foreach my $l (@{&read_file_lines("$wpdir/wp-includes/version.php", 1)}) {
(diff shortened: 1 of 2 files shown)


scripts/classicpress.pl
```diff
@@ -570,6 +570,16 @@ sub script_classicpress_detect
 	my $wppath = $wpdir;
 	$wppath =~ s/^\Q$phd\E//;
 	$wppath ||= "/";
+	my @version = &find_recursive_files($wpdir, 'version.php');
```
jcameron, line 573: Is this version.php potentially in some random directory under the wordpress root? If not, we should just use a fixed path relative to `wp-config.php` and  not call `find_recursive_files`

iliajie: That’s a reasonable suggestion! Right now, it’s in a specific directory, but future clones could change that. That said, I don’t see much benefit in making it static, so I’d rather keep it as is, if that works for you.

jcameron: No let's make it static. There's a good practice in programming called YAGNI which basically says, don't add code "just in case" unless you know it is going to do something useful. Because it introduces complexity and makes the reader wonder why we're checking for `version.php` in locations other than the one that it's always at.

iliajie: Alright, agreed! That sounds good to me. It's fixed now. Thank you!

jcameron: Excellent thanks for the fix!

PR #1049, opened 2025-05-06
Add custom branding for login page

iliajie:
Hey Jamie!

And this PR adds a change to actually create `/etc/webmin/brand.info` file in post-install.

### DIFF
--- a/postinstall.pl
+++ b/postinstall.pl
@@ -595,6 +595,26 @@ if ($gconfig{'forgot_pass'} && &foreign_check("virtualmin-password-recovery")) {
 				      "virtualmin-password-recovery");
 	}
 
+# Add custom branding for login page
+my $brand_info = "$config_directory/brand.info";
+my $brand_update = 1;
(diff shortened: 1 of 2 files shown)


postinstall.pl
```diff
@@ -595,6 +595,26 @@ sub module_install
...
+	&read_file($brand_info, \%brand_info);
+	$brand_update = 0 if (!$brand_info{'update'});
+	}
+if ($brand_update) {
+	my %minfo = &get_module_info($module_name);
+	my $title = $minfo{'shortdesc'} || 'Virtualmin';
+	&lock_file($brand_info);
+	my %brand;
+	$brand{'update'} = 1;
+	$brand{'icon'} = "fa2 fa2-virtualmin";
```
jcameron, line 612: Doesn't this have to be the path to an icon image file?

iliajie: For the `file` key, it should be, but for the `icon` key, it’s just a CSS class.

Issue #1050, opened 2025-05-07
New virtual servers lack remote MariaDB hostname, causing connection denial


iliaross:
Hey Jamie!

The bug seems to be with the current code is that when the remote MySQL/MariaDB is set up in “System Settings ⇾ Database Servers” and chosen as the default, any new virtual server created doesn’t have its hostname set in the “Edit Databases ⇾ Remote Hosts” textbox. This leaves the remote database server denying connections since it only lists “127.0.1.1” and “localhost,” which don’t make much sense for a remote server.

jcameron:
So I just tested this with the latest code, and it works fine! 

On the Virtualmin system, what is the hostname and what IP address does that hostname resolve to locally?

iliajie:
> On the Virtualmin system, what is the hostname and what IP address does that hostname resolve to locally?

Both have a matching record in `/etc/hosts` and can communicate just fine with each other. I have double checked, and the issue still persists, e.g.:

<img width="1075" alt="Image" src="https://github.com/user-attachments/assets/07353a76-3795-4101-a603-e63ed9934e9d" />

jcameron:
Did you try applying my patch above?

Also, can you post the contents of /etc/hosts?

iliajie:
> Did you try applying my patch above?

Yes, all patches had been applied.

> Also, can you post the contents of /etc/hosts?

Yes, sure.

```
10.211.55.3 host.rocky9-pro.virtualmin.dev
fdb2:2c26:f4e4:0:21c:42ff:fe52:75fc host.rocky9-pro.virtualmin.dev
```

and

```
10.211.55.20 host.debian12-pro.virtualmin.dev
fdb2:2c26:f4e4:0:21c:42ff:fea2:f273 host.debian12-pro.virtualmin.dev
```

jcameron:
Is the domain you're creating in this case a new top-level virtual server, or a sub-server? For sub-servers, granted hosts from the existing admin user will be re-used.

iliajie:
I was creating a top-level domain in those tests.

jcameron:
Try out this further patch : https://github.com/virtualmin/virtualmin-gpl/commit/b5cbfc7e182889206e10dc12808c44b6656b957e

iliajie:
This patch still doesn't add the hostname to the allowed remote hosts textbox, with only `127.0.1.1` and `localhost` being set.

[...]

PR #1051, opened 2025-05-10
Fix to correctly add remote host

iliaross:
Hey Jamie,

This PR address the issue we [discussed here](https://github.com/virtualmin/virtualmin-gpl/issues/1050#issuecomment-2869134363).


jcameron:
Thanks for this patch!

PR #1052, opened 2025-05-16
Add general script configure API

iliaross:
Hey Jamie!

This PR properly implements the script configure API, which was temporarily part of the WP Workbench plugin and has now [been removed](https://github.com/virtualmin/virtualmin-wp-workbench/commit/511efecbe874058ef82c3dee4c314cdd751d889a).

The only change in this version is that it dynamically loads the modules that support this feature, allowing each to have its own `configure` and `help` functions.


configure-script.pl
```diff
@@ -0,0 +1,82 @@
...
+my $script_usage_func = &script_find_kit_func($web_app_name, 'usage');
+my $script_cli        = &script_find_kit_func($web_app_name, 'cli');
+
+# Bail out if there’s no CLI handler
+if (!$script_cli) {
+	usage("Script '$web_app_name' does not support configure API");
+	}
+
+# Parse common command-line flags
+&parse_common_cli_flags(\@ARGV);
```
jcameron, line 64: Shouldn't this come before the loop over ARGV?

configure-scripts.pl
```diff
@@ -0,0 +1,82 @@
+#!/usr/local/bin/perl
+package virtual_server;
+
+use File::Basename;
+
+=head1 configure-scripts.pl
```
jcameron, line 6: How is this different from configure-script.pl ?

configure-script.pl
iliaross: No, because we first need to get the script type.

configure-scripts.pl
iliaross: They will call different underlying API. One is designed to work on one web app, while the other on multiple.

scripts-lib.pl
```diff
@@ -3759,5 +3759,18 @@ sub detect_installed_scripts
 return @rv;
 }
 
+# script_find_kit_func(name, suffix)
+# Returns a reference to a function in the script kit package if it exists
+sub script_find_kit_func
+{
+my ($name, $suffix) = @_;
+my $func = "script_${name}_kit_${suffix}";
+for my $pkg (grep { /::$/ } keys %main::) {
```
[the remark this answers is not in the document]
iliaross: Yes! We know the function name, but we don't know the package that provides it.

configure-scripts.pl
jcameron: I would recommend combing them, and allowing multiple apps to be specified via flags. This is what we do for existing Virtualmin API commands like `modify-web` which can take the `--domain` flag multiple times.

configure-script.pl
```diff
@@ -0,0 +1,82 @@
...
+		}
+	else {
+		chop($pwd = `pwd`);
+		}
+	$0 = "$pwd/configure-script.pl";
+	require './virtual-server-lib.pl';
+	}
+
+# Load all modules that can configure web app scripts
+my @modules = grep { $_->{'config_script'} } &get_all_module_infos();
```
jcameron, line 35: Wouldn't it be better here to iterate through just Virtualmin plugins?

configure-script.pl
jcameron: Why do we need the script type to parse common flags?

scripts-lib.pl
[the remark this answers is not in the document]
jcameron: But wouldn't it be one of the plugins? Better to just iterate over their packages, or better still just the ones that have `config_script` set.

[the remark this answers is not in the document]
iliaross: > But wouldn't it be one of the plugins? Better to just iterate over their packages, or better still just the ones that have `config_script` set.

Done https://github.com/virtualmin/virtualmin-gpl/pull/1052/commits/e2933c802a4441a481257194b8d808dd510d23b3.

configure-scripts.pl
iliaross: > I would recommend combing them, and allowing multiple apps to be specified via flags. This is what we do for existing Virtualmin API commands like `modify-web` which can take the `--domain` flag multiple times.

These are completely different than just taking `--domain` flag. But, okay, I agree, it's better to combine them. Fixed here https://github.com/virtualmin/virtualmin-gpl/pull/1052/commits/baad806afd05f0cb089bfccc8c8ddb717ee10c08.

Thanks!

configure-script.pl
iliaross: No, I think it's just fine as we are dealing with `module.info`.

jcameron: But is WP workbench a Virtualmin plugin? Everywhere else in Virtualmin where we call out to other modules, they have to be registered as plugins first..

scripts-lib.pl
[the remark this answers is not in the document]
jcameron: Ah yes, that's a lot cleaner

configure-script.pl
jcameron: What I mean is that in all other scripts, common args are parsed first. For example : https://github.com/virtualmin/virtualmin-gpl/blob/master/list-domains.pl#L78

configure-script.pl
iliaross: Oh, that's right! The fix is here https://github.com/virtualmin/virtualmin-gpl/pull/1052/commits/ecf9470766c115e2389852aca48635164c46f009.

configure-script.pl
iliaross: Yes, thanks—I know. Yet, I believe it's better to load it afterward for the reasons I mentioned above.

scripts-lib.pl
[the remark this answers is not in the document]
iliaross: Yes, thanks!

configure-script.pl
iliaross: Sure, fixed now!

PR #1062, opened 2025-06-18
Fix to factor out PHP-FPM start-stop function to work with other webservers

iliaross:
Hey Jamie!

As [described here](https://forum.virtualmin.com/t/php-fpm-isnt-included-in-server-status-dashboard/133738/11?u=ilia) this PR adds independent support for PHP-FPM monitors.

### DIFF
--- a/virtual-server-lib.pl
+++ b/virtual-server-lib.pl
@@ -109,6 +109,7 @@ foreach my $fname (@features, "virt", "virt6") {
 		     "directadmin" );
 @startstop_features = ("web", "dns", "mail", "ftp", "unix", "virus", "spam",
 		       "mysql", "postgres");
+@startstop_always_features = ("phpfpm");
 @bandwidth_features = ( @features, "backup", "restore" );
 @config_features = grep { $config{$_} } @features;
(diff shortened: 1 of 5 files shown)


virtual-server-lib.pl
```diff
@@ -109,6 +109,7 @@
 		     "directadmin" );
 @startstop_features = ("web", "dns", "mail", "ftp", "unix", "virus", "spam",
 		       "mysql", "postgres");
+@startstop_always_features = ("phpfpm");
```
jcameron, line 112: How do we know that FPM will always be available? The system might not have any webserver configured..

iliaross: > How do we know that FPM will always be available? The system might not have any webserver configured..

The `list_php_fpm_configs` function will handle this. And, we don’t need to worry about the system having or not having a webserver. Even though it’s rare, there’s nothing stopping me from having a PHP-FPM service running on one system and a webserver connecting to it remotely via a TCP socket. We shouldn’t tie those two services together.

I’m not sure if anyone would ever want to use that, though. But, these are two separate services, and we need to provide two separate independent statuses for both.

PR #1066, opened 2025-06-21
[ENH] Load TikiWiki version from sourceforge and check instance version before installation

bim-g:
## PHP version issue
While trying to install a new instance via the installer, we notice few things:
* You can select any version while being on an unsupported version of PHP, but the script installer will process with the installation of your instance,  then discover after that you should update your PHP version. 
* No possible way to know which version is supported before the installation process.

## Solution
So, the solution was to restrict before the installation to check the PHP version based on your instance you want to install.
knowing that from Tiki 26 they are running with PHP 8.1, under Tiki26 they are running with PHP 7.4.

## Project name and version are hard coded
We noticed that on the installation script the version and project name are hard coded, which will required to each released to update the script.
## Solution
To get the version and name from Source Forge via RSS instead of being hard coded, which will make the list dynamic each time there is new release.

### DIFF
--- a/scripts/tikiwiki.pl
+++ b/scripts/tikiwiki.pl
@@ -58,9 +58,28 @@ return ("mysql", "postgres");
 
 sub script_tikiwiki_php_fullver
 {
+use YAML::Tiny;
+use LWP::Simple;
+use Scalar::Util 'looks_like_number';
+
 local ($d, $ver, $sinfo) = @_;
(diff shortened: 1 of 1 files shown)


bim-g:
Thanks @iliaross for the update, I'm going to test it, and provide a feedback.

bim-g:
Made un update to load supported version from Tiki-Manager YAML config file, instead of hard coded.

PR #1070, opened 2025-06-25
[ENH] Fetch tiki project name from sourceforge via rss.

bim-g:
## Project name and version are hard coded
We noticed that on the installation script the version and project name are hard coded, which will required to each released to update the script.

## Solution
To get the version and name from Source Forge via RSS instead of being hard coded, which will make the list dynamic each time there is new release.

Related PR: #1066

### DIFF
--- a/scripts/tikiwiki.pl
+++ b/scripts/tikiwiki.pl
@@ -278,36 +278,34 @@ return $db_conn_desc;
 # a newer one. Otherwise returns undef.
 sub script_tikiwiki_check_latest
 {
+use LWP::Simple;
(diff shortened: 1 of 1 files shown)


jcameron:
We can't really safely do this, because a new version of tiki might change the install process. That's why we manually updated available versions of the installed, after qualifying each new version.

bim-g:
> What you could do is have Tiki update itself like WordPress does. Then Virtualmin could read the version and update its internal record dynamically, without touching the app itself.

Thanks for the update, let me take a look on it.

Issue #1068, opened 2025-06-24
Clearer terminology for currently used “alias target”


iliaross:
Hey Jamie!

As [discussed here](https://github.com/virtualmin/virtualmin-gpl/commit/bee2aad104eaa6d909cf66d6c7d86584fd59aa89#r160678766), I'm creating this issue to discuss how we can make the currently used "alias target" term less confusing.

My initial suggestion was:

> Can we please go through all the language and help files and make these "target" terms changed to something more user-friendly and unambiguous? You don’t have to do it yourself—let’s just agree on better naming, and I’ll help with the rest...

Later, I suggested using "primary" in this context, e.g.:

```
setup_edepmailalias=A target server must have mail for domain feature enabled if any aliases do
→
setup_edepmailalias=A primary virtual server must have mail for domain feature enabled if any aliases do
```

```
setup_webalias=Updating target virtual website ..
→
setup_webalias=Updating primary virtual server ..
```

```
setup_ewebalias=.. failed to find target virtual website!
→
setup_ewebalias=.. failed to find primary virtual server!
```

```
delete_apachealias=Removing from target virtual website ..
→
delete_apachealias=Removing alias from primary virtual server ..
```

```
save_apache5=Updating target virtual website ..
→
save_apache5=Updating primary virtual server ..
```

```
save_mailunalias1=Removing mail aliases copied from target domain ..
→
save_mailunalias1=Removing mail aliases copied from primary virtual server ..
```

```
cmass_ealias=Alias target server $1 does not exist
→
cmass_ealias=Alias primary virtual server $1 does not exist
```

```
spf_aliasdns0=Copy records from target domain
→
spf_aliasdns0=Copy records from primary virtual server
```

```
spf_noaliasdns=Switching alias domain to copy DNS records from target ..
→
spf_noaliasdns=Switching alias to copy DNS records from primary virtual server ..
```

```
unalias_rusure=Are you sure you want to convert this alias virtual server into a sub-server? It will no longer serve the web pages of the old target virtual server, but can have its own web content, mailboxes and mail aliases.
→
unalias_rusure=Are you sure you want to convert this alias into a sub-server? It will no longer serve the web pages of the old primary virtual server, but can have its own web content, mailboxes and mail aliases.
```

```
transfer_noalias=Alias target virtual server $1 does not exist on the destination system
→
transfer_noalias=Alias primary virtual server $1 does not exist on the destination system
```

[...]

PR #1071, opened 2025-06-26
Fix for to use clearer terminology for currently used “alias target” #1068

iliaross:
Hey Jamie,

As discussed [here](https://github.com/virtualmin/virtualmin-gpl/issues/1068).


lang/en
```diff
@@ -479,7 +479,7 @@ setup_edepmysql=A sub-server cannot have MySQL enabled unless the parent server
 setup_edepmysqlsub=A top-level server must have MySQL enabled if any sub-servers do
 setup_edeppostgres=A sub-server cannot have PostgreSQL enabled unless the parent server does
 setup_edeppostgressub=A top-level server must have PostgreSQL enabled if any sub-servers do
-setup_edepmailalias=A target server must have mail for domain feature enabled if any aliases do
+setup_edepmailalias=A origin virtual server must have mail for domain feature enabled if any aliases do
```
jcameron, line 482: Should be "An origin" , not "A origin"

PR #1074, opened 2025-07-04
Add ability to check the other SSL providers issuer

iliaross:
Hey Jamie!

As we discussed [here](https://github.com/virtualmin/virtualmin-gpl/commit/b1508fc4a9cb51ff629cca307cdb65546a5d36cb#commitcomment-161442283), this patch resolves the issue of auto-renewals not working for ZeroSSL and other SSL providers but LE.

I have also tested this patch—after uploading it to my production server, the ZeroSSL certificate was renewed within a few minutes.

The patch to make SSL providers' names available in a function is [here](https://github.com/virtualmin/virtualmin-pro/commit/c9646544aa34774262e80f0ca2aa2692fe3708a5).

### DIFF
--- a/feature-ssl.pl
+++ b/feature-ssl.pl
@@ -419,7 +419,7 @@ if ($d->{'dom'} ne $oldd->{'dom'} && &self_signed_cert($d) &&
 
 if ($d->{'dom'} ne $oldd->{'dom'} &&
     !$d->{'ssl_same'} &&
-    &is_letsencrypt_cert($d) &&
+    &is_acme_cert($d) &&
     !&check_domain_certificate($d->{'dom'}, $d)) {
 	# Domain name has changed ... re-request let's encrypt cert
 	&$first_print($text{'save_ssl12'});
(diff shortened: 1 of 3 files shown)


feature-ssl.pl
```diff
@@ -2961,8 +2961,18 @@ sub is_letsencrypt_cert
 	# Looks like a virtual server
 	$info = &cert_info($info);
 	}
-return $info && ($info->{'issuer_cn'} =~ /Let's\s+Encrypt/i ||
-		 $info->{'issuer_o'} =~ /Let's\s+Encrypt/i);
+
+# Check all provider names (if list fn exists)
+my @patterns = ("Let's\\s+Encrypt");
+if (defined &list_known_acme_providers) {
+	push(@patterns, map { quotemeta($_->{'name'}) }
```
jcameron, line 2968: I think we should add a field to the return value from `list_known_acme_providers` for the issuer regexp, because there's no guarantee that will be the same as the name we display in the UI. Like maybe ZeroSSL will rename the company next week, but keep using "ZeroSSL" in the issuer name.

jcameron: Oh I see I didn't realize you added that field! But maybe rename it to something that reflects that it is used for, like `issuer`.

Also, there is one case where `is_letsencrypt_cert` really does check for Let's Encrypt here : https://github.com/virtualmin/virtualmin-gpl/blob/6fde32043646f8bf7cb849dcc7e08d9c6c58f85c/feature-ssl.pl#L3233

Maybe instead of redefining what this function does, create a new one called `is_acme_cert` that checks for all the providers?

iliaross: ...and I'm not sure why you added CAA complexity as with no CAA records at all, every public CA is allowed to issue for the domain...

jcameron: Thanks for those fixes! The point of CAA records is to use DNS to limit which CA is allowed to issue certs for the domain, as a defence against MITM attacks.

iliaross: Sure I'd like, will do...

Issue #1002, opened 2025-01-24
Clean up the Virtualmin package


iliaross:
Hey Jamie, 

As discussed here https://github.com/virtualmin/virtualmin-gpl/issues/998#issuecomment-2609513070, we should not be shipping packages and mention options that have been obsolete for 15–20 years as it doesn’t leave a good impression, especially when trying to attract new users.

Therefore, I suggest we drop all code related to Ensim, LXAdmin, including removing `PHP/Serialization.pm`.

And, S3 modules can now be remove as well, as discussed in the same ticket here https://github.com/virtualmin/virtualmin-gpl/issues/998#issuecomment-2612339356.

jcameron:
I'm OK with dropping LXadmin and Ensim support since it looks like those control panels are long gone.

And also PHP/Serialization since it's only used by the LXadmin migration code. 

Please send me a PR for these!

iliajie:
Thanks! I will, in a bit!

shoulders:
Linking to this issue just incase it is of any use. Removal of these admin panels were requested here + others.

[#872 Remove Legacy migrate backup file types](https://github.com/virtualmin/virtualmin-gpl/issues/872)

iliaross:
> I'm OK with dropping LXadmin and Ensim support since it looks like those control panels are long gone.
> 
> And also PHP/Serialization since it's only used by the LXadmin migration code.
> 
> Please send me a PR for these!

Addressed here https://github.com/virtualmin/virtualmin-gpl/pull/1076.

PR #1076, opened 2025-07-07
Fix to drop obsolete restore systems and leave out only latest Plesk

iliaross:
Howdy Jamie!

This PR finally removes all the obsolete restore systems, as [discussed here](https://github.com/virtualmin/virtualmin-gpl/issues/1002#issuecomment-2614025542).

It also drops support for old Plesk 7 and 8 systems, which have been gone for decades. Now only Plesk, cPanel, and DirectAdmin remain—much clearer!

It also renames API `plesk9` to `plesk`, dropping Plesk 7 and 8 support entirely since it was released over 20 years ago, and the chances of anyone restoring that kind of backup are basically zero.


iliaross:
> but can you first check if anyone has filed a support ticket for migration of any of the control panel types removed here over the past couple of years?

I can’t remember a single support request for them! Ensim was discontinued back in 2007, LXadmin development stopped completely in 2013. Plesk 7 (from 2004) and Plesk 8 (from 2006) are both completely obsolete—Plesk is now at version 18+ under WebPros, the same parent as cPanel.

The ones I removed are decades old, some nearly two decades. I’m 99.9% sure no one will ever need to migrate from such outdated, unsupported systems.

We’re safe to drop them.

PR #1079, opened 2025-07-15
Replaced IDNA::Punycode with the actively maintained Net::LibIDN2 for full IDNA2008 support

iliaross:
Hey Jamie,

This PR removes the outdated and imperfect `IDNA::Punycode` and replaces it with the actively maintained `Net::LibIDN2`.

It also ensures that users will have it installed automatically upon next package upgrade.

Users who don't use RPM or DEB will need to manually install the widely available `Net::LibIDN2`. Plus, with the upcoming Virtualmin 8 release, this change is quite alright either way.


jcameron:
Can we not package Net::LibIDN2 with Virtualmin, like we did for the other module? I am worried that this change will break some users who don't have Net::LibIDN2 installed.

iliaross:
We can't package it, unfortunately, because it has `.so` file, which is system-dependent. However, the Virtualmin package upgrade will automatically get `Net::LibIDN2` from the OS repos...

jcameron:
No I mean, what will happen in the UI if the package is missing? Even though the package exists, we can't guarantee users will install it right away, or ever.

iliaross:
> Oh ok, that's fine then. As long as users aren't totally broken..

Yeah, it'll be just fine. Also, creating a new domain would fail cleanly with an error saying that domain names can only include letters, numbers, dots, and dashes.

However, now that I think about it, maybe we can show a clearer message suggesting to install `Net::LibIDN2`?

PR #1084, opened 2025-07-29
Fix to properly update IMAP user password

iliaross:
Hey Jamie,

This PR fixes a problem when neither domain owner nor mailbox user IMAP credentials weren't updated correctly.

For domain owner it takes a simple route and checks with submitted data.

For mailbox user it does what it should do, and always updates it because old user will never have `plainpass` set with hashing enabled.

I'm submitting PR instead of patching because I think you may want to discuss current implementation for domain owner.

### DIFF
--- a/api-create-domain-lib.pl
+++ b/api-create-domain-lib.pl
@@ -1008,6 +1008,14 @@ elsif ($sshmode == 2) {
 		}
 	}
 
+# Update the IMAP password
+my %u;
+$u{'user'} = $dom{'user'};
+$u{'home'} = $dom{'home'};
+$u{'plainpass'} = $pass;
+&set_usermin_imap_password(\%u);
+
+# Run post actions
 &run_post_actions_silently();
 &unlock_domain_name($domain);
 return \%dom;
--- a/create-domain.pl
+++ b/create-domain.pl
@@ -1086,6 +1086,14 @@ elsif ($sshmode == 2) {
 		}
 	}
 
+# Update the IMAP password
+my %u;
+$u{'user'} = $dom{'user'};
+$u{'home'} = $dom{'home'};
+$u{'plainpass'} = $pass;
+&set_usermin_imap_password(\%u);
+
+# Log the API call and run post actions
 &virtualmin_api_log(\@OLDARGV, \%dom, $dom{'hashpass'} ? [ "pass" ] : [ ]);
 &run_post_actions_silently();
 &unlock_domain_name($domain);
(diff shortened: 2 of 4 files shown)


iliaross:
Yet it still wouldn't work when calling `create-domain.pl`...

...but I'm really confused why we _micromanage_ the `plainpass` key in the hash references in so many different places, when we could _always_ just pass it around. Then, when `save_domain` or `save_user` are called, we could clean it up there. Or better, set a `$_->{'_tmp'}->{'plainpass'}`, pass it around with any other tmp variables (if needed), and clean it before saving to the disk. And, right now, it's hard to address it universally and requires quite a few changes here and there...

jcameron:
I don't follow what is going on here - what is the underlying problem or failure case that this PR is fixing?

iliaross:
> Hmm, I think I'd prefer we focus on another solution and just stop attempting to keep `inbox.imap` up to date in all these cases. As you pointed out, it's a bad idea to store plaintext passwords anyway.

Agreed. However, does it mean we should not set the mail system to IMAP in post-install? I don't want to keep shipping it broken. And, Usermin will still try to create `inbox.imap` anyway.

It feels like a much larger and more difficult change otherwise...

iliaross:
> The current state isn't that bad though, as Usermin users can still enter a login and password for IMAP the first time they open their mailbox.

It's bad because the delay before the IMAP login page appears in Usermin is so long that any regular user might think it's broken. And, yes, after they enter the password, even after waiting for no reason, the password will still be saved in the same `inbox.imap` file. This PR just fixes exactly this.

> We should fix it, but there's no need to rush out a partial solution.

It's not even a partial solution or a solution at all. It's just a bug fix for what we already have that wasn't working properly...

iliaross:
> Instead of spending time on a half-baked solution, we should simply leave things be for now, and work on a proper alternative after the next release.

Alright! I will close this then! Thanks!

> There are a lot of issues with this fix though that I haven't even started to get into :-)

May I know though which are those issues?

jcameron:
> May I know though which are those issues?

For starters, use of `$in` inside the `set_usermin_imap_password` function is bad practice.

iliaross:
> > May I know though which are those issues?
> 
> For starters, use of `$in` inside the `set_usermin_imap_password` function is bad practice.

Yeah, and this is something I pointed out right from the start—I meant essentially. Or, is this the only actual problem? Or, we should just completely remove this code?

jcameron:
So I've been working on adding support for using the `/usr/libexec/dovecot/imap` command instead of logging into a local Dovecot, and it seems like it will be pretty simple! Once that's done, we can do away with all this mess entirely!

jcameron:
FYI, support for a local IMAP server in Usermin has been implemented for the next release. This will remove the need to login with a username and password..

PR #1088, opened 2025-08-11
[FIX] Database postgress support: remove postgresql, is no longer supported. 

bim-g:
Post was used on older version of tiki, and they are no longer supported.
Based on [tiki documentation](https://tiki.org/TikiPostgres), postgress is no longer supported.

### DIFF
--- a/scripts/tikiwiki.pl
+++ b/scripts/tikiwiki.pl
@@ -46,14 +46,14 @@ sub script_tikiwiki_php_modules
 {
 local ($d, $ver, $phpver, $opts) = @_;
 local ($dbtype, $dbname) = split(/_/, $opts->{'db'}, 2);
-local @rv = $dbtype eq "mysql" ? ("mysql") : ("pgsql");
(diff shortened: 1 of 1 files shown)


scripts/tikiwiki.pl
```diff
@@ -46,14 +46,14 @@ sub script_tikiwiki_php_modules
 {
 local ($d, $ver, $phpver, $opts) = @_;
 local ($dbtype, $dbname) = split(/_/, $opts->{'db'}, 2);
-local @rv = $dbtype eq "mysql" ? ("mysql") : ("pgsql");
+local @rv = $dbtype eq "mysql" ? ("mysql") : return undef;
 push(@rv, "xml", "intl", "zip", "gd", "mbstring");
```
jcameron, line 50: You could just change these two lines to : 

```
my @rv = ("mysql", "xml", "intl", "zip", "gd", "mbstring");
```

bim-g: Got it, updated.

jcameron:
Thanks for the fix!

PR #1091, opened 2025-08-15
[ENH] Install Instance: add option to auto config the instance.

bim-g:
After the installation of the tiki instance, it required to proceed with instance configuration to finalize the installation, which can be time consuming because it will required to have database information and other sensitive information.
So, by adding the auto config this will finalize the instance configuration with default domain configuration and have an instance ready to used.


jcameron:
Thanks for this patch!

Issue #1106, opened 2025-09-12
Should the WP Workbench feature be disabled by default in backups?


iliaross:
Hey Jamie!

As discussed [here](https://forum.virtualmin.com/t/backup-issues-after-wp-workbench-2-0-0-release/134989/20?u=ilia), should we make WP Workbench to be disabled by default for include-all-features mode?

> I’m thinking now of making WP WB excluded by default if the option to back up all features is used. That way, the WP WB feature would always be excluded from all backups by default. It should be, as `dir` and `db` backups would overlap with it. So, for WP WB to work, it would need to be selected explicitly.

iliaross:
Agreed! I'll send you the patch or a PR for review, depending on how simple it is, after I finish with the theme fixes.

iliaross:
Implemented here https://github.com/virtualmin/virtualmin-gpl/pull/1111.

PR #1111, opened 2025-09-17
Add ability to exclude some backup features to avoid overhead

iliaross:
Hello Jamie!

As [discussed here](https://github.com/virtualmin/virtualmin-gpl/issues/1106#issuecomment-3289502635)—it's been tested and implemented.

Now if the scheduled backup has all features enabled, it will never back up a plugin that has `feature_backup_no_all_features` sub defined to avoid overhead.

On the UI, it will nicely and clearly show those features as grayed-out (disabled). Also, the backup logic will ignore that feature if all features mode is enabled.

I was thinking of just making a patch but made a PR in case you have some questions.

### DIFF
--- a/backup_form.cgi
+++ b/backup_form.cgi
@@ -163,18 +163,48 @@ foreach $f (&get_available_backup_features()) {
 		$ftable .= "<br>\n";
 		}
 	}
-foreach $f (&list_backup_plugins()) {
+my $hasbackupplugin = 0;
+foreach my $f (&list_backup_plugins()) {
+	my $dis = 0;
+	my $noall = 0;
+	if (&plugin_defined($f, "feature_backup_no_all_features") &&
+	    &plugin_call($f, "feature_backup_no_all_features")) {
+		$noall = $dis = 1;
+		$dis = 0 if (!$sched->{'feature_all'});
+		$hasbackupplugin++;
+		}
 	$ftable .= &ui_checkbox("feature", $f,
 		&plugin_call($f, "feature_backup_name") ||
 		    &plugin_call($f, "feature_name"),
 		&indexof($f, @schedfeats) >= 0,
-		"onClick='form.feature_all[1].checked = true'")."\n";
+		"onClick='form.feature_all[1].checked = true' ".
+		"data-feature-noall='$noall'", $dis)."\n";
 	if (&plugin_defined($f, "feature_backup_opts")) {
 		local %opts = map { split(/=/, $_) }
 				 split(/,/, $sched->{'backup_opts_'.$f});
-		$ftable .= &plugin_call($f, "feature_backup_opts", \%opts);
+		$ftable .= &plugin_call($f, "feature_backup_opts",
+					\%opts, $dis);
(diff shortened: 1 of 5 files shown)


backup_form.cgi
```diff
@@ -163,19 +163,49 @@ foreach $f (&get_available_backup_features()) {
 		$ftable .= "<br>\n";
 		}
 	}
-foreach $f (&list_backup_plugins()) {
+my $hasbackupplugin = 0;
+foreach my $f (&list_backup_plugins()) {
+	my $dis = 0;
+	my $noall = 0;
+	if (&plugin_defined($f, "feature_backup_no_all_features") &&
```
jcameron, line 170: Can't you use the function `prune_all_features_for_backup` here?

backups-lib.pl
```diff
@@ -7347,5 +7347,17 @@ sub can_trust_backup
...
+# prune_all_features_for_backup(&features)
+# Remove features whose plugin does not support backup for all features
+sub prune_all_features_for_backup
+{
+my ($features) = @_;
+my %rm = map { $_, 1 }
+	 grep { &plugin_defined($_, 'feature_backup_no_all_features') &&
+		&plugin_call($_, 'feature_backup_no_all_features')
+	 } &list_backup_plugins();
+@$features = grep { !$rm{$_} } @$features;
```
jcameron, line 7359: It would be much cleaner if this function just returned a new list rather than modifying the param

backup_form.cgi
iliaross: No, it'd be messy, and most importantly we're already inside `list_backup_plugins ` loop.

backups-lib.pl
iliaross: > It would be much cleaner if this function just returned a new list rather than modifying the param

No, not really. It wouldn't be cleaner in this case because the caller doesn't need a return value. However, what this function needs is a clearer comment—fixed now.

backup_form.cgi
```diff
@@ -163,19 +163,49 @@ foreach $f (&get_available_backup_features()) {
...
+		$ftable .= &plugin_call($f, "feature_backup_opts",
+					\%opts, $dis);
 		}
 	$ftable .= "<br>\n";
 	}
+if ($hasbackupplugin) {
+	print <<EOF;
+<script>
+(function() {
+	const radios  = document.querySelectorAll('input[name="feature_all"]'),
```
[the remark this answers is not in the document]
jcameron: I don't get why we need this? Can't we just make it so that if "all features" is selected, then the WP Workbench feature is not include. But if "Only selected" is chosen, users can optionally select it.

backups-lib.pl
iliaross: Alright, fixed on the latest patch.

backup_form.cgi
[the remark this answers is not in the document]
iliaross: We need this as to indicated the user that we are not backing "WP Workbench" in "Backup all features" mode.

PR #1114, opened 2025-09-19
Fix to set record type too, as transport alone not enough

iliaross:
Hey Jamie,

This is a small bug, but you wanted a PR, so here it is.


jcameron:
Thanks for the fix!

PR #1115, opened 2025-09-19
Fix to separate lookup URLs to correctly get external IPv4 and IPv6 addresses

iliaross:
Hello Jamie,

This fixes an issue we discussed in the chat, to clearly get the correct IPv4 and IPv6 using different URLs.

It should be merged when Joe sets this up.


jcameron:
Looks good to me! Let me know when this is ready to merge ...

PR #1118, opened 2025-09-21
Add the final touches to the work on DNS IPv6 support

iliaross:
Hey Jamie!

I finished the work you started last night to support DNS IPv6!

### DIFF
--- a/backups-lib.pl
+++ b/backups-lib.pl
@@ -2985,7 +2985,7 @@ if ($ok) {
 					       : &get_dns_ip($d->{'reseller'});
 			if (&supports_ip6()) {
 				$d->{'old_dns_ip6'} = $d->{'dns_ip6'};
-				$d->{'dns_ip6'} = 
+				$d->{'dns_ip6'} = $d->{'virt6'} ? undef :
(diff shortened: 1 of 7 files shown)


backups-lib.pl
```diff
@@ -2983,6 +2983,11 @@ sub restore_domains
 			$d->{'old_dns_ip'} = $d->{'dns_ip'};
 			$d->{'dns_ip'} = $virt ? undef
 					       : &get_dns_ip($d->{'reseller'});
+			if (&supports_ip6()) {
+				$d->{'old_dns_ip6'} = $d->{'dns_ip6'};
+				$d->{'dns_ip6'} =
```
jcameron, line 2988: Shouldn't there be a check for $virt6 here, like above?

iliaross: I really don't know what `$virt` is because it isn't defined anywhere... Should it be?

jcameron: Oh it should be $d->{'virt'} then

jcameron:
Thanks for this fix!

PR #1121, opened 2025-09-25
Add a pure Perl implementation for getting SSL certificate really fast

iliaross:
Hello, Jamie!

This patch fixes the major currently present bottleneck when getting SSL certificate info using the `openssl` command.

The performance improvement is not as great as the Webmin implementation https://github.com/webmin/webmin/pull/2558, but it's still incredibly faster, with an average speed gain 60 times faster. It's slower than the Webmin version because we need to parse additional fields in Virtualmin, modulus in particular, but it's still much faster than calling the `openssl` command by an order of magnitude.

It took a lot of time and effort to figure everything out and get all the parts working together correctly; and, I tested it on all my systems, and it worked really well for me.

Please let me know if you have any questions.


iliaross:
To be clear, I ran all these tests on powerful Apple Silicon M3 hardware using virtual machines. Running 1000 iterations took about 6 seconds when shelling out versus 0.1 seconds with pure Perl. That’s a big gap, but 6 seconds didn’t feel too bad.

However, on an older CentOS 7 box with a Celeron J4125, the difference was huge; the same 1000 iterations took 36 seconds when shelling out and just 0.4 seconds in pure Perl—over 80× faster.

Since not all users have high-end hardware, and those hosting many domains will notice this even more, the performance boost here is significant.

feature-ssl.pl
```diff
@@ -1016,9 +1016,139 @@ sub cert_data_info
...
+			: undef;
+$rv{algo} = $algo;
+return $shutdown->() if !$algo;	# unsupported algo
+
+# Key size
+my $size = Net::SSLeay::EVP_PKEY_bits($pkey);
+$rv{size} = "$size" if $size && $size > 0;
+
+# Modulus and exponent
+eval { require Crypt::OpenSSL::X509; };
```
jcameron, line 1097: Why not do this check earlier in the function? If the perl module is missing, there's no need to do all the previous work just to fail here.

iliaross: Agreed! Fixed now!

jcameron:
> Yes, some users have hundreds of domains, and some even have thousands. The overhead in this case is significant, especially when cloud instances are not well-performant.

Right, but when does some operation read them all at once?

I get that this new implementation is faster, which is great ... but it also adds extra code and complexity because we still need to maintain the old and new paths, and ensure they produce the same return value. So we should only do it when there is a clear benefit, like some operation running noticeably slow.

iliaross:
> Right, but when does some operation read them all at once?

Yes, almost everywhere it's called in a loop via `*cert_info` wrappers. With larger amount of domain it will be significantly improvement.

> So we should only do it when there is a clear benefit, like some operation running noticeably slow.

Earlier, you said, "I wish that was possible but I don't see a way"—I found a way, and now we're having this discussion, so I'm confused.

And, generally speaking, I think, it's always better to use pure Perl without process spawning and to use a well-supported module with good abstraction, so we don't have to worry about (any) command output changes.

iliaross:
I have ran more tests, and technically we can avoid using `Crypt::OpenSSL::X509` dependency, and it will even be another 30% faster on top of this one in this PR. However, the downside is that we will have to manage low-level stuff that is done by Perl modules ourselves.

In this case we can solely rely on `Net::SSLeay`.

iliaross:
> I ran more tests, and technically we can avoid using the `Crypt::OpenSSL::X509` dependency, and it will even be 30% faster on top of the version in this PR.

Ah, no, it will be 25-30% _slower than_ the version in this PR; but, it's still ~40 times faster than the call to `openssl`.

I still think `Crypt::OpenSSL::X509` is the best and fastest solution.

iliaross:
> If we can drop this dependency and rely only on Net::SSLeay

I think it's a mistake, but okay—check my latest change!

> since we know for sure that it will be available we can safely just delete the code that calls openssl

Sure, but not in the next few releases. We'll log it if there's an error and see if users report it.

jcameron:
> Sure, but not in the next few releases. We'll log it if there's an error and see if users report it.

But would we fall back to using `openssl` if there are any errors reading certs?

Also, I don't think that relying on users to report problems with the new code is the right approach. Instead we should do extensive testing on all types of certs and Linux distros to truly validate that the new code works.

feature-ssl.pl
```diff
@@ -1135,10 +1444,28 @@ sub cert_file_info
...
+# Returns a hash of details of a cert in some file using either Perl or OpenSSL
+sub cert_file_info
+{
+my ($file, $d) = @_;
+my $info;
+eval {
+	$info = &cert_file_info_perl($file);
+	};
+if ($@ || !$info) {
+	&error_stderr("SSL certificate parsing using Perl failed : $@");
```
jcameron, line 1460: Let's not sure users an error that's related to an internal failure. We should be confident enough in the new code that this isn't needed..

iliaross:
> Instead we should do extensive testing on all types of certs and Linux distros to truly validate that the new code works.

So be it! After extra hours of testing, I can confirm that **it works just fine**.

I dropped the old code!

For the record, tests were run with self-signed and CA LE/ZeroSSL EC and RSA certificates on:

CentOS 7
Rocky 8
Rocky 9
Debian 13
Ubuntu 24.04

Please try it on your end as well, and let me know if you find any issues.

jcameron:
Awesome thanks! I will run some tests ..

jcameron:
Well ... actually it didn't work, at least on my system. I get the error `Net::SSLeay::BIO_new_file`

jcameron:
Yes, from the CLI

PR #1123, opened 2025-09-28
A bug hunter example for Webmin modules

iliaross:
Hey Jamie,

A while ago, I created a function called `read_file_contents_limit` that can read a file partially. We can use it to find specific error messages in a log and display them on the dashboard in a dismissible alert message, asking the user to report them to us.

This is one way to test a new feature to make sure it works as expected for all users.

A default file to check for logged messages would be `miniserv.error` and each Webmin module would be able to configure its own sets of error messages to check.

But the main logic will be in Webmin.

### DIFF
--- a/virtual-server-lib-funcs.pl
+++ b/virtual-server-lib-funcs.pl
@@ -12583,6 +12583,17 @@ if ($config{'dns'}) {
 		}
 	}
 
+# Hunt message from a log file that need attention to show to the user
+if (defined &list_logs_messages) {
+	my $file;
+	my $log = { file => $file,      # if empty, check default miniserv.error
+		    msg  => [
+			'SSL certificate parsing using Perl failed',
+		    ] };
+	my @msgs = &list_logs_messages($log);
+	push(@rv, \@msgs);
+	}
+
 return @rv;
 }
 


iliaross:
For example as in this PR:

https://github.com/virtualmin/virtualmin-gpl/pull/1121/commits/c4523d6d1ca80c4e83264cd44c5551a1be66209b

We can't drop the old `openssl` shell-out implementation right now because we need to test it first. Even though it looks fine and tempting to do, some complex code should be phased out instead of removed instantly.

We would look for `SSL certificate parsing using Perl failed` in particular. There could, of course, be others.

So, this was the driving force for this idea.

jcameron:
Uh ... like I said elsewhere, I don't like that we are not confident in the new parsing code that we have to add a special-case error log check for failures.

But if we *really* need to do this, it should happen in validate_ssl ,  which is where these kinds of checks are already implemented.

iliaross:
> I don't like that we are not confident in the new parsing code that we have to add a special-case error log check for failures.

Well, I don't suggest keeping it forever. And, I'm pretty confident about the new code because I spent days on it. However, it's very complex and can't be simplified—you raised the bar to high, wanting super complex stuff done by us instead of just downloading Perl's `Crypt::OpenSSL::X509` from the repos. :--) 

So, in my opinion, we should have a fallback. It’s more important to avoid disrupting users’ systems than to worry about the small overhead of keeping temporary code. This can happen with any code, and since the old code has been in use for decades, it’s safe to keep it for a few more months before removing it.

jcameron:
Sure, the fallback is good! But I think that if we are going to add a check that certs can be parsed, it should happen in `validate_ssl`

iliaross:
> But I think that if we are going to add a check that certs can be parsed, it should happen in `validate_ssl`

We aren't really checking if the certificate can be parsed. We're checking if the new code works the same as the old `openssl` code. Later, when we're ready, we'll just remove `cert_file_info_openssl` and the check itself. There's no need to check it in `validate_ssl`.

Anyway, we can talk about it later after the new releases. It can wait.

jcameron:
OK, seems like maybe we don't need this because the fallback mode for parsing certs is never used?

iliaross:
Yeah, alright! Agreed!

PR #1180, opened 2026-02-12
Add a simplified UI for managing website redirects

iliaross:
Hello, Jamie!

This PR rewrites the website redirect form to be simpler for users, keeps existing redirect caps, and fixes several edge-case bugs in parsing and editing logic.

### Example
image

image

image

### Summary
- Replace the 4-option radio table for redirect destination with a single
  textbox that auto-detects the type (URL, path, directory alias, or
  host-preserving URL)
- Replace protocol checkboxes and sub-path radios with select dropdowns
- Add hostname picker from detected website hostnames with auto-detection
  of plain hostnames, wildcards, and regex patterns
- Add redirect status code column to the list view
- Add sub-path column to the list view
- Add IDN hostname support to redirect pages
- Fix non-destructive redirect modify (restore old redirect on save failure)
- Fix redirect ID generation for entries with special characters
- Fix webmail/admin redirect detection to avoid false positives

### DIFF
--- a/save_redirect.cgi
+++ b/save_redirect.cgi
@@ -231,15 +231,6 @@ else {
 				$r->{'host'} = $host;
 				$r->{'hostregexp'} = 0;
 				}
-			elsif ($host =~ /^[a-z0-9\.\_\-\*\?]+$/i) {
-				# Convert wildcard hostname (e.g. *.example.com)
-				# to a safe anchored regex
-				my $re = quotemeta($host);
-				$re =~ s/\\\*/.*/g;
-				$re =~ s/\\\?/./g;
-				$r->{'host'} = "^".$re."\$";
-				$r->{'hostregexp'} = 1;
-				}
 			else {
 				# Non-plain, no whitespace so treat as regex
 				# pattern
--- a/edit_redirect.cgi
+++ b/edit_redirect.cgi
@@ -102,7 +102,7 @@ if (&has_web_host_redirects($d)) {
 			"this.form.host_mode.length > $htexti) { ".
 			"this.form.host_mode[$htexti].checked = true; }\" ".
 		"placeholder=\"$text{'index_global_eg'} ".
-		"www.$d->{'dom'}\"");
+		"dom.tld or ^regex\$\"");
 	my @hmode_opts = ( [ 0, $text{'redirect_host_def'} ] );
 	if (@hosts) {
 		my @hopts = ( [ undef, $text{'redirect_host_pick'} ],
--- a/help/redirect_host.html
+++ b/help/redirect_host.html
@@ -10,15 +10,8 @@ Available options are:
 When using custom input:
 <ul>
 <li>Plain hostnames (like <tt>www.example.com</tt>) are matched exactly.</li>
-<li>Simple wildcard patterns using <tt>*</tt> and <tt>?</tt> (like
-<tt>*.example.com</tt> or <tt>mail?.example.com</tt>) are automatically
-converted to a regular expression.</li>
-<li>Other non-space patterns are treated as regular expressions.</li>
+<li>Custom patterns must be regular expressions (for example
+<tt>^mail[0-9]+\.example\.com$</tt>).</li>
 </ul>
-<p>
-Because wildcard patterns are stored internally as regular expressions, re-opening
-the redirect may show the equivalent regular expression instead of the original
-wildcard text.
-</p>
 
 <footer>


help/redirect_dest.html
```diff
@@ -1,19 +1,19 @@
...
+<li><b>Full URL</b> (for example <tt>https://other.example/path</tt>)<br>
+    Redirect to another website.
+<li><b>URL path</b> (for example <tt>/new-path</tt>)<br>
+    Redirect to another path on this website.
+<li><b>Directory path</b> (for example
+    <tt>/home/user/public_html/files/</tt>)<br>
+    If this is an existing directory under the website root, it is used as an
+    alias target.
+<li><b>Host-preserving URL</b> (for example
+    <tt>http://%{HTTP_HOST}/path</tt>)<br>
```
shoulders: maybe make the `http` into `https` or the line above into `http`

I did not know if this arrangement was intentional especially as the default nowadays is to use https

save_redirect.cgi
```diff
@@ -22,6 +22,85 @@ if ($in{'delete'}) {
...
+			$in{'dest'} =~ /^\Q$d->{'home'}\E\//)) {
+			$looks_dir = 1;
+			}
+		elsif ($rroot && $rroot ne "/" &&
+			($in{'dest'} eq $rroot ||
+			$in{'dest'} =~ /^\Q$rroot\E\//)) {
+			$looks_dir = 1;
+			}
+		# Treat as directory alias only if it really exists on disk
+		if ($actualdir && -d $actualdir &&
```
jcameron: This seems unsafe, because if the user deletes a directory it would cause Virtualmin to no longer see it as an alias.

list_redirects.cgi
```diff
@@ -21,23 +26,42 @@ foreach $r (@redirects) {
...
+			}
+		}
 	my $iswebmail = &is_webmail_redirect($d, $r);
 	my $iswww = &is_www_redirect($d, $r);
 	my $canedit = !$iswebmail && !$iswww;
+	my $code = $r->{'alias'} ? "&nbsp;&nbsp;-" : ($r->{'code'} || 302);
+	my $host = $r->{'host'};
+	my $host_disp = $host || $text{'redirects_any'};
+	if ($host && !$r->{'hostregexp'} &&
+	    $host !~ /[%\$\{\}\[\]\(\)\^\*\?\+\|\\]/) {
```
jcameron: What is this regexp for?

redirects-lib.pl
```diff
@@ -558,8 +568,15 @@ sub is_webmail_redirect
 {
 my ($d, $r) = @_;
 return 0 if (!$r->{'host'});
-return 1 if ($r->{'host'} =~ /^admin\.\Q$d->{'dom'}\E$/);
-return 2 if ($r->{'host'} =~ /^webmail\.\Q$d->{'dom'}\E$/);
+return 0 if ($r->{'path'} ne '/');
+my ($dhost, $dport) = lc($r->{'dest'} || '') =~
```
jcameron: Why not use the standard function parse_http_url here?

save_redirect.cgi
iliaross: ~Agreed. Fixed now!~

save_redirect.cgi
```diff
@@ -22,6 +22,85 @@ if ($in{'delete'}) {
...
+		$in{'dpath'} = $2 || "/";
+		}
+	elsif ($in{'dest'} =~ /^(http|https):\/\/\S+$/) {
+		$in{'mode'} = 0;
+		}
+	elsif ($in{'dest'} =~ /^\/\S*$/) {
+		my $rroot = &get_redirect_root($d);
+		my $actualdir = $in{'dest'};
+		if ($actualdir =~ s/\$.*$//) {
+			# If path contains $1, reduce to parent dir
```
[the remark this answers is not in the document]
iliaross: Agreed. The comment is fixed now!

save_redirect.cgi
iliaross: Not really unsafe as this code runs only when saving, not when listing.

list_redirects.cgi
```diff
@@ -21,23 +26,42 @@ foreach $r (@redirects) {
 	    $dest =~ /^(http|https):\/\/%\{HTTP_HOST\}(\/.*)$/) {
 		$dest = &text('redirects_with', "$2", uc($1));
 		}
+	elsif (!$r->{'alias'} && $dest =~ /^https?:\/\//) {
+		my ($phost) = &parse_http_url($dest);
+		if ($phost) {
+			my $dhost = &show_domain_name($phost, 2);
```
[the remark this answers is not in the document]
jcameron: So does this mean that the UI shows human-friendly domain names even in the redirect destination?

redirects-lib.pl
jcameron: Why only match redirects with an explicit port though?

list_redirects.cgi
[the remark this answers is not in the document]
iliaross: Redirect destination? Do you mean the webserver config? In the webserver config, the redirect is in `xn--` format.

list_redirects.cgi
iliaross: Yes, it would! Agreed! Fixed now!

save_redirect.cgi
```diff
@@ -72,40 +158,95 @@ else {
...
+					&error($text{'redirect_ehost'});
+				}
+			if ($host =~ /^[a-z0-9\.\_\-]+$/i) {
+				# Plain hostname with exact host match
+				$r->{'host'} = $host;
+				$r->{'hostregexp'} = 0;
+				}
+			elsif ($host =~ /^[a-z0-9\.\_\-\*\?]+$/i) {
+				# Shell-like wildcard hostname from ServerAlias
+				# is converted to a safe anchored regex
```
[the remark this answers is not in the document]
jcameron: Could we maybe simplify this to just support the pattern `*.domain.com` ? I feel like there's no safe way to otherwise convert arbitrary patterns like `example*something.com` to regexps,  because this is already a valid regexp.

save_redirect.cgi
jcameron: What I mean is, if a user creates an alias to a directory, then deletes that directory, then re-edits and saves the alias without touching anything in the UI, it should not change to a redirect.

redirects-lib.pl
jcameron: So are you checking here for the port being explicitly set to 10000 or 20000 in the redirect, as an indicator that it's a redirect to Webmin or Usermin? If so, I would strongly recommend using the standard `parse_http_url` function so we centralize URL parsing in one place. Imagine if in 5 years a new URL format like `httpx://foo.com` comes along, and this code is still only handling the `http` and `https` formats!

iliaross: 🙂 Sorry, I probably wasn't clear enough before. This code is not checking for any ports specifically.

It only checks that the destination URL contains an explicitly written port in the original string.

I think we need that because `parse_http_url()` normalizes missing ports to defaults, i.e. `80` and `443`, so it loses the distinction between `https://host/` and `https://host:443/`, which is the guard that avoids false positives here.

And, we intentionally only match `http` or `https` here because `is_webmail_redirect()` is not a generic URL parser.

jcameron: But why is port normalization an issue if Webmin and Usermin are on ports 10000 and 20000 ?

iliaross: Yeah, but can't Apache listen on ports 8080 and 8443? I know it's unlikely, but still.

If you really want to use `parse_http_url ` func, then this function should have a flag to avoid normalizing missing ports 80 and 443.

jcameron: I would very strongly prefer to not implement URL parsing in multiple places just to handle this rare case. It's just bad coding practice to repeat logic.

PR #1213, opened 2026-05-08
Fix to retire ProFTPD virtual FTP feature

iliaross:
Hello, Jamie!

As you asked, creating a PR for this change so we can discuss it.

> In particular, this change breaks other FTP-related features such as chroot directories, which we do want to preserve.

Thanks, I’ll take a look soon! I just want to finish something first.


iliaross:
> > In particular, this change breaks other FTP-related features such as chroot directories, which we do want to preserve.
> 
> Thanks, I’ll take a look soon! I just want to finish something first.

Jamie, please see this patch https://github.com/virtualmin/virtualmin-gpl/pull/1213/commits/0e336da3a7be0da4be65926dfe5df90761de7219.

copycert-lib.pl
```diff
@@ -702,9 +680,14 @@ sub copy_proftpd_ssl_service
 my ($d) = @_;
 
 # Get the ProFTPd config and cert files
+&has_proftpd_support() || &error($text{'copycert_eproftpd'});
 &foreign_require("proftpd");
 &proftpd::lock_proftpd_files();
-my $conf = &proftpd::get_config();
+my $conf = eval { &proftpd::get_config() };
```
jcameron: Why is this eval needed?

backup-domain.pl
```diff
@@ -116,6 +116,10 @@ package virtual_server;
 		}
 	elsif ($a eq "--feature") {
 		local $f = shift(@ARGV);
+		if (&indexof($f, @retired_features) >= 0) {
```
jcameron: What is this retired_features array for? Is the goal to still support  running`backup-domain` with the `--feature ftp` flag even though it's no longer supported?

copycert-lib.pl
iliaross: Because calling it is not a harmless check, it can fail if the Webmin ProFTPd module is installed and configured but the actual ProFTPd config is missing or broken.

This helps prevent users from breaking their systems and then asking us questions how to fix them.

However, I admit it's an overly defensive copy-paste from `has_proftpd_support`, which is the only place that really needs this check. I’ve made a patch https://github.com/virtualmin/virtualmin-gpl/pull/1213/commits/94fb04c598e3d42e967b5e6d18156774c32f042e to clean it up.

backup-domain.pl
iliaross: > What is this retired_features array for? Is the goal to still support running`backup-domain` with the `--feature ftp` flag even though it's no longer supported?

Yes, that's right! It’s just a compatibility list that used to exist but should no longer function like standard Virtualmin features.

jcameron: Can we just drop it? If some user is calling `backup-domain` with the `ftp` flag for a feature that doesn't exist, we should generate an error instead of silently proceeding. Then they can fix their scripts.

copycert-lib.pl
jcameron: Excellent thanks!

backup-domain.pl
iliaross: Well, okay! Fixed here on the latest commit.

jcameron:
So is the goal here to remove FTP support entirely, or just the virtual FTP feature? Because I see PR is deleting things like the ability to stop and start the FTP server, which I think we should keep. Also support for chroot'd FTP.

iliaross:
> Because I see PR is deleting things like the ability to stop and start the FTP server, which I think we should keep. Also support for chroot'd FTP.

You're right! Sorry about that! The fix is here https://github.com/virtualmin/virtualmin-gpl/pull/1213/commits/e79471fd0651674f76eef26a77d5ce438541a84c.

iliaross:
> > Because I see PR is deleting things like the ability to stop and start the FTP server, which I think we should keep. Also support for chroot'd FTP.
> 
> You're right! Sorry about that! The fix is here [e79471f](https://github.com/virtualmin/virtualmin-gpl/commit/e79471fd0651674f76eef26a77d5ce438541a84c).

Jamie, code review is wrong, right?

> The virtual FTP removal mostly updates feature lists and callers, but the now-unconditional chroot backup entry can break Virtualmin backups when ProFTPd is not available.

backup-domain.pl
```diff
@@ -116,7 +116,8 @@ package virtual_server;
 		}
 	elsif ($a eq "--feature") {
 		local $f = shift(@ARGV);
-		$f eq "virtualmin" || $config{$f} ||
+		$f eq "virtualmin" ||
+		   (&indexof($f, @features) >= 0 && $config{$f}) ||
```
jcameron: We could remove the need for this extra check of `@features` by just removing `ftp` from the module config in `postinstall.pl`

jcameron:
Yeah I don't get what that code review is thinking, sorry. The change seems pretty much OK to me, except for one comment I added.

backup-domain.pl
iliaross: > We could remove the need for this extra check of @features by just removing ftp from the module config in postinstall.pl

Probably better to keep the check here so we don’t break anything on existing systems?

iliaross:
> Yeah I don't get what that code review is thinking, sorry. The change seems pretty much OK to me, except for one comment I added.

There was a bug back then in the code review code for PRs that should be fixed now. Thanks.

backup-domain.pl
jcameron: But that wouldn't be possible if we removed `ftp=1` from `/etc/webmin/virtual-server/config` at upgrade time.

jcameron: Or alternately, create a common function like `is_enabled_feature` and move the current logic there.

iliaross: > Or alternately, create a common function like `is_enabled_feature` and move the current logic there.

That works find for me. Fixed!

iliaross:
> > Because I see PR is deleting things like the ability to stop and start the FTP server, which I think we should keep. Also support for chroot'd FTP.
> 
> You're right! Sorry about that! The fix is here [e79471f](https://github.com/virtualmin/virtualmin-gpl/commit/e79471fd0651674f76eef26a77d5ce438541a84c).

And, let me check more in this regard...

iliaross:
> > > Because I see PR is deleting things like the ability to stop and start the FTP server, which I think we should keep. Also support for chroot'd FTP.
> > 
> > 
> > You're right! Sorry about that! The fix is here [e79471f](https://github.com/virtualmin/virtualmin-gpl/commit/e79471fd0651674f76eef26a77d5ce438541a84c).
> 
> And, let me check more in this regard...

Alright! I think we’re good to merge!

backup-domain.pl
```diff
@@ -116,8 +116,7 @@ package virtual_server;
 		}
 	elsif ($a eq "--feature") {
 		local $f = shift(@ARGV);
-		$f eq "virtualmin" || $config{$f} ||
-		   &indexof($f, &list_backup_plugins()) >= 0 ||
+		&is_enabled_feature($f, &list_backup_plugins()) ||
```
jcameron: This is an odd way of defining the function ... instead of passing in `list_backup_plugins`, why not just create a function called `is_backup_feature` ?

iliaross: > This is an odd way of defining the function

I'm really confused! What exactly are we deleting here?

iliaross: > > This is an odd way of defining the function
> 
> I'm really confused! What exactly are we deleting here?

Oh, "defining" the function! Sorry, I misread this initially.

And to answer your question, I was thinking of `is_enabled_feature()` as a general helper that could accept different plugin sets, but I see your point that we only have the backup use case right now.

I kept the generic helper, but added `is_enabled_backup_feature()` so the backup and restore callers no longer need to pass `list_backup_plugins()` directly.

See this patch https://github.com/virtualmin/virtualmin-gpl/pull/1213/commits/9855d1a9c6d46e3304c98b6bc6cc3d5398252048.

jcameron: Awesome thanks!

PR #1214, opened 2026-05-11
Fix Dovecot `local_name` insertion order

iliaross:
Hi, Jamie!

As [discussed earlier](https://github.com/virtualmin/virtualmin-gpl/commit/94211df1411d8310b6864d4914cf9aed7c64e741#commitcomment-184911306), this PR simply places newly created Dovecot SSL SNI `local_name` blocks before overlapping exact host blocks when needed, without reordering existing configs.

### DIFF
--- a/functional-test.pl
+++ b/functional-test.pl
@@ -7865,20 +7865,6 @@ $sslserv_tests = [
 		      [ 'service', 'dovecot' ] ],
 	},
 
-	# Check that parent wildcard is before the child host
-	{ 'command' => 'cat '.&dovecot::get_config_file(),
-	  'grep' => [ 'local_name \\*\\.'.$test_domain,
-		      'local_name '.$test_ssl_subdomain,
-		    ],
-	},
-	{ 'command' => "perl -ne 'if (/local_name \\\\Q*.$test_domain\\\\E\\\\b/) ".
-		       "{ \$w = \$.; } ".
-		       "if (/local_name \\\\Q$test_ssl_subdomain\\\\E\\\\b/) ".
-		       "{ \$s = \$.; } ".
-		       "END { exit(!(\$w && \$s && \$w < \$s)); }' ".
-		       &dovecot::get_config_file(),
-	},
-
 	# Validate that child Dovecot SNI prefers the specific host cert
 	{ 'command' => 'test-imap.pl',
 	  'args' => [ [ 'user', $test_ssl_subdomain_user ],
--- a/functional-test.pl
+++ b/functional-test.pl
@@ -7601,13 +7601,13 @@ $sslserv_tests = [
 		      @create_args, ],
         },
 
-	# Create an alias as well
+	# Create an alias with mail and SSL as well
 	{ 'command' => 'create-domain.pl',
 	  'args' => [ [ 'domain', $test_subdomain ],
 		      [ 'alias', $test_domain ],
 		      [ 'prefix', 'example2' ],
 		      [ 'desc', 'Test alias-domain' ],
-		      [ 'dir' ], [ $web ], [ 'dns' ], [ 'mail' ],
+		      [ 'dir' ], [ $web ], [ 'dns' ], [ 'mail' ], [ $ssl ],
 		      @create_args, ],
 	},
 
@@ -7746,6 +7746,8 @@ $sslserv_tests = [
 	{ 'command' => 'cat '.&dovecot::get_config_file(),
 	  'grep' => [ 'local_name '.$test_domain,
 		      'local_name \\*\\.'.$test_domain,
+		      'local_name '.$test_subdomain,
+		      'local_name \\*\\.'.$test_subdomain,
 		    ],
 	},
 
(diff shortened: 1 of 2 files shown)


functional-test.pl
```diff
@@ -7835,6 +7837,68 @@ package virtual_server;
...
+		      [ 'service', 'dovecot' ] ],
+	},
+
+	# Check that parent wildcard is before the child host
+	{ 'command' => 'cat '.&dovecot::get_config_file(),
+	  'grep' => [ 'local_name \\*\\.'.$test_domain,
+		      'local_name '.$test_ssl_subdomain,
+		    ],
+	},
+	{ 'command' => "perl -ne 'if (/local_name \\\\Q*.$test_domain\\\\E\\\\b/) ".
```
jcameron: What is this testing for exactly? I think in general it's best to just check for the behavior we want from the wget or openssl commands, rather than the format of the config file which can be brittle.

jcameron:
Thanks for writing this! It mostly looks good, except for one comment I had about the test..

functional-test.pl
iliaross: Ah, okay! I agree, it's brittle—I dropped it. Thanks!

jcameron:
Wait are you sure this works? I just tested with `virtualmin functional-test --test sslserv` and it ended up putting ssl.example.com after example.com in the Dovecot config.

BTW, I strongly recommend using the functional tests wherever possible to test changes, and expanding them as needed.

iliaross:
> Wait are you sure this works?

I'm quite sure it works because I tested it manually.

> I just tested with virtualmin functional-test --test sslserv and it ended up putting ssl.example.com after example.com in the Dovecot config.

Perfect! Indeed this is the order we need. Anything else would be wrong, since `*.example.com` also matches `ssl.example.com`, and Dovecot uses the last matching `local_name` block.

> I strongly recommend using the functional tests wherever possible to test changes, and expanding them as needed.

I manually test all code before submitting!

And, I realize functional tests are better than none and can catch bugs in future changes, but I don’t like relying on them alone too much speaking generally. Yet, for the Virtualmin CLI, they’re almost always work fine.

iliaross:
I've also observed over decades how industry tooling lags behind. Linters and CI pipelines update years after better techniques exist. That said, LLMs genuinely shine at catching context-dependent typos that a linter can't reason about, like `text{'label'}` vs `$text{'label'}`, and they're much simpler to configure and set up.

Thinking about it further, I'll actually add a review step on push/PR diffs and make CI fail if Claude flags medium- or high-severity issues. It should only take a few lines of code and won't touch the existing codebase.

If we want better, more secure code, we should **improve it that way** going forward, as reshaping old but perfectly working code just to satisfy static or functional tests gets it the other way around.

iliaross:
Alright, heads-up! Webmin CI/CD now supports proper automated Claude Code review for all pushes. If something’s wrong, it fails the job, and we get an email from GitHub with an annotation icon image. When clicked, it opens the failed job runner with more details on why it failed, for example:

image

Or, on the inside of the job:

image

For testing, I started with the private Virtualmin Podman repo, so you can go to its [GitHub Actions section](https://github.com/virtualmin/virtualmin-podman/actions) and check the latest `Dev Review` jobs to see how I made it fail at first and then how it was fixed.

It’s a pretty solid code review system and helps prevent silly mistakes that could make it into a release!

I'm enabling code review for all our other buildable repos. I hope you like it!

jcameron:
+1 to using linters!

Sorry I guess I misunderstood the way Dovecot works, I assume that the *first* match would win. Let's take another look then and see why this doesn't work for me.

jcameron:
> Sorry I guess I misunderstood the way Dovecot works, I assume that the first match would win. Let's take another look then and see why this doesn't work for me.

@iliaross I just tested this on my system with Dovecot 2.2.36, and the first `local_name` match is the one that that wins! Are you sure this wasn't changed in some later Dovecot version?

iliaross:
> I just tested this on my system with Dovecot 2.2.36, and the first local_name match is the one that that wins! Are you sure this wasn't changed in some later Dovecot version?

Yeah, this is what’s stated in the [Dovecot docs](https://doc.dovecot.org/main/core/config/ssl.html#with-client-tls-sni-server-name-indication-support):

> Also note that the ordering of [local_name](https://doc.dovecot.org/main/core/settings/syntax.html#connection-filters) is important. The wildcard certificate must be listed first, followed by more specific names. Note that the last matching [local_name](https://doc.dovecot.org/main/core/settings/syntax.html#connection-filters) filter will be used.

However, I looked into it further and, to my surprise, found that on CentOS 7 with Dovecot 2.2.36, it works as you described—the first matching `local_name` takes precedence when overlapping blocks use the same SSL keys.

I’d say we should just let CentOS 7 go. It’s a 12-year-old distro, and no one should rely on it for security.

jcameron:
Well it's not so much about the CentOS release as the Dovecot version. Which Dovecot were you testing with? We could easily add a version check to handle both cases.

iliaross:
All versions I tested were Dovecot 2.3.x. And, I doubt we need to worry about security issues in Dovecot 2.2.x, as those versions have been end-of-life for years.

PR #1216, opened 2026-05-13
Prep for testing, add compile.t

swelljoe:
Same prep for testing as [this PR](https://github.com/webmin/webmin/pull/2701) does for Webmin.

This one is hairier, as we have `execute_webmin_function` in `json-lib.pl` which does this:

```
	eval "
		\%pkg::ENV = \%ENV;
		package $pkg;
		do \$cmd;
		die \$@ if (\$@);
		";
	exit($@ ? 1 : 0);
```

To run scripts, which sets `caller` and also expects side effects. I can't figure out why it does it this way, but it's easy/safe to work around with an environment variable `VIRTUALMIN_NO_MAIN` that gets set in tests for the handful of effected files, which is checked instead of `caller`.

Also adds a `compile.t` test which just does `perl -c` across all Perl files, and a `t/README.md` documenting how to make tests and the special case we're doing for `execute_webmin_command`.

### DIFF
--- a/backup.pl
+++ b/backup.pl
@@ -2,6 +2,9 @@
 # Do a scheduled virtual server backup
 
 package virtual_server;
+
+unless (caller) {
+
 $main::no_acl_check++;
 require './virtual-server-lib.pl';
 $host = &get_system_hostname();
@@ -337,6 +340,8 @@ if ($sched->{'email_doms'} && $has_mailboxes &&
 	    { 'doms' => [ map { $_->{'dom'} } @doms ],
 	      'failed' => !$ok,
 	      'sched' => $id, });
+
+} # end of unless (caller)
 
 # Override print functions to capture output
 sub first_save_print
--- a/backup.pl
+++ b/backup.pl
@@ -3,7 +3,7 @@
 
 package virtual_server;
 
-unless (caller) {
+unless ($ENV{VIRTUALMIN_NO_MAIN}) {
 
 $main::no_acl_check++;
 require './virtual-server-lib.pl';
@@ -341,7 +341,7 @@ if ($sched->{'email_doms'} && $has_mailboxes &&
 	      'failed' => !$ok,
 	      'sched' => $id, });
 
-} # end of unless (caller)
+} # end of unless ($ENV{VIRTUALMIN_NO_MAIN})
 
 # Override print functions to capture output
 sub first_save_print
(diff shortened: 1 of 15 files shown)


iliaross:
Hey Joe!

I’m sorry, but I’m not comfortable with the direction of this PR, as well as https://github.com/webmin/webmin/pull/2701 and https://github.com/webmin/webmin/pull/2695, as-is.

I generally support adding tests, especially broad compile smoke tests, but changing production script flow primarily so files can be loaded by tests feels like **the wrong order of operations**. The production-code changes seem to be preparing for future tests rather than supporting tests added here.

I’d prefer we first improve or extract the code into cleaner, testable blocks, then add tests against those blocks.

On top of that, my concern is that these small guard tweaks can subtly change behavior or scope in old Perl code and **give us false confidence**. A compile test can pass while runtime behavior changes, especially around file-scope lexicals, package globals, and scripts loaded via `do` or `require`.

The compile test itself is valuable, but I think it should be separated from production refactors unless those refactors are independently justified and covered by tests that exercise the changed behavior.

For now, I’d _strongly_ prefer we put this direction on hold until we do a broader cleanup of the code to fit proper modern Perl standards. I can do a proper broader cleanup after I finish working on Cloudmin 10, sometime this summer.

swelljoe:
You can't safely do a large refactor without a test suite.

swelljoe:
Fine, I'll change them all to use the variable method.

You cannot argue the variable method can have unintended effects. It doesn't rely on caller, at all.

I'm really not comfortable with embarking on any kind of major changes without a really good test suite. You act like you're afraid of breaking things, and then you want to do the thing that's guaranteed to break a lot of things before we have a good way to check for breakage. That doesn't make sense.

jcameron:
Is this PR ready for review? Sorry I missed it initially ..

iliaross:
Joe, this was a good PR and the compile-test idea is worth keeping, which we'll do!

Since this branch is under `swelljoe`, I can’t update the PR branch directly, so I’ll close this version. As we discussed, the compile-test idea is good, but this implementation changes the actual script entry points by wrapping their main bodies in guard logic. That solves the `require` problem for tests, but it also risks changing how these scripts behave when run normally by Webmin, cron, or CGI.

The reworked version https://github.com/virtualmin/virtualmin-gpl/commit/bdb7643 that I just added keeps the very useful part of your original PR, like checking that all `.pl` and `.cgi` files still compile. The difference is that it does this with `perl -c`, so we do not need to wrap or modify the actual Virtualmin scripts just to make them testable.

For deeper unit tests, we should always move reusable helper logic into library files and test that directly.

Thanks!!

PR #1222, opened 2026-05-26
Fix milter-greylist mail rate-limit bypasses

iliaross:
Hello, Jamie!

This PR normalizes Virtualmin milter-greylist rate-limit configs so quota ACLs run before broad whitelists like `my network`, and ensure noauth/nospf are present so authenticated and SPF-pass senders are still metered.

And, also fixes the Postfix chroot detection typo (`=` instead of `eq`), guard `get_mailserver_chroot` against an undef `$mail_system` when called from `postinstall.pl`, and flush the lines cache after the socket swap on the backup restore path.

https://forum.virtualmin.com/t/tutorial-fix-mail-rate-limiting-not-working-on-ubuntu-24-04-postfix-chroot-milter-greylist-config-order/137303?u=ilia

### DIFF
--- a/ratelimit-lib.pl
+++ b/ratelimit-lib.pl
@@ -393,6 +393,8 @@ return 1;
 sub enable_ratelimit
 {
 # Build stop and start commands (needed when installed from source)
+my $cfile = &get_ratelimit_config_file();
+&lock_file($cfile);
 my $conf = &get_ratelimit_config();
 my ($pidfile) = grep { $_->{'name'} eq 'pidfile' } @$conf;
 if (!$pidfile) {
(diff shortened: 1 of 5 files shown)


iliaross:
I think the latest review is wrong, unlike the first one. I think we're good.

ratelimit-lib.pl
```diff
@@ -208,6 +208,94 @@ sub save_ratelimit_directive
...
+		&save_ratelimit_directive($conf, undef, $new,
+					  $control_before);
+		$changed++;
+		}
+	}
+
+# Keep rate-limit ACLs before ordinary whitelists such as "my network",
+# otherwise milter-greylist accepts the message before checking quotas.
+my %rate_froms;
+foreach my $c (@$conf) {
```
jcameron, line 278: This loop and the one after look very repetitive ... can they be combined or refactored somehow?

iliaross: Yeah, check this patch out https://github.com/virtualmin/virtualmin-gpl/pull/1222/commits/3c8374d44f60d4eb4abb9379272906df87b094a1.

jcameron: Nice thanks!

PR #1223, opened 2026-05-27
Fix to normalize apex record shorthand

iliaross:
Hey Jamie,

This PR addresses what appears to be a bug discussed [here](https://forum.virtualmin.com/t/change-ns-host-soa-in-all-domains/137313/2?u=ilia). It normalizes `@` DNS record names in the `modify-dns` add, remove, and update ops.

It also adds a regression test for adding and removing apex records via the CLI.


jcameron:
Thanks for this fix!

Issue #1224, opened 2026-05-28
Virtual server transfer deletes source before restore fails on sub-server database name conflict


jens-ingo:
Hello,

We hit a problematic recovery case during a Virtualmin “Transfer Virtual Server”.

A virtual server with sub-servers was transferred to another Virtualmin host. The backup and validation steps completed successfully, and the source server was then deleted. During restore on the target, the main domain and some sub-servers were restored, but one sub-server failed because its MariaDB database name already existed and belonged to another virtual server.

Error:

Re-creating virtual server blog.surfacenoise.be ..
.. a clash was detected : A MariaDB database named blog is already owned by virtual server blog. ::: .be

At that point the source had already been deleted, leaving the target only partially restored and requiring manual recovery from:

/tmp/virtualmin-transfer-16032

Expected behavior:

- Detect database name conflicts before deleting the source server.
- Or offer/perform a safe database rename during restore.
- Or keep the source intact until the target restore has fully completed.
- Or provide an automated recovery/continue mode for only the failed sub-server.

Actual behavior:

- Source server was deleted.
- Target restore was partial.
- The failed sub-server had to be manually recreated.
- The database dump had to be imported under a new name and the application config adjusted manually.

This is risky on shared hosting systems, where generic database names like `blog`, `wordpress`, or `site` can easily already exist. Could be nice to add a handler for theses cases into the script

Thanks.

iliaross:
Hello,

Thanks for reporting it!

Jamie, why don’t we run pre-flight clash tests and fail them without deleting any data?

jcameron:
That's a good suggestion ... to implement it we'd need to add a flag to the `restore-domain` command to just perform all checks first, and only delete the original after running that. But it's certainly possible. I'll put this on my TODO list.

iliaross:
> That's a good suggestion

Cool!

> I'll put this on my TODO list.

No need, I’ve been looking into this since it was reported. Check out this PR https://github.com/virtualmin/virtualmin-gpl/pull/1225.

jcameron:
The next Virtualmin release will handle this situation better by testing for any conflicts before deleting the original domain as part of a transfer.

PR #1225, opened 2026-05-30
Add pre-check transfer restores before source delete

iliaross:
Hey Jamie,

As [discussed here](https://github.com/virtualmin/virtualmin-gpl/issues/1224#issuecomment-4583599249), this PR adds a restore pre-check for virtual server transfers that delete the source server.

The target now validates the backup and restore-time clashes before the source is removed, so predictable failures like database ownership conflicts abort safely without deleting the original server.


jcameron:
Let me implement this one as I said I would on the ticket - I have a different approach that is much simpler.

iliaross:
Well, okay. I wish you had replied to that ticket sooner so I could have saved some time and effort implementing it in the first place.

No problem though, I’ll close it now.

jcameron:
I would advise that if you do plan non-trivial work to fix a ticket that you comment on it before starting, so we know who's doing what

iliaross:
Fair enough. I also recommend checking commit https://github.com/virtualmin/virtualmin-gpl/pull/1225/changes/71ef9d26b1e3f4b5e85bef9ff5d14b7960f52c20 from this PR independently.

PR #1226, opened 2026-05-31
Fix alias-domain creation so it respects ACME opt-outs

iliaross:
Hello Jamie!

This PR addresses an issue when adding a web alias to a domain with an existing ACME certificate. Before Virtualmin automatically re-requested the parent certificate to include the alias. This ignored explicit opt-outs, causing unwanted Let’s Encrypt requests and possible rate-limit hits during bulk alias creation.

Discussed here:
https://forum.virtualmin.com/t/multisite-wordpress-and-virtualmin-lets-encrypt/137295/39


jcameron:
Thanks for the patch!

PR #1227, opened 2026-06-01
Fix Dovecot/Webmin SSL conflicts on cert renewal for domains with dedicated IPs

bob7123:
## Summary

Cert renewals for domains with dedicated IPs create conflicting SSL configurations in both Dovecot and Webmin (miniserv), breaking IMAP/POP3 TLS and serving wrong certificates.

## Root Cause

`sync_dovecot_ssl_cert()` creates `local IP` blocks in `dovecot.conf` when `$d->{'virt'}` is true. If `local_name` entries also exist for the same domain (from a previous state, migration, or older Virtualmin version), Dovecot fails with `Conflict in setting ssl_cert` and all TLS connections fail silently.

`setup_ipkeys()` adds the IP to `ipcert` entries in `miniserv.conf`. Miniserv.pl only registers the SNI callback on the default SSL context, not per-IP contexts. On shared IPs, connections get the wrong certificate.

## Fix

Always use hostname-based (SNI) mechanisms instead of IP-based ones:
- **Dovecot**: skip `local IP` block creation, always use `local_name` (SNI works on Dovecot 2+)
- **Webmin**: skip adding IPs to `ipcert` entries, use domain names only (SNI callback handles hostname matching)

Original code is commented out, not deleted.

## Testing

Tested on Ubuntu 22.04 with Webmin 2.621 and Dovecot 2.3.16. Ran `virtualmin generate-letsencrypt-cert --domain <domain> --renew` multiple times. Verified:
- No `local IP` blocks created in `dovecot.conf`
- No IP-based `ipcert` entries in `miniserv.conf`
- `local_name` blocks created correctly for all domains
- No Dovecot config errors (`doveconf -n`)
- IMAPS/POP3S serving correct LE certs
- Webmin serving correct cert via SNI

## Related Issues

Fixes #1134
Related: #91, #214, #323

PR #1228, opened 2026-06-02
Fix Dovecot SSL conflicts on cert renewal for domains with dedicated IPs

### LINKED ISSUE #1134: “Use Default Certificate Path” option causes Dovecot to fail
Using the “Use Default Certificate Path” option causes Dovecot to fail. After using this option to move SSL certificate files from /etc/ssl/virtualmin/... to the virtual server’s home directory, the Dovecot configuration becomes inconsistent, referencing both old and new certificate paths, which prevents the service from starting. Debian 12.

Corrupted Dovecot config looks like:

local_name server1.net { 
  ssl_cert = </home/server1/ssl.combined
  ssl_cert = </home/server1/ssl.combined
  ssl_key = </home/server1/ssl.key
  ssl_key = </home/server1/ssl.key
}
  ssl_cert = </home/server1/ssl.combined
local_name *.server1.net {
  ssl_key = </home/server1/ssl.key
  ssl_cert = </home/server1/ssl.combined
  ssl_key = </home/server1/ssl.key
}
local_name server2.com { 
  ssl_cert = </home/server2/ssl.combined
  ssl_cert = </etc/ssl/virtualmin/16751732773193704/ssl.combined
  ssl_key = </home/server2/ssl.key
  ssl_key = </etc/ssl/virtualmin/16751732773193704/ssl.key
}
  ssl_cert = </etc/ssl/virtualmin/16751732773193704/ssl.combined
local_name *.server2.com {
  ssl_key = </home/server2/ssl.key
  ssl_cert = </home/server2/ssl.combined
  ssl_key = </etc/ssl/virtualmin/16751732773193704/ssl.key
}

bob7123:
## Summary

Cert renewals for domains with dedicated IPs create conflicting `local IP` blocks in dovecot.conf, breaking all IMAP/POP3 TLS connections.

Split from #1227 per Jamie's request. This PR covers Dovecot only; Webmin/miniserv fix is in a separate PR.

## Root Cause

`sync_dovecot_ssl_cert()` creates `local IP` blocks when `$d->{'virt'}` is true. If `local_name` entries also exist for the same domain (from migration, older Virtualmin version, or mixed config), Dovecot fails with `Conflict in setting ssl_cert`. All TLS connections fail silently.

## Fix

Always use `local_name` (SNI) instead of `local IP` blocks. Since Dovecot 2+ supports SNI via `local_name`, IP-based entries are unnecessary and cause conflicts on servers with multiple domains sharing an IP.

Two changes:
1. Skip populating `@ips` when domain has a dedicated IP (prevents `local IP` block creation)
2. Always enter the `local_name` code path (remove `!\->{'virt'}` guard)

Original code commented out, not deleted.

## Testing

Tested on Ubuntu 22.04 with Webmin 2.641 and Dovecot 2.3.16. Ran `virtualmin generate-letsencrypt-cert --domain <domain> --renew` multiple times. Verified:
- No `local IP` blocks created in dovecot.conf
- `local_name` blocks created correctly for all domains (24 blocks)
- No Dovecot config errors (`doveconf -n`)
- IMAPS (993) and POP3S (995) serving correct LE certs
- Full test script with 7 automated checks, all pass

Fixes #1134

### DIFF
--- a/feature-ssl.pl
+++ b/feature-ssl.pl
@@ -2464,11 +2464,18 @@ my $chain = &get_website_ssl_file($d, "ca");
 my $nochange = 0;
 
 my @ips;
-if ($d->{'virt'}) {
-	# Domain has it's own IPv4, and maybe v6
-	push(@ips, $d->{'ip'});
-	push(@ips, "[".$d->{'ip6'}."]") if ($d->{'virt6'} && $d->{'ip6'});
-	}
+# Skipping local IP block creation - always use local_name (SNI) instead.
+# local IP blocks conflict with local_name entries and Dovecot cannot
+# resolve both for the same connection (config error: "Conflict in
+# setting ssl_cert"). Since Dovecot 2+ supports SNI via local_name,
+# that is the correct mechanism for all domains regardless of whether
+# they have a dedicated IP.
+# Ref: https://github.com/virtualmin/virtualmin-gpl/issues/1134
+#if ($d->{'virt'}) {
+#	# Domain has it's own IPv4, and maybe v6
+#	push(@ips, $d->{'ip'});
+#	push(@ips, "[".$d->{'ip6'}."]") if ($d->{'virt6'} && $d->{'ip6'});
+#	}
 my ($cname, $cvalue) = &get_dovecot_ssl_dir("cert", $d->{'ssl_combined'});
 my ($kname, $kvalue) = &get_dovecot_ssl_dir("key", $d->{'ssl_key'});
 foreach my $ip (@ips) {
@@ -2567,7 +2574,8 @@ foreach my $ip (@ips) {
 		}
 	}
 
-if (!$d->{'virt'}) {
+if (1) {
+	# Always use local_name (SNI) for all domains, including those with dedicated IPs
 	# Domain has no IP, but Dovecot supports SNI in version 2
 	my @loc = grep { $_->{'name'} eq 'local_name' &&
 			 $_->{'section'} } @$conf;


bob7123:
One important note from testing: when applying this patch to an existing system, domains that previously had dedicated IPs will not have `local_name` entries yet (because the old code only created `local IP` blocks for them, never `local_name`). After applying the patch, a cert renewal must be triggered for each such domain to create the missing `local_name` entries. On our test system, we ran `virtualmin generate-letsencrypt-cert --domain <domain> --renew` for each affected domain and confirmed the `local_name` entries were created correctly.

Without this step, those domains fall through to the global default cert in `10-ssl.conf` until the next scheduled renewal.

iliaross:
Hello,

We deliberately use Dovecot `local {}` blocks for dedicated IP domains and `local_name {}` SNI blocks only for shared-IP domains.

Your PR forces all domains onto `local_name` and stops creating local IP blocks. That is way too broad.

We already have similar fixes in that direction in particular https://github.com/virtualmin/virtualmin-gpl/commit/04e683f9c that respects the Dovecot add-config file when inserting SNI sections.

This PR won't be merged, sorry!

bob7123:
Hi Ilia. Again, no need to apologize. My apologies instead for not considering all cases. I went back to square one and targeted a very narrow fix which I believe addresses your concerns. If it still misses the mark, please do not hesitate to let me know any issues you think I should investigate further.

New PR: https://github.com/virtualmin/virtualmin-gpl/pull/1236

jcameron:
Thanks for the PR, I will take a look at it now..

PR #1229, opened 2026-06-02
Fix Webmin serving wrong SSL cert on shared IPs after cert renewal

bob7123:
## Summary

Cert renewals for domains with dedicated IPs add the IP address to `ipcert` entries in `miniserv.conf`, causing Webmin to serve the wrong SSL certificate on shared IPs.

Split from #1227 per Jamie's request. This PR covers Webmin/miniserv only; Dovecot fix is in #1228.

## Root Cause

`setup_ipkeys()` adds `$d->{'ip'}` to the ipkeys array alongside domain names. Miniserv.pl creates per-IP SSL contexts for these entries but only registers the SNI callback on the default (`*`) context. When a client connects to a shared IP, miniserv uses the per-IP context directly and the SNI callback never fires, so hostname-based cert selection doesn't work. The client gets whichever domain's cert was last assigned to that IP (non-deterministic hash ordering).

## Fix

Skip adding IPs to ipkey entries. Use domain names only. The SNI callback on the default context handles hostname matching correctly for all domains.

Original code commented out, not deleted.

## Testing

Tested on Ubuntu 22.04 with Webmin 2.641. Ran `virtualmin generate-letsencrypt-cert --domain <domain> --renew` multiple times. Verified:
- No IP-based `ipcert` entries in `miniserv.conf` after renewal
- Webmin serves correct cert for each domain via SNI
- Tested with `openssl s_client -connect <shared-ip>:9595 -servername <hostname>`

Related: #1134

### DIFF
--- a/feature-ssl.pl
+++ b/feature-ssl.pl
@@ -1738,12 +1738,17 @@ local %miniserv;
 &$getfunc(\%miniserv);
 local @ipkeys = &webmin::get_ipkeys(\%miniserv);
 local @ips;
-if ($d->{'virt'}) {
-	push(@ips, $d->{'ip'});
-	}
-if ($d->{'virt6'}) {
-	push(@ips, $d->{'ip6'});
-	}
+# Skip adding IPs to the ipkey entry - use domain names only for SNI.
+# Including IPs causes miniserv to create per-IP SSL contexts that
+# bypass the SNI callback, serving the wrong cert on shared IPs.
+# Domain names alone are sufficient for SNI matching.
+# Ref: https://github.com/virtualmin/virtualmin-gpl/issues/1134
+#if ($d->{'virt'}) {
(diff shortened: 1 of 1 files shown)


iliaross:
Hello,

Thanks for looking into this!

Removing the IPs completely breaks the dedicated-IP case (clients without SNI would receive the default certificate). The real fix should be made in Webmin miniserv, probably by attaching the SNI callback to all contexts rather than only the default one.

Sorry, we cannot merge it as it is.

bob7123:
Hi Ilia. No need to apologize. I should apologize for not more carefully reviewing your code and architecture before providing a solution. I attempted to follow your suggestion and make changes to miniserv.pl. Please see https://github.com/webmin/webmin/pull/2766 for details.

PR #1231, opened 2026-06-04
Fix to deprivilege untrusted ZIP directory restores

iliaross:
Hey Jamie!

This PR hardens untrusted home-directory ZIP restores by validating a root-owned temporary copy of the archive before extraction, rejecting unsafe ZIP entries, and running unzip as the domain owner instead of the root Webmin process.


jcameron:
Can you explain the exploit path that this PR fixes? Looking at the original issue, it seems like we could solve it by just running unzip as the domain user?

iliaross:
Yes, running `unzip` as the domain user is the core good-enough fix, and matches what the TAR path already does.

And, after more testing, I came to the conclusion that you are right, as adding ZIP pre-validation was mostly a heuristic step with an archive copy for extra safety, which just doubled disk usage for large backups and made the process less practical, with little gain. So I just dropped it to keep things simple.

PR #1232, opened 2026-06-05
Fix to strip special bits after root tar restores

iliaross:
Hey Jamie!

This PR removes setuid/setgid bits from regular files after master/root tar directory restores.


jcameron:
But why is this needed? I didn't see any referenced to special bits in the linked ticket..

iliaross:
You’re right. I misunderstood that. I thought files under `~/homes` weren’t chowned after restore, but they are.

Since that already clears the main `setuid` and `setgid`, this PR is unnecessary, so I’m closing it.

PR #1234, opened 2026-06-12
Add Systemd user units support for domain owners

iliaross:
Hey Jamie!

This PR adds Virtualmin integration for [the new](https://github.com/webmin/webmin/pull/2760) Webmin Systemd Services and Units module.

It adds a Systemd User Units option to the Administrator's Webmin modules template section. When enabled for a domain, Virtualmin grants the domain owner scoped access to their own systemd user units using the Systemd module's `systemd_user_unit_acl()` helper.

The integration is guarded with `foreign_check("systemd")`, so nothing changes or fails when the Systemd module is not installed. ACL policy remains owned by the Systemd module instead of being duplicated in Virtualmin.

### DIFF
--- a/feature-webmin.pl
+++ b/feature-webmin.pl
@@ -670,6 +670,15 @@ if ($mods{'cron'} && !$noextra && $d->{'unix'} && !$chroot) {
 	push(@mods, "cron");
 	}
 
+if ($mods{'systemd'} && !$noextra && &foreign_check("systemd")) {
+	# Can manage supported systemd user units owned by the domain owner
+	&foreign_require("systemd");
+	my %acl = &systemd::systemd_user_unit_acl($d->{'user'});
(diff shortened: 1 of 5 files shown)


feature-webmin.pl
```diff
@@ -670,6 +670,15 @@ sub set_user_modules
 	push(@mods, "cron");
 	}
 
+if ($mods{'systemd'} && !$noextra && &foreign_check("systemd")) {
+	# Can manage supported systemd user units owned by the domain owner
+	&foreign_require("systemd");
+	my %acl = &systemd::systemd_user_unit_acl($d->{'user'});
```
jcameron: Is this function supposed to return a safe ACL? If so, perhaps rename it so it is clearer what it does

iliaross: Agreed. Renamed. Thanks!

PR #1235, opened 2026-06-15
Hotfix for no.auto.

DandelionSprout:
As simple as that, hopefully.

### DIFF
--- a/lang/no.auto
+++ b/lang/no.auto
@@ -977,7 +977,7 @@ newtmpl_def=Misligholde
 newtmpl_edelete=Standardmaler kan ikke slettes
 newtmpl_setdef=Angi standard mal på toppnivå
 newtmpl_setdefsub=Angi standard underservermal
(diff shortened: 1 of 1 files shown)


lang/no.auto
```diff
@@ -977,7 +977,7 @@ newtmpl_def=Misligholde
 newtmpl_edelete=Standardmaler kan ikke slettes
 newtmpl_setdef=Angi standard mal på toppnivå
 newtmpl_setdefsub=Angi standard underservermal
-newtmpl_name0=Standard instillinger
+newtmpl_name0=Standardinnstillinger
```
iliaross, line 980: Thanks! The `‎lang/no.auto‎` file is only meant for machine translations. The language strings that were reviewed or fixed by humans should be placed in `‎lang/no`, and this fixed key should be removed from `‎lang/no.auto‎` entirely.

iliaross:
Thanks for the patch!

PR #1236, opened 2026-06-16
Fix Dovecot SSL conflict: remove local_name entries before creating local IP blocks

### LINKED ISSUE #1134: “Use Default Certificate Path” option causes Dovecot to fail

bob7123:
When a dedicated-IP domain has both `local {}` and `local_name {}` entries in dovecot.conf, Dovecot fails with "Conflict in setting ssl_cert" and all IMAP/POP3 TLS connections fail silently. This can happen when `local_name` entries exist from a previous configuration state (migration, IP change, or older Virtualmin version).

**Fix:** In `sync_dovecot_ssl_cert()`, before creating or updating a `local IP` block for a dedicated-IP domain, check for and remove any conflicting `local_name` entries for that domain's hostnames. Uses the same `hostname_under_domain()` and `delete_section()` patterns already used elsewhere in the function.

This preserves the intended architecture: dedicated-IP domains use `local {}`, shared-IP domains use `local_name {}`. Only the conflict case is addressed.

Tested with `virtualmin generate-letsencrypt-cert --renew` on a server with dedicated-IP domains. After renewal: `local IP` blocks created, conflicting `local_name` entries removed, zero Dovecot config errors, correct certs served on IMAPS and POP3S, subdomain `local_name` entries for shared-IP subdomains preserved.

Fixes #1134

If there are any further concerns, please do not hesitate to let me know.

### DIFF
--- a/feature-ssl.pl
+++ b/feature-ssl.pl
@@ -2491,6 +2491,26 @@ foreach my $ip (@ips) {
 		}
 
 	if ($enable) {
+		# Remove any conflicting local_name blocks for this
+		# domain before creating the local IP block. Dedicated-IP
+		# domains should use local {} not local_name {}.
+		# Ref: https://github.com/virtualmin/virtualmin-gpl/issues/1134
+		my @lnloc = grep { $_->{'name'} eq 'local_name' &&
+				   $_->{'section'} } @$conf;
+		my @doms = ( $d, &get_domain_by("alias", $d->{'id'}) );
+		my @conflicting = grep {
+			&hostname_under_domain(\@doms, $_->{'value'})
+			} @lnloc;
+		foreach my $cl (@conflicting) {
+			&dovecot::delete_section($conf, $cl);
+			@$conf = grep { $_ ne $cl } @$conf;
+			@$conf = grep {
+				$_->{'sectionname'} ne $cl->{'name'} ||
+				$_->{'sectionvalue'} ne $cl->{'value'}
+				} @$conf;
+			&flush_file_lines($cl->{'file'});
+			}
+
 		# Needs a cert for the IP
 		if (!$l) {
 			$l = { 'name' => 'local',


feature-ssl.pl
```diff
@@ -2491,6 +2491,26 @@ sub sync_dovecot_ssl_cert
...
 	if ($enable) {
+		# Remove any conflicting local_name blocks for this
+		# domain before creating the local IP block. Dedicated-IP
+		# domains should use local {} not local_name {}.
+		# Ref: https://github.com/virtualmin/virtualmin-gpl/issues/1134
+		my @lnloc = grep { $_->{'name'} eq 'local_name' &&
+				   $_->{'section'} } @$conf;
+		my @doms = ( $d, &get_domain_by("alias", $d->{'id'}) );
+		my @conflicting = grep {
+			&hostname_under_domain(\@doms, $_->{'value'})
```
jcameron, line 2502: Woudn't this also match sub-domains that might not have a private IP address?

bob7123: Good question. I traced `hostname_under_domain()` through every case I could think of to make sure we are not breaking subdomains.

Given: parent domain `example.com` has a dedicated IP. Our patch runs during its cert renewal.

**Cases where `local_name` entries ARE removed by our patch:**

Parent domain itself (`example.com`) and its wildcard (`*.example.com`): removed. Correct, because the parent's `local IP` block on the dedicated IP serves the same cert.

Service hostnames under the parent (`mail.example.com`, `www.example.com`): removed. Correct, because the parent cert covers these and the `local IP` block handles connections to the dedicated IP.

Alias domains (`example.net` aliased to `example.com`): removed. Correct, because aliases share the parent's IP and cert.

Orphaned entries (deleted virtual server, `local_name` still in dovecot.conf): removed. Correct cleanup. `get_domain_by` returns nothing, `hostname_under_domain` returns 1.

Manually added `local_name` entries for hostnames not registered in Virtualmin: removed. This matches how `hostname_under_domain()` works elsewhere in the codebase. It considers hostnames under a parent domain as belonging to it unless registered as separate domains. Our patch does not introduce new matching behavior here.

**Cases where `local_name` entries are NOT removed (preserved):**

Subdomains that are registered virtual servers, with or without a private IP: preserved. `get_domain_by("dom", "dev.example.com")` at line 2403 finds them and returns 0. This is the case you asked about.

Subdomain wildcards (`*.dev.example.com`): preserved. The regex at line 2401 splits on the first dot, giving `="*"` and `="dev.example.com"`. Since `` does not equal `example.com`, it returns 0.

Completely unrelated domains: preserved. Different domain names fail all checks.

I endeavored to think of all possibilities. However, if there is a specific case I have missed, I would like to know so I can address it.

jcameron: Ah yes it looks like hostname_under_domain excludes sub-domains already, so we're good

PR #1238, opened 2026-06-22
Fix selected differential backup chain state handling

iliaross:
Hello Jamie!

This fixes several edge cases in selected full-backup differential chains considering your latest changes to _master_!

Previously, full backup snapshot state could be advanced before the full backup was actually usable, or selected differential schedules could be configured against baselines that could not be represented or did not exist.

What changed (you can see more details in each linked commit):

- Only advance the default differential snapshot after a chained full backup archive is successfully created.
- Move full-backup snapshot promotion to the final backup result path, so state is advanced only after all destinations succeed.
- Clear new full-backup snapshot state when the backup fails, including failures after tar has already updated state, such as remote upload failures.
- Keep destination locks held until snapshot state has been updated.
- Disallow using legacy scheduled backup ID `1` as a selected full-backup baseline, because `increment == 1` already represents legacy differential mode.
- Always keep per-schedule snapshot state for scheduled full backups so selected differential chains can be created later.
- Fail selected differentials clearly when the chosen full-backup snapshot is missing, instead of silently creating a misleading backup from the wrong baseline.

### DIFF
--- a/backups-lib.pl
+++ b/backups-lib.pl
@@ -804,9 +804,6 @@ local $ok = 1;
 local @donedoms;
 local ($okcount, $errcount) = (0, 0);
 local @errdoms;
-my %fullstateok;
-my $selected_full = $increment == 0 &&
-		    defined($id) && $id =~ /^\d+$/ && $id >= 3;
 local %donefeatures;				# Map from domain name->features
 local @cleanuphomes;				# Temporary homes
 local %donedoms;				# Map from domain name->hash
@@ -1032,8 +1029,6 @@ DOMAIN: foreach $d (sort { $a->{'dom'} cmp $b->{'dom'} } @$doms) {
 	if ($onebyone && $homefmt && $dok && $anyremote) {
 		# Transfer this domain now
 		local $df = "$d->{'dom'}.$hfsuffix";
-		$fullstateok{$d->{'id'}} = 1
-			if ($selected_full && $anylocal && $d->{'id'});
 		&$cbfunc($d, 1, "$dest/$df") if ($cbfunc);
 		local $tstart = time();
 		local $binfo = { $d->{'dom'} =>
@@ -1212,8 +1207,6 @@ DOMAIN: foreach $d (sort { $a->{'dom'} cmp $b->{'dom'} } @$doms) {
 				}
 			else {
 				&$second_print($text{'setup_done'});
-				$fullstateok{$d->{'id'}} = 1
-					if ($selected_full && $d->{'id'});
 				local @tst = stat("$dest/$df");
 				if ($mode != 0 && !$done_transferred_sz++) {
 					$transferred_sz += $tst[7];
@@ -1528,17 +1521,6 @@ else {
 	}
 $sz += $transferred_sz;
 
-my %archiveerrids = map { $_->{'id'}, 1 }
-		    grep { $_->{'id'} } @errdoms;
-my $mark_fullstate_ok = sub {
-	return if (!$selected_full);
-	foreach my $sd (@_) {
-		next if (!$sd || !$sd->{'id'} || $archiveerrids{$sd->{'id'}});
-		$fullstateok{$sd->{'id'}} = 1;
-		}
-	};
-&$mark_fullstate_ok(@donedoms) if ($ok && $anylocal);
-
 foreach my $desturl (@$desturls) {
 	local ($mode, $user, $pass, $server, $path, $port) =
 		&parse_backup_url($desturl);
@@ -1622,7 +1604,6 @@ foreach my $desturl (@$desturls) {
 			}
 		&unlink_file($infotemp);
 		&unlink_file($domtemp);
-		&$mark_fullstate_ok(@donedoms) if (!$err);
 		&$second_print($text{'setup_done'}) if ($ok);
 		}
 	elsif ($ok && ($mode == 2 || $mode == 13) && (@destfiles || !$dirfmt)) {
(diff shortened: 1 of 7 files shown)


backups-lib.pl
```diff
@@ -2081,21 +2081,37 @@ sub backup_domains
...
-	}
-
-# For any domains that failed and were full backups, clear the differential
-# file so that future differential backups aren't diffs against it
+# Update full-backup differential state only after the complete backup result
+# is known. A failed upload can happen after tar has already advanced the
+# snapshot files, so failed full backups must not become a future baseline.
 if ($increment == 0 && &has_incremental_tar()) {
-	foreach my $d (@errdoms) {
+	my %errids = map { $_->{'id'} ? ($_->{'id'}, 1) : () } @errdoms;
```
jcameron: This is pretty hard to read - best to use grep and map chained

jcameron:
Didn't the old incremental backups before I added this feature already have the behavior that the incremental state could be advanced even before all destinations were complete?

iliaross:
> Didn't the old incremental backups before I added this feature already have the behavior that the incremental state could be advanced even before all destinations were complete?

Yeah, `tar` used to advance the incremental snapshot during archive creation before all destinations were complete. If a later destination failed, the old code could leave that advanced snapshot in place even though the full backup was not usable everywhere.

jcameron:
That's kind of a judgement call , because maybe the admin does prefer the current behavior if they are fine with at least one destination being OK. So maybe limit this PR to bugs in the new feature?

iliaross:
> That's kind of a judgement call , because maybe the admin does prefer the current behavior if they are fine with at least one destination being OK. So maybe limit this PR to bugs in the new feature?

Ah, I see what you mean now! See this patch https://github.com/virtualmin/virtualmin-gpl/pull/1238/commits/d88185d6df766a217fd959c59fe48b845e8a3a88. It will keep selected full-backup snapshot state only after the backup reaches destination, while preserving old behavior.

jcameron:
Hang on, this is getting even more complex and we're fixing a whole bunch of separate issues in this PR that makes it hard to understand! Can we limit this to just bug fixes, and not change the existing behavior of when the incremental state is advanced? That *may* be something we want, but it should be in a separate PR.

iliaross:
> Can we limit this to just bug fixes, and not change the existing behavior of when the incremental state is advanced? That _may_ be something we want, but it should be in a separate PR.

Okay. I've split it into its own https://github.com/virtualmin/virtualmin-gpl/pull/1246 PR that needs to go in once this one lands.

backups-lib.pl
```diff
@@ -2109,13 +2109,10 @@ sub get_incremental_file
...
-	# This is a full backup, but are there other incrementals referencing it?
-	my @incrs = grep { $_->{'increment'} == $id } &list_scheduled_backups();
-	if (@incrs) {
-		# Yes, so it gets its own incremental file
-		return "$incremental_backups_dir/$id/$d->{'id'}";
-		}
+elsif ($mode == 0 && $id =~ /^\d+$/ && $id >= 3) {
+	# This is a scheduled full backup. Keep its own snapshot even before
+	# any differential schedule references it, so future chains can use it.
+	return "$incremental_backups_dir/$id/$d->{'id'}";
```
jcameron: This can't be done safely for existing backups, as it would cause the incremental state file to change on Virtualmin upgrade. That's why I did it the way I did..

backups-lib.pl
```diff
@@ -2081,23 +2081,23 @@ sub backup_domains
 		}
 	}
 
-# Release lock on dest file
-foreach my $lockfile (@lockfiles) {
-	&unlock_file($lockfile);
-	}
+	# Release lock on dest file
```
jcameron: Why this indentation change?

backups-lib.pl
```diff
@@ -2081,23 +2081,23 @@ sub backup_domains
...
+			if ($d->{'id'}) {
+				&unlink_file(&get_incremental_file($d, $increment, $id));
+				}
 			}
 		}
-	}
 
-return ($ok, $sz, \@errdoms);
-}
+	return ($ok, $sz, \@errdoms);
```
jcameron: Same here with the indentation

backups-lib.pl
iliaross: Can you elaborate and explain exactly why "this can't be done safely for existing backups" so I can get a better picture of your view of the problem?

jcameron: What I mean is, with this change the incremental file for a full backup would change EVEN if there is no differential backup specifically referencing it, which we don't want.

Also, what's with this check for `$id =~ /^\d+$/ && $id >= 3` ? The scheduled backup ID doesn't mean anything, except for legacy ID 1. There would never ben an ID 0.

iliaross: Understood now! Thanks for pointing this out! I have fixed it in this https://github.com/virtualmin/virtualmin-gpl/commit/5635bcb7f5b788ab2391b197d39651ac1fe4d24b change. Please check.

backups-lib.pl
iliaross: These indentation issues are fixed too.

backups-lib.pl
```diff
@@ -2109,13 +2109,11 @@ sub get_incremental_file
 	# Incremental against a specific backup
 	return "$incremental_backups_dir/$mode/$d->{'id'}";
 	}
-elsif ($mode == 0 && $id) {
-	# This is a full backup, but are there other incrementals referencing it?
+elsif ($mode == 0 && $id && $id ne "1") {
+	# This full backup gets its own snapshot only if a differential
+	# schedule actually references it.
 	my @incrs = grep { $_->{'increment'} == $id } &list_scheduled_backups();
-	if (@incrs) {
```
jcameron, line 2115: Why is the check for @incrs removed here? We only want to have a per-backup incremental file for full backups if some other incremental backups specifically refer to it..

iliaross: The check is still there, just collapsed into a one liner:

```perl
return "$incremental_backups_dir/$id/$d->{'id'}" if (@incrs);
```

I have brought original more readable style with the comment.

jcameron: Oh sorry I missed that! Looks good ..

Issue #1239, opened 2026-06-24
SSL fail when creating new server.


GitHubAIH:
Creating SSL certificate and private key ..
.. done
Adding new SSL virtual website ..
.. certificate file is not valid : Data starts with , but expected -----BEGIN (CERTIFICATE|PUBLIC KEY)-----

Server is created but SSL option is off.
Reactivating SSL gives:

Creating SSL certificate and private key ..
.. done
Adding new SSL virtual website ..
.. certificate file is not valid : Data starts with , but expected -----BEGIN (CERTIFICATE|PUBLIC KEY)-----

iliaross:
Hello,

What Virtualmin and Webmin versions do you have installed?

GitHubAIH:
Webmin version     2.641
Virtualmin version 8.1.0 GPL 
Operating system AlmaLinux 9.8

iliaross:
What does the disk quota say for the virtual server in question, and what is the `df -h` command output?

GitHubAIH:
[root@~]# df -h
Filesystem                        Size  Used Avail Use% Mounted on
devtmpfs                          4.0M     0  4.0M   0% /dev
tmpfs                              12G     0   12G   0% /dev/shm
tmpfs                             4.7G  464M  4.2G  10% /run
efivarfs                          256K   32K  220K  13% /sys/firmware/efi/efivars
/dev/mapper/almalinux_empty-root  147G   91G   57G  62% /
/dev/sda2                         960M  347M  614M  37% /boot
/dev/sda1                         599M  7.7M  592M   2% /boot/efi
tmpfs                             2.4G     0  2.4G   0% /run/user/0

<img width="1465" height="209" alt="Image" src="https://github.com/user-attachments/assets/6a1efc18-a2f2-4a14-bf3e-38045b5eafed" />

Disk space isn't the issue.

iliaross:
And if you check SSL certificates files from "Manage Virtual Server ⇾ Setup SSL Certificate: Current Certificate" page do they look okay?

GitHubAIH:
<img width="920" height="176" alt="Image" src="https://github.com/user-attachments/assets/1963583f-ac74-46aa-8459-a55d27fd9be7" />

Empty, also home dir has no certificates.  And /etc/letsencrypt/ has no reference of certificates in archive or live.

iliaross:
[...]

PR #1240, opened 2026-06-24
Fix silent SSL certificate write failures

iliaross:
Hello Jamie!

This PR fixes SSL setup failures caused by empty certificate or key files being treated as successfully generated.

Self-signed SSL generation now requires non-empty OpenSSL output, validates the cert/key pair before writing, and re-validates the final destination files after writing. This prevents Virtualmin from reporting certificate generation as done when the actual files are empty or invalid.

The domain-user file write path now also reports failed closes and failed atomic renames instead of silently succeeding. `write_ssl_file_contents` preserves the existing fatal-on-write-failure behavior for existing callers, while SSL generation opts into returned errors so setup can report a clean failure message.

Also fixes a lock leak on SSL generation failure and regenerates default certs when both existing cert/key files are empty.

Reported here https://github.com/virtualmin/virtualmin-gpl/issues/1239.

### DIFF
--- a/feature-ssl.pl
+++ b/feature-ssl.pl
@@ -2090,15 +2090,33 @@ else {
 		" -subj ".quotemeta($subject)." $addtextsup -utf8 2>&1");
 	}
 my $rv = $?;
-if (!-r $certtemp || !-r $keytemp || $rv) {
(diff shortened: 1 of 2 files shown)


jcameron:
Thanks for the fix!

PR #1242, opened 2026-06-30
Add IP address support for ACME SSL certificates

iliaross:
Hello, Jamie!

This PR adds Virtualmin support for ACME certificates containing IP address SANs when the installed `certbot` version  supports them.

- Add support for requesting ACME certificates with IP address identifiers when the installed Certbot versions support them.

- Validate IP certificate requests early, requiring the IP to belong to the virtual server and allowing shared IPs only for the default website.

- Force IP address certificate requests to use web-based validation and prevent DNS fallback for IP identifiers, as DNS validation proves control of a domain name but not control of the IP address itself.

- Enforce safe automatic renewal settings for short-lived IP certificates by rejecting renewal thresholds above the supported maximum.

- Update Webmin and Usermin miniserv SNI mappings so IP SAN certificates are served correctly for direct IP access.

- Preserve existing per-service certificate behavior while adding IP-aware sync for Webmin, Usermin, Dovecot, and Postfix.

- Extend certificate parsing to accept SAN-only certificates and parse IP address SAN entries.

- Improve certificate display and CA issuer checks for SAN-only/IP certificates.

- Update UI and CLI wording from “hostnames” to “certificate names” or “identifiers” where IP addresses are now allowed.

- Add functional coverage for installing and serving an IP SAN certificate without requiring live ACME access.


copycert-lib.pl
```diff
@@ -360,28 +360,67 @@ sub get_all_domain_service_ssl_certs
...
 sub update_all_domain_service_ssl_certs
 {
-my ($d, $before) = @_;
+my ($d, $before, $force_domain) = @_;
 my $tmpl = &get_template($d->{'template'});
+my $force_dnames = ref($force_domain) ? $force_domain : undef;
+# The requested identifiers are only needed while service SNI records are
+# rebuilt, so keep them out of the saved domain object. Using local also
+# restores the original value if a sync function dies mid-rebuild.
+local $d->{'_service_ssl_dnames'};
```
jcameron: I don't get what this `local` is doing here?

copycert-lib.pl
```diff
@@ -433,9 +472,17 @@ sub ipkeys_to_domain_cert
 {
 my ($d, $ipkeys) = @_;
 foreach my $k (@$ipkeys) {
-	my $md = &indexof($d->{'dom'}, @{$k->{'ips'}}) >= 0;
-	my $m = $d->{'virt'} && &indexof($d->{'ip'}, @{$k->{'ips'}}) >= 0;
-	my $m6 = $d->{'virt6'} && &indexof($d->{'ip6'}, @{$k->{'ips'}}) >= 0;
+	my @dnames = ($d->{'dom'}, "*.".$d->{'dom'});
+	my $md = defined(&miniserv_ipkey_has_any) ?
```
jcameron: Why would `miniserv_ipkey_has_any` not be defined here?

functional-test.pl
```diff
@@ -7737,6 +7749,29 @@ package virtual_server;
...
 	  'save' => 'PRIVATE_IP',
 	},
+	{ 'command' => 'list-domains.pl',
+	  'args' => [ [ 'multiline' ],
+		      [ 'domain', $test_domain ] ],
+	  'grep' => [ 'IP address: $PRIVATE_IP ' ],
+	  'antigrep' => [ 'IP address: $PRIVATE_IP \\(Shared\\)' ],
+	},
+
+	# Check that malformed IP identifiers fail before domain validation.
```
jcameron, line 7759: Surely we can't test a real ACME cert request without a real domain name?

functional-test.pl
```diff
@@ -7797,7 +7834,8 @@ package virtual_server;
 	{ 'command' => 'openssl s_client -host $PRIVATE_IP'.
 		       ' -servername mail.'.$test_domain.
 		       ' -port 993 </dev/null',
-	  'grep' => [ 'O=Test SSL domain', 'CN=(\\*\\.)?'.$test_domain ],
+	  'grep' => [ 'O\s*=\s*Test SSL domain',
```
jcameron: Why the extra \s* here?

functional-test.pl
```diff
@@ -7809,7 +7847,64 @@ package virtual_server;
 	{ 'command' => 'openssl s_client -host $PRIVATE_IP'.
 		       ' -servername mail.'.$test_domain.
 		       ' -port 465 </dev/null',
-	  'grep' => [ 'O=Test SSL domain', 'CN=(\\*\\.)?'.$test_domain ],
+	  'grep' => [ 'O\s*=\s*Test SSL domain',
+		      'CN\s*=\s*(\\*\\.)?'.$test_domain ],
+	},
+
+	# Install a SAN-only cert with an IP identifier. This catches regressions
```
jcameron, line 7851: What's a SAN-only cert?

functional-test.pl
```diff
@@ -7905,23 +8004,23 @@ package virtual_server;
 	{ 'command' => 'openssl s_client -host mail.'.$test_domain.
 		       ' -servername '.$test_domain.
 		       ' -port 993 </dev/null',
-	  'grep' => [ 'O=Test SSL domain', 'CN=(\\*\\.)?'.$test_domain ],
+	  'grep' => [ 'O\s*=\s*Test SSL domain',
+		      'CN\s*=\s*(\\*\\.)?'.$test_domain ],
 	},
 
-	# Validate that Dovecot cert still works with SNI on the alias
```
jcameron, line 7911: Why not use the original test-imap comamnd, which also does an proper IMAP login?

generate-acme-cert.pl
```diff
@@ -181,14 +182,28 @@ package virtual_server;
 else {
 	# Hostnames given
 	foreach my $dname (@dnames) {
-                my $checkname = $dname;
-                $checkname =~ s/^www\.//;
-                $checkname =~ s/^\*\.//;
-                $err = &valid_domain_name($checkname);
-                &usage($err) if ($err);
+		next if (&letsencrypt_is_ip_cert_identifier($dname));
```
jcameron, line 185: Isn't this going to skip the checks for all hostnames that are not IPs?

letsencrypt.cgi
```diff
@@ -189,7 +214,9 @@ else {
 		}
 	else {
 		$info = &cert_file_info($cert);
-		@gotnames = &unique($info->{'cn'}, @{$info->{'alt'}});
+		@gotnames = &unique(grep { $_ } ($info->{'cn'},
```
jcameron: Why the need for `grep { $_ }` here? Can `cn` be empty?

virtual-server-lib-funcs.pl
```diff
@@ -8915,10 +8915,45 @@ sub create_initial_letsencrypt_cert
...
+		}
+	return 0;
+	}
+my $renew = !defined($d->{'letsencrypt_renew'}) ||
+	    $d->{'letsencrypt_renew'} eq '' ||
+	    $d->{'letsencrypt_renew'};
+my $rerr = &validate_letsencrypt_ip_cert_renewal($d, \@dnames, $renew);
+if ($rerr) {
+	# Initial IP certificates inherit automatic renewal from templates,
+	# so reject unsafe renew-before-expiry values before requesting one.
```
jcameron, line 8942: Can you explain further what is going on here?

copycert-lib.pl
iliaross: This lets miniserv IP-key generation see the IPs requested for the cert without saving `_service_ssl_dnames` into the domain file.

copycert-lib.pl
iliaross: Good catch, it doesn't need to be checked!  Fixed here https://github.com/virtualmin/virtualmin-gpl/pull/1242/commits/4b4318e48cf84e9fe0c61a929774284077ccdba1.

functional-test.pl
iliaross: No, we cannot reliably test a real ACME IP cert request. Instead, the test covers the feature in two parts. First, invalid ACME inputs fail before Certbot is called, and second, the successful path is tested by installing a locally generated CA-signed cert with an IP SAN, then verifying parsing, miniserv IP-key creation, and service presentation over the IP.

functional-test.pl
iliaross: To cover OpenSSL output differences by version build.

functional-test.pl
iliaross: A SAN-only certificate has no Subject Common Name (CN). Its identities are only in the Subject Alternative Name (SAN) extension.

iliaross: Just to clarify, for standard domain certificates, most still include a CN field even though browsers and TLS clients actually validate using SAN these days.

That meant Virtualmin's older logic could usually survive relying on "CN is present" assumptions. IP certificates revealed that weakness since they typically contain only SAN entries.

On top of that, an IP address can't be properly expressed as a DNS-style CN match. The actual identity lives in the SAN `iPAddress` field, which is X.509 type 7.

So SAN was already used for domain aliases, but this PR makes Virtualmin handle the stricter modern case where the cert has no CN at all and the only usable names are SANs, including IP SANs.

functional-test.pl
iliaross: Because `test-imap --server alias.example.com` tests DNS resolution and mail login path too.

But in this spot we only need to test Dovecot SNI certificate selection for the alias.

generate-acme-cert.pl
iliaross: No, no! It only skips validation for the current identifier, when that identifier is an IP address.

letsencrypt.cgi
iliaross: As explained earlier, CN can be empty and SAN-only certificates intentionally have no CN.

Without `grep`, an empty CN could be counted as a cert name, causing odd output.

virtual-server-lib-funcs.pl
iliaross: IP address certs are short-lived. The normal template default of renewing 21 days before expiry is invalid for them.

For initial cert setup, `letsencrypt_renew` might not be defined, in which case the system falls back to the template's default behavior.

So, this code checks whether renewal is effectively enabled, and if the requested cert contains IPs. It will reject unsafe template renewal period before requesting the cert.

letsencrypt.cgi
jcameron: Hmm, this is rather difficult to read. Perhaps instead extract this logic into a separate function like : 

```
sub get_all_certificate_names
{
my ($info) = @_;
my @rv;
push(@rv, $info->{'cn'}) if ($info->{'cn'});
push(@rv, @{$info->{'alt'}}) if ($info->{'alt'});
return @rv;
}
```

Actually maybe this could be done in `cert_info`

generate-acme-cert.pl
jcameron: Ok got it thanks!

functional-test.pl
jcameron: Ok but why can't `test-imap` check that dovecot SNI cert selection is working? We can modify it if needed to validate that the SSL cert hostname is valid.

functional-test.pl
jcameron: Thanks! But why do we need to use `openssl` to generate a SAN-only cert here, rather than just using Virtualmin's own `generate-cert` API command?

functional-test.pl
jcameron: Oh got it, thanks!

functional-test.pl
jcameron: Ah I see, so it's just testing the validation of the generate-acme-cert command, rather than actually requesting a cert?

copycert-lib.pl
jcameron: Thanks!

copycert-lib.pl
jcameron: This seems like a very convoluted way of passing in the IP. Why does the existing method of passing though hostnames in the cert not work?

letsencrypt.cgi
iliaross: I would not push this down into `cert_info`, as that function should keep returning parsed certificate fields. Here is the better fix https://github.com/virtualmin/virtualmin-gpl/pull/1242/commits/9f352c7771d318dd5bb647a3a728d0797b1cf1d2.

functional-test.pl
iliaross: Oh, agreed! Here is the patch https://github.com/virtualmin/virtualmin-gpl/pull/1242/commits/be54a2793804a61f9e107cb48ab5c1056c3fb446.

functional-test.pl
iliaross: Because `generate-cert.pl` always supplies a CN, defaulting to `*.domain` when `--cn` is omitted, so it cannot create a true SAN-only certificate.

functional-test.pl
iliaross: Yes, these are negative validation tests for `generate-acme-cert.pl` only, so we can verify to reject malformed IPs before calling `certbot` command.

copycert-lib.pl
iliaross: Okay, I agree that passing the requested names through the domain object was not clean. I rechecked the flow, and we do not really need to pass the requested names at all.

The installed certificate is now the source of truth. `domain_miniserv_ssl_ips` reads the actual CN/SANs from the cert, including IP SANs.

The ACME path only passes a `force-sync` boolean so service SNI records are rebuilt when an IP cert was requested.

See this patch https://github.com/virtualmin/virtualmin-gpl/pull/1242/commits/e9450b700449bc49c044d09992c336810d7387c0.

jcameron: Thanks!

functional-test.pl
jcameron: Sounds good!

functional-test.pl
jcameron: Ok maybe we can fix this in a follow-up PR, to improve the cleanliness of the testing code, and also allow users to create IP-only certs.

iliaross: No, this part in particular belongs to this PR specifically. We should keep it here.

functional-test.pl
jcameron: That's nice, although in a followup PR could you change mailboxes::imap_login to return cert info so that the whole IMAP login protocol doesn't have to be implemented in test-imap.pl ? Instead it could just use that returned cert to validate that the CN and altName is correct.

letsencrypt.cgi
jcameron: Nice thanks

virtual-server-lib-funcs.pl
jcameron: Ok got it. Although we should consider just making the renewal setting on/off instead of a number of days, and let Virtualmin decide how many days in advance to renew based on the expiry and whether an IP is being used.

copycert-lib.pl
```diff
@@ -361,27 +361,60 @@ sub get_all_domain_service_ssl_certs
...
 		}
+	else {
+		my $func = "copy_".$svc->{'id'}."_ssl_service";
+		&$func($d) if (defined(&$func));
+		}
+	}
+# IP identifiers may be new to this cert, so the matching per-domain service
+# records might not have existed in the "before" snapshot. Force enabled SNI
+# services, and always include Webmin/Usermin so miniserv can serve the IP cert.
+if ($force_sync || &domain_cert_has_ip_identifiers($d)) {
```
jcameron, line 407: Can you explain in more detail why this block is needed? Specifically in what end-user scenario it would apply?

functional-test.pl
iliaross: Yes, I think.

copycert-lib.pl
iliaross: Yes. This is needed because `before` only includes services already using the old cert. When a new cert adds an IP SAN, the matching per-IP service records may not exist yet, so they won’t appear in that snapshot.

For example, if an admin adds the server IP to a domain's ACME cert, miniserv needs new `ipcert_...`/`ipkey_...` records so Webmin and Usermin can serve that cert for `https://IP:10000` and https://IP:20000`. Without this fallback, only existing service mappings are refreshed, so newly valid IP mappings would be missed.

jcameron: Ah I see, thanks!

PR #1244, opened 2026-07-01
Fix to replace lexical `local` declarations with `my`

iliaross:
Hello Jamie!

This PR replaces `local` declarations that were being used as ordinary temporary variables with lexical `my` declarations.

The change preserves `local` where dynamic scoping is required, including package globals, special Perl variables, signal/environment localization, `%access`, and callback-visible context variables.


jcameron:
Beautiful, thanks!

PR #1246, opened 2026-07-02
Fix tracking selected full backup state

iliaross:
Hello Jamie!

This PR improves snapshot handling for selected full backup chains as discussed in https://github.com/virtualmin/virtualmin-gpl/pull/1238#issuecomment-4862187329.

Now selected full-backup state is retained only for domains whose full backup reaches at least one durable destination. Successful selected full snapshots are promoted, while snapshots from failed selected full backups are removed so future differentials do not silently depend on an unusable full backup.


iliaross:
Jamie, I've fixed the merge conflict. Please take a look.

iliaross:
I believe it needs more fixes... I will update this with progress.

iliaross:
Jamie, this is ready for review now.

iliaross:
Jamie, is this good to merge or not?

jcameron:
So can you explain the failure mode that this PR fixes? Because in a regular backup to a local destination, the incremental file is updated in feature-dir.pl after the archive is created..

iliaross:
> So can you explain the failure mode that this PR fixes? Because in a regular backup to a local destination, the incremental file is updated in feature-dir.pl after the archive is created..

Yeah, for local backups.

Yet this issue only affects a full backup selected as a differential base. If final packaging or upload fails after tar updates its state, that state points to a backup that doesn’t exist.

This PR keeps the state if at least one destination succeeds or otherwise, it removes it, with existing incremental behavior is unchanged.

jcameron:
Ok so the potential problem is that if an incremental backup fails to be uploaded to it's destination, we shouldn't update the state file?

Or is it that if a full backup fails to upload we shouldn't clear the state file?

I'm not quite sure of the full sequence of events this is fixing..

iliaross:
The issue is with a full backup that is selected as the base for future differentials, not with a failed differential upload.

The failure sequence is:

1. `tar --listed-incremental` updates the selected full’s snapshot state while creating the home archive.
2. The backup can still fail later, during packaging, metadata writing, or upload.
3. If we keep that new state, the next differential can be created against a full backup that never reached any durable destination.
4. That leaves an unrestorable chain.

So yes, the important case is "if a selected full backup fails to upload, we should not keep and publish the new state”.  And, I added a few more fixes on the latest commit because the earlier version still could published state too early.

The current version of this PR should make selected-full state fail closed.

backups-lib.pl
```diff
@@ -804,6 +804,84 @@ sub backup_domains
...
+my $selected_full = 0;
+if ($increment == 0 && $id && $id ne "1") {
+	my @incrs = grep { $_->{'increment'} == $id } &list_scheduled_backups();
+	$selected_full = @incrs ? 1 : 0;
+	}
+
+# Serialize access to every snapshot this backup can read or replace. These
+# locks are acquired before domain locks and held until state is finalized, so
+# overlapping full and differential runs cannot publish, restore or remove each
+# other's snapshot generation.
```
jcameron, line 819: Isn't there already protection against multiple backups running at the same time for the same domain though, via locking? That should protect against this kind of failure.

backups-lib.pl
```diff
@@ -804,6 +804,84 @@ sub backup_domains
...
+			push(@dfiles, &get_incremental_file(
+				$ld, $increment, $id));
+			}
+		foreach my $lockfile (@dfiles) {
+			push(@incremental_lockfiles, $lockfile)
+				if ($lockfile && !$lockseen{$lockfile}++);
+			}
+		}
+	@incremental_lockfiles = sort @incremental_lockfiles;
+	foreach my $lockfile (@incremental_lockfiles) {
```
jcameron, line 849: Doesn't get_incremental_file already create the directory for the file it returns?

backups-lib.pl
```diff
@@ -804,6 +804,84 @@ sub backup_domains
...
+		}
+	foreach my $lockfile (@lockfiles) {
+		&unlock_file($lockfile);
+		}
+	};
+
+# Invalidate every selected snapshot before the first domain is archived. If
+# this process is terminated part-way through a multi-domain backup, no domain
+# that was not reached can retain state for an older full backup generation.
+if ($selected_full) {
```
jcameron, line 866: I think you can exclude this, that feels like an over-complication for a very rare situation.

backups-lib.pl
```diff
@@ -2080,23 +2259,129 @@ sub backup_domains
...
+	&unlink_file($file);
+	return "Failed to remove unusable snapshot $file"
+		if (-e $file || -l $file);
+	}
+return undef;
+}
+
+# verify_selected_full_backup_state(&state, file)
+# Returns undef if a published snapshot has the expected mode and ownership.
+sub verify_selected_full_backup_state
```
jcameron: Why are these checks needed? Seems like over-complexity to check for permissions on a file Virtualmin itself created.

backups-lib.pl
iliaross: Not sufficient protection. The existing domain lock is released, while upload and snapshot publication happen much later.

So destination locking only protects jobs using the same destination URL. Two backups for the same domain but different destinations can potentially overlap, like:

- Full backup creates its staged snapshot
- It releases the domain lock and starts uploading
- A differential backup acquires the domain lock and accesses snapshot state
- The full backup later publishes or discards that state

backups-lib.pl
iliaross: No. It only returns a pathname.

backups-lib.pl
iliaross: I would rather simplify this if necessary, but not remove this crash protection completely. That prevents an interrupted run from leaving old state for a domain whose archive may already have been replaced, without invalidating every domain at the start, right?

backups-lib.pl
iliaross: Yet, these https://github.com/virtualmin/virtualmin-gpl/pull/1246/commits/a3a206f8c978464bb8d9e4fb6ba68417b9841f80 locking will be better to avoid `lock_file` timeouts.

backups-lib.pl
iliaross: Agreed. These checks are truly unnecessary. I've removed them https://github.com/virtualmin/virtualmin-gpl/pull/1246/commits/4dc3950d210b00492a8d9cd9a924922c282a02de.

jcameron:
So I had a look again at the current logic that is supposed to handle the case of a failed full backup, and realized that it doesn't treat failed uploads properly in the case of a full backup, as you correctly pointed out. But the fix to me seems as simple as : https://github.com/virtualmin/virtualmin-gpl/commit/ef6b1483fd880342c75d33612d844aaf209c0b7a

If this doesn't work, I would be really interested to see a sequence of Virtualmin CLI commands that demonstrate the problem. I was able to trigger it via a SSH full backup for a single to a working destination, then another SSH backup to a different destination with permissions that prevented an upload.

iliaross:
I think you're right. My PR though also handled crashes and partial multi-destination success, but that extra complexity likely isn't justified. I will rather close this for now.

And, for your latest change you could use `!$ok` instead of tracking new `$remotefailed`, right?

jcameron:
> I think you're right. My PR though also handled crashes and partial multi-destination success, but that extra complexity likely isn't justified. I will rather close this for now.

Thanks! But does my fix really address all the issues you found originally though? I am surprised that it's so much simpler.

> And, for your latest change you could use !$ok instead of tracking new $remotefailed, right?

I considered that, but since $ok is used as an indicator that the archiving process failed in which case @errdoms should already be populated, I felt it was more readable to use $remotefailed.

iliaross:
> But does my fix really address all the issues you found originally though?

No, not really. Yet, I thought you wanted to keep it more simple... and this is why I closed it.

> I considered that, but since $ok is used as an indicator that the archiving process failed in which case @errdoms should already be populated, I felt it was more readable to use $remotefailed.

Got it, thanks!

PR #1249, opened 2026-07-05
Fix to use FindBin for script path resolution

iliaross:
Hello Jamie!

As we just discussed on the other PR, this PR replaces existing hand-rolled `$0` path parsing in Virtualmin script bootstraps with Perl’s core `FindBin::RealBin`.

This keeps the current bootstrap behavior and existing `$0` assignments, while avoiding custom path-detection logic.


jcameron:
Thanks for this fix!!

PR #1252, opened 2026-07-19
Fix to make Webmin module access a per-domain server owner limit

iliaross:
Hey Jamie,

This PR moves domain owner Webmin module access from its standalone server-template section to:

- **Manage Virtual Server → Edit Owner Limits → Allowed Webmin modules**
- **Server Templates → Administration user → Webmin modules allowed by default**

Template settings now only initialize newly created top-level virtual servers. After creation, each server retains its own Webmin module access policy, so later template changes do not affect existing server owners.

## Changes

- Make Webmin module access a per-domain server owner limit. Templates now provide defaults copied only when a top-level virtual server is created, and later template changes do not affect existing servers.
- Preserve existing effective access through a one-shot migration, with consistent handling across creation, imports, restores, and reparenting.
- Separate domain owner defaults from the legacy Pro reseller `avail_*` settings, ensuring reseller permissions are neither widened nor reset during migration.
- Modernize the UI, remove the obsolete template section, and add documented `modify-limits` support through `--webmin-module` and `--no-webmin-module`.


edit_limits.cgi
```diff
@@ -169,6 +168,30 @@ if (defined(&list_scripts)) {
 
 print &ui_hidden_table_end("limits");
 
+# Webmin module access section
+print &ui_hidden_table_start($text{'limits_header5'}, "width=100%", 2,
+				     "wmods", 0, [ "width=30%" ]);
+
+if (&can_webmin_modules()) {
```
jcameron, line 175: Which kind of user would be able to edit limits, but not manage webmin modules for a domain owner?

modify-limits.pl
```diff
@@ -93,6 +93,119 @@ =head1 modify-limits.pl
 To hide the Webmin tab in the UI from the domain owner, use the
 C<-no-cat-webmin> flag. Or to make it visible again, use C<--cat-webmin>.
 
+To set access to one of the restricted Webmin modules normally granted by
+Virtualmin, use C<--webmin-module> followed by a module code and access level,
```
jcameron: Maybe rename this to --add-webmin-module and --remove-webmin-module, to be consistent with other API commands?

postinstall.pl
```diff
@@ -118,6 +118,35 @@ sub module_install
...
+# existing server owners as per-domain settings. This migration is idempotent
+# so an interrupted post-install can safely be run again.
+if (!$config{'migrated_domain_webmin_avail'}) {
+	# A zero for a plugin module was historically ignored at runtime and was
+	# also written automatically when saving an unrelated default-template
+	# section. Preserve the old effective default in the new owner-policy key,
+	# without changing the legacy avail_* keys still used by Pro resellers.
+	&migrate_default_webmin_avail();
+	undef(@list_templates_cache);
+	my @templates = &list_templates();
```
jcameron: Can you explain further which the templates need to be modified here?

postinstall.pl
```diff
@@ -118,6 +118,35 @@ sub module_install
...
+# Preserve the effective template-based Webmin module permissions of all
+# existing server owners as per-domain settings. This migration is idempotent
+# so an interrupted post-install can safely be run again.
+if (!$config{'migrated_domain_webmin_avail'}) {
+	# A zero for a plugin module was historically ignored at runtime and was
+	# also written automatically when saving an unrelated default-template
+	# section. Preserve the old effective default in the new owner-policy key,
+	# without changing the legacy avail_* keys still used by Pro resellers.
+	&migrate_default_webmin_avail();
+	undef(@list_templates_cache);
```
jcameron: Why does the cache need to be cleared?

postinstall.pl
```diff
@@ -118,6 +118,35 @@ sub module_install
...
+			    @templates) {
+		my $avail = &legacy_webmin_avail($tmpl->{'avail'});
+		if ($avail ne $tmpl->{'avail'}) {
+			$tmpl->{'avail'} = $avail;
+			&save_template($tmpl);
+			}
+		}
+	foreach my $d (grep { !$_->{'parent'} } &list_domains()) {
+		if (&init_domain_webmin_avail($d, 1)) {
+			&save_domain($d);
```
jcameron, line 127: Is the domain locked?

virtual-server-lib-funcs.pl
```diff
@@ -582,6 +582,14 @@ sub unlock_domain
 sub save_domain
 {
 my ($d, $creating) = @_;
+
+# Every newly persisted top-level server must own a stable copy of its Webmin
+# module policy, including imports and legacy restore paths that bypass the
+# normal virtual-server creation workflow.
+&init_domain_webmin_avail($d)
```
jcameron: If this only happens when creating a domain, can this code be moved to create_virtual_server ?

virtual-server-lib.pl
```diff
@@ -92,9 +92,9 @@
 @subdom_features = ( @opt_subdom_features );
 @database_features = ( 'mysql', 'postgres' );
 @template_features = ( 'basic', 'resources', @features, 'virt', 'virtualmin',
-		       'plugins', 'scripts', 'autoconfig', 'php', 'avail',
```
jcameron, line 95: Doesn't the 'avail' template page still need to exist to set default modules available?

edit_limits.cgi
iliaross: A Virtualmin Pro reseller can edit owner limits but cannot manage the domain owner’s Webmin modules.

modify-limits.pl
iliaross: Agreed. Fixed here https://github.com/virtualmin/virtualmin-gpl/pull/1252/commits/26559edb51ca938832cec8a6a367663f18d36a03.

postinstall.pl
iliaross: Only custom templates for top-level servers that have their own Webmin module settings are updated.

The default template is migrated separately, while inherited and sub-server templates need no changes. The update preserves the module access these templates provided before the upgrade.

postinstall.pl
iliaross: Earlier in `module_install`, the template list may already have been loaded and cached. So `migrate_default_webmin_avail()` then updates the default module settings in `%config`, but it does not update that cached template.

Clearing the cache makes the next `list_templates()` call reload the default template with the newly migrated settings. Otherwise, existing domains could be migrated using the old cached values...

postinstall.pl
iliaross: Ah, good catch, fixed now https://github.com/virtualmin/virtualmin-gpl/pull/1252/commits/4521d0b8e743bbcac433b022ca9cff756d557204.

virtual-server-lib-funcs.pl
iliaross: Not all domains are created through `create_virtual_server()`. Imports and some restores save them directly, so this check in `save_domain()` covers every creation path.

virtual-server-lib.pl
iliaross: No. The setting still exists, but it now appears in the template’s **Unix user** section as **Webmin modules allowed by default**, so the separate `avail` template page is no longer needed.

webmin: I don't think it makes sense to be in the Unix user section of the template then, as that's for settings related to the /etc/passwd and /etc/group entries

iliaross: I think this is the best place for it in templates. This section is only available to top-level domains and is logically related to "Administration user/domain user", which is a Unix user. If you have a better suggestion, let me know.

virtual-server-lib-funcs.pl
jcameron: I don't think this kind of complex logic belongs in save_domain though - it's a low-level API design to just write out the domain file. This should be extracted to a separate function that is called by create_virtual_server and any other special-case creation paths. Also, when a domain is created upon restore, it does call create_virtual_server

postinstall.pl
jcameron: Thanks!

postinstall.pl
jcameron: Doesn't that mean that migrate_default_webmin_avail is doing things the wrong way if it updates template-level settings without using an API like save_template ?

postinstall.pl
jcameron: Ok but what is this migration doing exactly?

modify-limits.pl
jcameron: Thanks!

edit_limits.cgi
jcameron: Oh right , thanks!

virtual-server-lib-funcs.pl
iliaross: Alright, agreed. Fixed now in this https://github.com/virtualmin/virtualmin-gpl/pull/1252/commits/0a39fbd3e4be8907c970e0c95067fbb21792faa4 patch.

postinstall.pl
iliaross: No. The default template is stored in Virtualmin’s main config, so `migrate_default_webmin_avail` updates the correct source and `save_module_config()` saves it.

Only custom templates are stored separately, so those use `save_template()`. The cache is cleared because it may still contain the old default value...

jcameron: Well yes, the default template is stored in the main config file internally. But the save_template function will still update it properly, and will preserve the cache.

iliaross: ...`save_template()` would work, but it would rewrite the entire default template when we only need to migrate one config value. The value is saved with the migration flag below, and the cache is cleared explicitly so it is rebuilt with the new value...

jcameron: Always better to use the correct API though. Imagine if in future we changed the way the default template was stored? If there is a bunch of code that assumes it's always in /etc/webmin/virtual-server/config , that would now be broken.

virtual-server-lib-funcs.pl

postinstall.pl
iliaross: Okay, I’ve updated the migration to load and save the default template through the template API https://github.com/virtualmin/virtualmin-gpl/pull/1252/commits/3d7d39de250c9cdaa02835ff7c67f56afb636271.

jcameron: Beautiful, thanks!

virtual-server-lib.pl
jcameron: Couldn't we just have it's own template section like it did before?

iliaross: We could, but I'd rather not. It fits quite well into the "Administration User" section.

jcameron: Ok, makes sense..

postinstall.pl
jcameron: Ping on this comment - I don't get why template migration is needed?

virtual-server-lib-funcs.pl
```diff
@@ -18157,123 +18176,350 @@ sub parse_template_plugins
...
-# Returns a list of modules that can be granted to domain owners, as array refs
-# with module name, description and list of options (optional) entries.
-sub list_domain_owner_modules
+# domain_owner_module_registry()
+# Returns all domain-owner module definitions and their backing Webmin modules.
+sub domain_owner_module_registry
 {
 &require_mysql();
 my $mytype = $mysql::mysql_version =~ /mariadb/i ? "MariaDB" : "MySQL";
+my %foreign_modules = (
```
jcameron: Couldn't this list be combined with the array below? Seems like we're repeating the list of modules twice..

virtual-server-lib-funcs.pl
```diff
@@ -18157,123 +18176,350 @@ sub parse_template_plugins
...
-		}
+return &normalize_webmin_avail($avail);
+}
+
+# get_default_webmin_avail()
+# Returns the default template's domain-owner module policy. Before this policy
+# had its own config key, the values were shared with Pro reseller access.
+sub get_default_webmin_avail
+{
+return $config{'default_webmin_avail'}
```
jcameron: Don't access %config here, instead fetch the default template and use it's fields.

virtual-server-lib-funcs.pl
```diff
@@ -18157,123 +18176,350 @@ sub parse_template_plugins
...
+return join(' ', @avail);
+}
+
+# migrate_default_webmin_avail()
+# Creates the separate default domain-owner policy while leaving the legacy
+# avail_* settings used by Pro resellers unchanged. Returns 1 if changed.
+sub migrate_default_webmin_avail
+{
+my $avail = &legacy_webmin_avail(&get_default_webmin_avail());
+return 0 if (defined($config{'default_webmin_avail'}) &&
```
jcameron: Same here with %config

postinstall.pl
iliaross: Module access previously came from templates. The migration copies the current access to existing domains and updates custom templates so their behavior does not change after the upgrade.

virtual-server-lib-funcs.pl
```diff
@@ -18157,123 +18176,350 @@ sub parse_template_plugins
...
+		next;
+		}
+	push(@available, $m) if (&foreign_available($fmod));
+	}
+return @available;
+}
+
+# get_domain_webmin_avail(&domain)
+# Returns the Webmin module access settings for a top-level domain owner.
+# Older domains fall back to the template value until postinstall has copied
```
jcameron: Instead of having a lot of code to deal with the short window where postinstall hasn't run yet, why not simplify and assume that it will always be executed on upgrade?

virtual-server-lib-funcs.pl
iliaross: I don't really think that, as combining them removes duplication, but fills every registry row with `undef` placeholders and mixes display data with installation-check details. The separate map is easier to read and clearly lists only modules requiring a Webmin installation check...

virtual-server-lib-funcs.pl
iliaross: That would cause recursion as `get_template(0)` calls `list_templates()`, and `list_templates()` calls `get_default_webmin_avail()` to populate the default template.

This function is the low-level, used to build the default template, so reading `%config` here is necessary.

virtual-server-lib-funcs.pl
iliaross: In fact, the new code can be loaded before postinstall finishes, and postinstall can also be interrupted. During that time, existing domains do not have the new setting yet.

This fallback keeps their current template-based access until migration completes. Afterward, it is no longer used...

...so, the fallback is small, and protects existing domains until postinstall completes... I would keep it!

postinstall.pl
jcameron: Ok so I get the migration to existing domains, but why do the templates themselves need to be changed?

virtual-server-lib-funcs.pl
```diff
@@ -18223,123 +18242,350 @@ sub parse_template_plugins
...
+        [ 'postgres', 'PostgreSQL Database Server' ],
+        [ 'spam', 'SpamAssassin Mail Filter' ],
+        [ 'filemin', 'File Manager', undef, 'config_avail_file' ],
         [ 'passwd', 'Change Password',
 	  [ [ 2, 'User and mailbox passwords' ],
 	    [ 1, 'User password' ],
 	    [ 0, 'No' ] ] ],
-        [ 'proc', 'Running Processes (user\'s processes only)',
+        [ 'proc', 'Running Processes',
 	  [ [ 2, 'See own processes' ],
```
jcameron, line 18277: Indent seems off here if this is supposed to be nested under the block for `proc`

virtual-server-lib-funcs.pl
jcameron: Well ok so what are the meanings of the `foreign_modules` and `rv` lists here? And in fact why is `foreign_modules` mostly a map from values to themselves??

virtual-server-lib-funcs.pl
jcameron: I'd rather not, as long as the failure isn't catastrophic in some way. 

This is an example of what I meant about complex code to solve a simple problem.

virtual-server-lib-funcs.pl
jcameron: That makes sense to read %config when building the default template. But in that case, why not just move this code into `list_templates` ?

virtual-server-lib-funcs.pl
jcameron: Actually I don't see why this function needs to exist at all if `list_templates` already calls `get_default_webmin_avail`

virtual-server-lib-funcs.pl
```diff
@@ -18223,123 +18242,350 @@ sub parse_template_plugins
...
+foreach my $m (&list_available_domain_owner_modules()) {
+	$avail{$m->[0]} = $values->{'avail_'.$m->[0]};
+	}
+my ($value, $bad) = &make_webmin_avail(\%avail);
+return $bad if ($bad);
+$tmpl->{'avail'} = $value;
+return undef;
+}
+
+# normalize_webmin_avail(string, [ignore-module-defaults])
```
jcameron: Can you explain what problem this function is fixing?

postinstall.pl
iliaross: The domain migration protects existing domains. The template migration protects future domains created from custom templates.

In the pre-upgrade code, `0` values for modules exposed through the plugin API were discarded, so those modules were still granted.

The new code honors `0`, so migration converts those legacy values to `1` to preserve existing behavior.

If no such custom template exists, nothing is changed.

virtual-server-lib-funcs.pl
iliaross: Umm... I guess not? It’s exactly like your old surrounding code.

virtual-server-lib-funcs.pl
iliaross: `@rv` is the list shown to the user, and `%foreign_modules` says which installed Webmin module each item depends on. Most names are the same, so they map to themselves. Some differ, such as `dns` using `bind8` and etc.

virtual-server-lib-funcs.pl
iliaross: Okay, I removed the fallback https://github.com/virtualmin/virtualmin-gpl/pull/1252/commits/df1482e593803f894d194ec4d405d13ae454ef19.

To be clear, until `postinstall` finishes, some owners may temporarily lose access to optional Webmin modules, can be especially noticeable on large or slow systems.

So, the simpler approach is reasonable, yet this potential temporary interruption is the tradeoff, so not totally a simple problem. 🙂

virtual-server-lib-funcs.pl
iliaross: We can move that logic into `list_templates`, okay https://github.com/virtualmin/virtualmin-gpl/pull/1252/commits/61bad253984cb96069b62cecca6b0b1c7c145a96.

virtual-server-lib-funcs.pl
iliaross: No, not really. `list_templates` only reads the old value, but migration saves it as the new, separate owner setting.

We could inline it into `postinstall.pl`, but the same logic would still be needed.

virtual-server-lib-funcs.pl
iliaross: `normalize_webmin_avail` updates module settings saved by older versions. It adds safe values for modules that were not present when the settings were saved, converts old enabled values to `1`, and disables invalid values.

jcameron: So this is just cleaning up how we store values? Or is the behavior being changed in some way that would effect how they are interpreted? What would break if this function didn't exist?

virtual-server-lib-funcs.pl
jcameron: Can't we make this a lot simpler by removing entries where modules map to themselves then, and just assume that if there is no entry?

virtual-server-lib-funcs.pl
jcameron: Can we make this simpler by updating resellers-lib.pl to just use the template instead of the avail_ config variables?

virtual-server-lib-funcs.pl
iliaross: No, not just a cleanup. It defines how older values are interpreted. Valid settings remain unchanged, missing values receive safe defaults, old truthy values become `1`, and invalid values become `0`. This is why the sub is called `normalize`.

So once again, without it, older templates or backups could be interpreted differently after upgrade and previously available modules might become unavailable.

virtual-server-lib-funcs.pl
jcameron: Eh, maybe this is just due to how github shows indentation. Never mind..

postinstall.pl
jcameron: Why can't the new code just treat values the same way the old code did, to keep things simpler?

virtual-server-lib-funcs.pl
iliaross: Ahhh, this is a very good suggestion! Thanks! Fixed here https://github.com/virtualmin/virtualmin-gpl/pull/1252/commits/6a6dc6f367bfdfa540666ac2cf938fcef3dbee13.

virtual-server-lib-funcs.pl
iliaross: Oh, no no. A template controls Webmin access for domain owners, and `avail_*` controls access for resellers. I don't think we should couple two separate permission sets.

postinstall.pl
iliaross: Jamie, no, previously, `0` was ignored for modules added through `feature_modules`, such as AWStats, so they remained always enabled!! It was simply an unnoticed bug in the old code.

Migration changes only those old values to `1` **to PRESERVE existing access**. Afterward, with my new code setting them to `0` **WILL** correctly disables the module.

jcameron: What do you mean by "0 was ignored" ? Can you point to the code that was doing this?

jcameron: Actually, if there was a bug previously, I think it would be a lot more comprehensible to create a PR to fix that specific bug first. The PR has "too much mustard on the hot dog" as some engineers like to say :-)

iliaross: LOL. I only spotted this bug while working on this PR, and by that point it felt too late to bother with. And honestly, it didn't seem worth fixing separately anyway since the new code will take care of it on its own.

My main preference was to keep the existing behavior intact so the update wouldn't break anything for current users or make them wonder why their AWStats reporting link suddenly disappeared when logged in as a domain owner...

...so, as I mentioned earlier, the migration will keep the old permissions as they are, while any new changes to the domain's available module permissions will work correctly going forward.

jcameron: Well I'm not even sure I understand the current bug! Can you explain what is breaking for the user currently?

Overall my problem is that this PR is very hard to review because there's too much going on.

virtual-server-lib-funcs.pl
jcameron: Great thanks!

postinstall.pl
iliaross: > Well I'm not even sure I understand the current bug! Can you explain what is breaking for the user currently?

Yes, very simple. Go to "System Settings ⇾ Server Templates: Edit Server Template / Administrator's Webmin modules" and change access from "No" to "Yes" (or vice versa) for a plugin, say for the "AWStats Reporting". You will see that despite where the toggle is set, access is always granted (the link is always shown on the menu) when logged in as domain owner.

> Overall my problem is that this PR is very hard to review because there's too much going on.

This is understandable and somewhat expected because the change itself is not simple and is tricky to keep working correctly in all case scenarios.

webmin: Oh I see the bug you're referring to now! But isn't fixing that as simple as https://github.com/virtualmin/virtualmin-gpl/commit/eefe433132b1d1a60d9ade2589326c3bdc354f71

iliaross: That does work. However, it will change the existing behavior when just re-saving the template. Are we okay with this?

Do you expect this PR to be simplified and ignore the initial migration path?

virtual-server-lib-funcs.pl
jcameron: Right but what I meant was, can't the reseller code just access the available modules by getting template 0 ?

postinstall.pl
jcameron: Yes I am happy if we change the behavior, even if this means that new plugins will not get granted to existing domains by default. And if this allows the PR to be simplified to remove the migration code, great!

virtual-server-lib-funcs.pl
iliaross: It could, but it would change reseller permissions. For example, the migration changes the AWStats value from `0` to `1` for domain owners. If resellers also read template 0, they would suddenly gain AWStats access.

I'd prefer keeping `avail_*` to avoid this regression...

postinstall.pl
iliaross: Alright!! I have removed all that code to reduce complexity! See this https://github.com/virtualmin/virtualmin-gpl/pull/1252/commits/6f9d6d2eba239568db034962b7239e22a3f56c13 patch please.

virtual-server-lib-funcs.pl
```diff
@@ -18223,123 +18246,260 @@ sub parse_template_plugins
...
-		-r &help_file($module_name, $h[1]) ? &hlink(@h) : $m->[1], $minp ]);
+	$known{$m->[0]} = 1;
+	my $value = $values->{$m->[0]};
+	return (undef, $m->[0])
+		if (!&valid_webmin_avail_value($m, $value));
+	push(@avail, $m->[0].'='.$value);
+	}
+
+# Retain syntactically valid settings for plugins or conditional modules that
+# are temporarily unavailable. If they return, they become known above and
```
jcameron: Why is this code needed? I don't see why we need to do validation of stored values against the registry when they can only be set by either the UI or API. Any validation should happen there, so we can reduce complexity by removing additional validation on internal functions.

virtual-server-lib-funcs.pl
```diff
@@ -18223,123 +18246,260 @@ sub parse_template_plugins
...
+	$avail{$m->[0]} = $values->{'avail_'.$m->[0]};
+	}
+my ($value, $bad) = &make_webmin_avail(\%avail);
+return $bad if ($bad);
+$tmpl->{'avail'} = $value;
+return undef;
+}
+
+# normalize_webmin_avail(string)
+# Converts incomplete settings into valid values. Invalid or missing values
```
jcameron: Where would these incomplete settings come from?

virtual-server-lib-funcs.pl
iliaross: The same, I originally added it for backward compatibility.

For instance, an old template might contain `custom=1` even though the Shell module was not installed. I thought we should preserve that value so, if Custom module was installed later, the owner would automatically receive access again.

That said, I'm fine removing it, if you think things would be cleaner without it https://github.com/virtualmin/virtualmin-gpl/pull/1252/commits/e039766135fcb2655e22c24a49cf710f7d86616a.

virtual-server-lib-funcs.pl
iliaross: They come from modules installed after the policy was saved. Existing policies have no entry for the new module, so normalization adds it explicitly as disabled `0`. I updated the comment to make this clear https://github.com/virtualmin/virtualmin-gpl/pull/1252/commits/8c25f41a954d9a85943d657ce06a7d70c1fc0865.

jcameron: Modules meaning plugins? But after my most recent change, aren't they treated as disabled already?

virtual-server-lib-funcs.pl
jcameron: thanks!

postinstall.pl
jcameron: Thanks!

virtual-server-lib-funcs.pl
jcameron: So ... why not just not change values in the migration?

virtual-server-lib-funcs.pl
jcameron: I suppose that would only happen if a domain was update in the middle of the install though, right?

virtual-server-lib-funcs.pl
jcameron: I still don't get the sequence of events that could cause an issue? Seems like we could just keep the current parsing of values, and then the only case we need to worry about is restoring older domains from backups.

virtual-server-lib-funcs.pl
iliaross: > Modules meaning plugins?

Yes, both are essentially Webmin modules.

> But after my most recent change, aren't they treated as disabled already?

Yes, they are, treated as disabled.

This normalization **only stores that state explicitly** as `module=0`. Technically it isn’t _necessary_ and could be removed, yet it has some benefits when manually reviewing and editing config files.

virtual-server-lib-funcs.pl
iliaross: Actually, that's what the migration does now. It copies each existing domain's current effective template module access settings into the domain's `webmin_avail` setting. It does not convert `0` to `1`, and neither templates nor reseller settings are modified.

virtual-server-lib-funcs.pl
iliaross: No, not really. Before the upgrade, existing domains have no `webmin_avail` setting because their access was coming directly from templates. The migration copies those current template values into each domain config.

So, without that copy, existing domains would lose module access with the new code from this PR, as domain config now should be the source of truth for available modules.

Older backups are handled separately when they are restored:

 - If the backup already contains `webmin_avail`, it is preserved.
 - If it is an older backup without that setting, `init_domain_webmin_avail` copies the current assigned template's module settings into the restored domain.
 
And, to answer your previous question about the use case I forgot about, `normalize_webmin_avail` is also useful if the assigned template cannot be found. So, it stores an explicit fail-closed policy such as `module1=0`, `module2=0` and so on, instead of having an empty `webmin_avail` value.

virtual-server-lib-funcs.pl
iliaross: Yeah, under a normal upgrade, nothing goes wrong. The only remaining risk is a concurrent Webmin ACL rebuild before migration finishes.

jcameron: That seems like a very very low likelihood event to me..

iliaross: Yes, that's why I was fine with dropping it, according to your recommendation.

virtual-server-lib-funcs.pl
jcameron: > No, not really. Before the upgrade, existing domains have no webmin_avail setting because their access was coming directly from templates. The migration copies those current template values into each domain config

Right I understand that. What I am asking is exactly in what situation would `normalize_webmin_avail` need to fix the config? 

> useful if the assigned template cannot be found.

That cannot happen though, it is impossible to delete a template that is used by domains, or restore a domain with a missing template. So there is no need to support this case.

virtual-server-lib-funcs.pl
jcameron: Oh well great, that's all I am asking for!

virtual-server-lib-funcs.pl
jcameron: Great!

virtual-server-lib-funcs.pl
jcameron: Ok let's remove it then. Perhaps some cleanup could be done in a later PR.

virtual-server-lib-funcs.pl
iliaross: Alright, removed https://github.com/virtualmin/virtualmin-gpl/pull/1252/commits/3a4295d728472f9bd6641ecdc47df839ebf7360a!

virtual-server-lib-funcs.pl
iliaross: Removed too https://github.com/virtualmin/virtualmin-gpl/pull/1252/commits/6ed1e3f1c12b6bc57405b06065d057968bad1425.

virtual-server-lib-funcs.pl
jcameron: I think this is the last open comment and then we're good!

iliaross: Then we're all set, since I believe my earlier comment already confirmed that the behavior is working the way you wanted, isn't it?

jcameron: oh ok cool thanks!

PR #1254, opened 2026-07-23
Fix IPv6-only domain lifecycle regressions

iliaross:
Hey Jamie!

This PR corrects IPv6-only creation, parent-address inheritance, DNS, mail, web/SSL cleanup, certificate linkage, and **restore behavior** _when_ IPv4 is absent.

### DIFF
--- a/create-domain.pl
+++ b/create-domain.pl
@@ -572,8 +572,8 @@ elsif ($virt) {
 	# Make sure manual IPv4 specification is allowed
 	$tmpl->{'ranges'} eq "none" || &usage("The --ip option cannot be used when automatic IP allocation is configured in templates - use --allocate-ip instead");
 	}
-elsif ($ip eq "default") {
-	# Use default IPv4, which may depend on reseller
+elsif ($ip eq "default" && !$aliasdomain && !$parentip) {
+	# Alias and parent IPv4 addresses are resolved below
 	$ip = $defip;
 	$ip || &usage("No default IP address found");
 	$virt = 0;
@@ -603,9 +603,11 @@ elsif ($virt6) {
 elsif ($ip6 eq "default") {
 	# Use default IPv6, which may depend on reseller
 	$ip6 = $defip6;
-	$ip6 || &usage("No default IPv6 address found");
+	if (!$ip6 && &indexof("--default-ip6", @OLDARGV) >= 0) {
+		&usage("No default IPv6 address found");
+		}
 	$virt6 = 0;
-	$name6 = 1;
+	$name6 = $ip6 ? 1 : 0;
 	}
 
 if (!defined($ip) && !defined($ip6)) {
@@ -820,6 +822,8 @@ if (!$alias) {
 		# IP comes from parent domain
 		$parent || &usage("The --parent-ip flag cannot be used for ".
 				  "top-level servers");
+		$ip = $parent->{'ip'};
+		$ip6 = $parent->{'ip6'} if (!$virt6);
 		}
 
 	if ($virt6) {
@@ -932,7 +936,7 @@ $pclash && &usage(&text('setup_eprefix3', $prefix, $pclash->{'dom'}));
 	 'dns_ip6', !$ip6 ? undef :
 		    defined($dns_ip6) ? $dns_ip6 :
 		    $alias ? $alias->{'dns_ip6'} :
-		    $virt ? undef : &get_dns_ip($resel, 6),
+		    $virt6 ? undef : &get_dns_ip($resel, 6),
          'virt', $virt,
          'virtalready', $virtalready,
 	 'ip6', $parentip ? $parent->{'ip6'} : $ip6,
@@ -1251,5 +1255,3 @@ print "                        [--alias-redirect | --no-alias-redirect]\n";
 print "                        [--ssl-redirect | --no-ssl-redirect]\n";
 exit(1);
 }
-
-
--- a/create-domain.pl
+++ b/create-domain.pl
@@ -603,9 +603,6 @@ elsif ($virt6) {
 elsif ($ip6 eq "default") {
 	# Use default IPv6, which may depend on reseller
 	$ip6 = $defip6;
-	if (!$ip6 && &indexof("--default-ip6", @OLDARGV) >= 0) {
-		&usage("No default IPv6 address found");
-		}
 	$virt6 = 0;
 	$name6 = $ip6 ? 1 : 0;
 	}
--- a/feature-dns.pl
+++ b/feature-dns.pl
@@ -1542,17 +1542,20 @@ if (!$tmpl->{'dns_replace'} || $d->{'dns_submode'}) {
 			# Make sure we have some NS records
 			@created_ns = &unique(@created_ns);
 			if ($tmpl->{'dns_indom'}) {
-				@created_ns = grep { &to_ipaddress($_) } @created_ns;
+				@created_ns = grep { &to_ipaddress($_) ||
+						     &to_ip6address($_) }
+						   @created_ns;
 				}
 			if (!@created_ns && !$d->{'dns_cloud'}) {
 				return $text{'setup_ednsns'};
 				}
 
 			if ($tmpl->{'dns_indom'}) {
-				# Add A records pointing to the nameserver IPs
+				# Add address records for the nameserver IPs
 				my $i = 1;
 				foreach my $ns (@created_ns) {
 					my $a = &to_ipaddress($ns);
+					my $aaaa = &to_ip6address($ns);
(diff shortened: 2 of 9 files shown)


create-domain.pl
```diff
@@ -603,9 +603,11 @@ package virtual_server;
 elsif ($ip6 eq "default") {
 	# Use default IPv6, which may depend on reseller
 	$ip6 = $defip6;
-	$ip6 || &usage("No default IPv6 address found");
+	if (!$ip6 && &indexof("--default-ip6", @OLDARGV) >= 0) {
```
jcameron: Best to not check for other flags like this, instead validation should be done after arg parsing by making sure that valid combinations of $ip and $ip6 are set.

feature-dns.pl
```diff
@@ -1542,17 +1542,20 @@ sub create_standard_records
 			# Make sure we have some NS records
 			@created_ns = &unique(@created_ns);
 			if ($tmpl->{'dns_indom'}) {
-				@created_ns = grep { &to_ipaddress($_) } @created_ns;
```
jcameron, line 1545: I actually considered doing this originally, but I don't think we can safely allow use of IPv6 NS records, because they would break the DNS domain for any system out there on the internet that doesn't support IPv6.

feature-dns.pl
```diff
@@ -2153,15 +2162,17 @@ sub save_domain_matchall_record
...
 if ($star && !$r) {
 	# Need to add
 	my $ip = $d->{'dns_ip'} || $d->{'ip'};
+	my $ip6 = $d->{'dns_ip6'} || $d->{'ip6'};
 	$r = { 'name' => $withstar,
-	       'type' => 'A',
+	       'type' => $ip ? 'A' : 'AAAA',
 	       'proxied' => $tmpl->{'dns_cloud_proxy'} == 1 ? 1 : 0,
-	       'values' => [ $ip ] };
+	       'values' => [ $ip || $ip6 ] };
```
jcameron: Since a record can either be only type A with $ip or type AAAA with $ip6, there should be separate branches. Or maybe we want to create *.domain.com for A and/or AAAA ?

feature-mail.pl
```diff
@@ -965,8 +965,8 @@ sub modify_mail
 # Update any outgoing IP mapping
 if ($supports_dependent) {
 	if ($d->{'dom'} ne $oldd->{'dom'} ||
-	    $d->{'ip'} && $oldd->{'ip'} && $d->{'ip'} ne $oldd->{'ip'} ||
-	    $d->{'ip6'} && $oldd->{'ip6'} && $d->{'ip6'} ne $oldd->{'ip6'}) {
+	    ($d->{'ip'} || "") ne ($oldd->{'ip'} || "") ||
```
jcameron, line 968: I don't think this is safe, we can't setup a dependent IP unless there is a v4 address

feature-mail.pl
```diff
@@ -6854,12 +6854,15 @@ sub enable_mta_sts
...
+	my $rtype = $ip ? 'A' : 'AAAA';
 	my ($mtsa) = grep { $_->{'name'} eq 'mta-sts.'.$d->{'dom'}.'.' &&
-			    $_->{'type'} eq 'A' } @$recs;
+			    $_->{'type'} eq $rtype } @$recs;
 	if (!$mtsa) {
 		$mtsa = { 'name' => 'mta-sts.'.$d->{'dom'}.'.',
-			  'type' => 'A',
-			  'values' => [ $d->{'dns_ip'} || $d->{'ip'} ] };
+			  'type' => $rtype,
+			  'values' => [ $ip || $ip6 ] };
```
jcameron: Since $rtype is known, the value has to be either $ip or $ip6

feature-ssl.pl
```diff
@@ -667,13 +667,15 @@ sub check_webmin_port_clash
 	&usermin::get_usermin_miniserv_config(\%uminiserv);
 	push(@checks, [ \%uminiserv, "Usermin" ]);
 	}
+my @ips = grep { $_ } ($d->{'ip'}, $d->{'ip6'});
```
jcameron: creating a tiny array of just at most 2 elements like this is hard to read. Instead just check if $s->[0] matches $d->{'ip'} or $d->{'ip6'} in the if statement below.

feature-ssl.pl
```diff
@@ -2869,16 +2874,18 @@ sub get_dovecot_ssl_cert_ip
 return ( ) if (!$l && !$l6);
 return ( ) if (!$imap && !$imap6);
 
-# We have at least one IP
-my $ione = $imap || $imap6;
+# Use the address for the protocol block whose certificate we read
+my ($imap_block, $cert_ip, $cert_ip6) = $imap ?
```
jcameron: I don't get this change? Seems like it does the same thing, but it a more difficult way?

create-domain.pl
iliaross: Agreed, fixed here https://github.com/virtualmin/virtualmin-gpl/pull/1254/commits/6a817c62a1c8aba060aa948c8aa5a4fa02f30b43.

jcameron:
This is mostly great, thanks! Added some comments though...

feature-dns.pl
iliaross: IPv6-only DNS is valid. IPv4-only systems won’t reach it, but that’s expected for an IPv6-only server. Requiring IPv4 here would defeat the purpose of this change...

feature-dns.pl
iliaross: Good point!! Fixed. `A` and `AAAA` wildcard records are now handled separately.

feature-mail.pl
iliaross: Fixed https://github.com/virtualmin/virtualmin-gpl/pull/1254/commits/c66567d7d4ebb0aa11b063931ab75dfedf0f23db. The mapping is only created when IPv4 is available.

feature-mail.pl
iliaross: Like this https://github.com/virtualmin/virtualmin-gpl/pull/1254/commits/9c6610cf111805a8d1e60ed6d0ebe4365e52062b? So, the record type and value are now selected together?

feature-ssl.pl
iliaross: Fixed https://github.com/virtualmin/virtualmin-gpl/pull/1254/commits/84b0ba338c74e6c98d93420e179d988d69f344cc. IPv4 and IPv6 are now checked directly.

feature-ssl.pl
iliaross: This makes the returned IP match the block we read!

Otherwise, when both blocks exist, the IPv4 certificate could be incorrectly reported as the IPv6 certificate too, right!!?

jcameron: But we only return one list containing the certificate and key, right? The old and new code still assume that the cert for a domain's IPv4 and IPv6 addresses are the same.

feature-mail.pl
jcameron: Ok thanks

feature-dns.pl
jcameron: Ok, makes sense..

feature-dns.pl
```diff
@@ -1542,17 +1542,20 @@ sub create_standard_records
 			# Make sure we have some NS records
 			@created_ns = &unique(@created_ns);
 			if ($tmpl->{'dns_indom'}) {
-				@created_ns = grep { &to_ipaddress($_) } @created_ns;
+				@created_ns = grep { &to_ipaddress($_) ||
+						     &to_ip6address($_) }
```
jcameron: Indentation seems off here?

feature-ssl.pl
iliaross: Ah, I see why I overcomplicated it now. Fixed in the latest https://github.com/virtualmin/virtualmin-gpl/commit/d95c013718acc7e537379afe41ce891067060cdd patch.

Though, should IPv4 and IPv6 certificates be handled separately in the future?

feature-dns.pl
iliaross: No, not really. Though I changed the formatting in https://github.com/virtualmin/virtualmin-gpl/pull/1254/commits/e9d569092acf6cae73360d4f7360908e23946681. Looks better now?

jcameron: thanks!

feature-ssl.pl
jcameron: Thanks for that patch! 

Virtualmin will never create a config with separate certs for v4 and v6, so I don't think we need to add support for parsing it.

feature-mail.pl
jcameron: Yeah sure, that works

PR #1258, opened 2026-07-30
Add ability for domain owners to use audited remote API commands

iliaross:
Howdy, Jamie!

This is an intentionally small first step to demonstrate and validate the approach for allowing non-master users to access audited commands through `remote.cgi`, as we discussed in the chat a few days ago.

Please take a look.

### DIFF
--- a/commands-lib.pl
+++ b/commands-lib.pl
@@ -454,10 +454,8 @@ return &$capfunc($d);
 sub require_remote_api_domain
 {
 my ($d, $requested, $program) = @_;
-# Default to the API script that called this function
-if (!$program) {
-	(undef, $program) = caller;
-	}
+# Default to the API script set by the dispatcher or standalone command
+$program ||= $0;
 return if (&remote_api_can_domain($d, $program));
 # Only distinguish capability failures for domains the caller can see
 my $msg = $d && &can_edit_domain($d) ?
@@ -472,10 +470,9 @@ exit(1);
 sub get_remote_api_domain
 {
 my ($field, $value, $requested) = @_;
-my (undef, $program) = caller;
 my $d = &get_domain_by($field, $value);
 &require_remote_api_domain(
-	$d, defined($requested) ? $requested : $value, $program);
+	$d, defined($requested) ? $requested : $value);
 return $d;
 }
 
--- a/commands-lib.pl
+++ b/commands-lib.pl
@@ -417,6 +417,68 @@ END {
 return 1;
 }
 
+# Core command-line API scripts audited for non-master remote use, and the
+# capability required for each one.
+my %user_remote_api_commands = (
+	"list-databases" => "can_edit_databases",
+	"list-proxies" => "can_edit_forward",
+	"list-redirects" => "can_edit_redirect",
+	"modify-user" => "can_edit_users",
+	);
+
+# get_user_remote_api_command(program-name)
+# Returns the capability function for an API command allowed for non-master users,
+# or undef if the command has not been allowed.
+sub get_user_remote_api_command
+{
+my ($program) = @_;
+$program = &normalize_api_command_name($program);
+$program =~ s/\.pl$//;
+return $user_remote_api_commands{$program};
+}
+
+# remote_api_can_domain(&domain, program-name)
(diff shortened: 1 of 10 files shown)


commands-lib.pl
```diff
@@ -417,6 +417,68 @@ END
...
+
+# require_remote_api_domain(&domain, requested-domain, [program-name])
+# Exits with a non-leaking error if the current user cannot run the calling
+# API command for the requested domain.
+sub require_remote_api_domain
+{
+my ($d, $requested, $program) = @_;
+# Default to the API script that called this function
+if (!$program) {
+	(undef, $program) = caller;
```
webmin: Isn't there an environment variable like SCRIPT_NAME you could use here instead of caller ?

jcameron:
This is great! Please add a functional test for the new API access though, perhaps in a follow-up PR.

commands-lib.pl
iliaross: Thanks! Good point about avoiding `caller`. Yet `SCRIPT_NAME` cannot be used because it identifies `remote.cgi`. I switched this to `$0` in this https://github.com/virtualmin/virtualmin-gpl/pull/1258/commits/ea97b65a5b562787d5fdfe14143361a7c31d617a patch.

iliaross:
> This is great! Please add a functional test for the new API access though, perhaps in a follow-up PR.

Wait, but we need to add many more commands using this same approach, right? Do you mind if I do using the same approach?

> Please add a functional test for the new API access though, perhaps in a follow-up PR.

I could do this yet we already have `remote-api-access.t` unit tests here. Is this enough?

commands-lib.pl
jcameron: Thanks!

jcameron:
> Wait, but we need to add many more commands using this same approach, right? Do you mind if I do using the same approach?

Sounds good! Although maybe we should wait till we have a new Virtualmin out first before making this major change?

> I could do this yet we already have remote-api-access.t unit tests here. Is this enough?

No, because that's a code test, not a true end-to-end test.  It would be just a variant of the `remote_tests` already in `functional-test.pl`

iliaross:
> Sounds good! Although maybe we should wait till we have a new Virtualmin out first before making this major change?

I don't know anymore if we should postpone it. I thought it would be more complex before I looked into it. Joe is right, people are expecting this to work sooner than later. And, if we think it's fine and approve the approach, then make a review on the final changes, perhaps it's not worth postponing. 

We are still working on the other PRs so it will take at least a couple more days.

jcameron:
If we do want to include this, we should consider a domain-level permission to control API access that is off by default.

iliaross:
> If we do want to include this, we should consider a domain-level permission to control API access that is off by default.

Can you please elaborate on this, I don't clearly understand what you mean.

jcameron:
> Can you please elaborate on this, I don't clearly understand what you mean.

I mean the `root` user should be able to control if domain owners can call the remote API.

iliaross:
> > Can you please elaborate on this, I don't clearly understand what you mean.
> 
> I mean the `root` user should be able to control if domain owners can call the remote API.

Ahh, I see. See this patch https://github.com/virtualmin/virtualmin-gpl/pull/1258/commits/80f00b2b30a599eabea94f505fef35a3374faad7.

iliaross:
And now, I think the shared permission and domain-boundary framework in this PR is ready to extend to cover all other domain-owner-applicable commands. I will look into wrapping this up in follow-up commits.

jcameron:
Thanks, merged! Please add a new functional-test that checks for both successful and failed API calls by domains owners in a followup PR though..

iliaross:
That was a bit too early to merge, as I was about to submit the second half of this PR to support all other subcommands. No problem, I will create a new follow-up PR for this.

jcameron:
Eh, it's fine to do the other commands in a separate PR

iliaross:
> Eh, it's fine to do the other commands in a separate PR

Yes, sure. Done finally in https://github.com/virtualmin/virtualmin-gpl/pull/1264. It took a while to test fully.

Issue #1259, opened 2026-08-01
btrfs quota support


derjohn:
I want to use btrfs for users /home with compression zstd . I think that saves lots of space , because e-mail boxes are uncompressed text (not for jpg or zip attachments obviously ...)

I read that btrfs uses a non-POSIX usr and grpquota, that's not supported by virtualmin, but I see that I can generally use btrfs: https://github.com/virtualmin/Virtualmin-Config/commit/483771c763999ab6004893a9e52a070ee648e8ed

So: That works but I wont have any quota support? If so, any plans to support btrfs quota ( qgroup etc ) ?

Best Regards,
derjohn

iliaross:
Thanks! I will take a closer look and update this ticket with progress.

derjohn:
Sidenote:  While it is not necessary for basic operations btrfs can also be mounted with "subvolumes", e.g. one for /home and another for /var/lib/mysql

derjohn:
Ah, wow! That was surprisingly fast !

If you will now modify the virtualmin part, too, is there any chance that we get a feature to sum up the used quota if we have e.g. /home and /var/lib/mysql on different mountpoints?

iliaross:
I'm working on this too. Btrfs quotas are complex. I'm trying to make sure the upcoming Virtualmin 8.2.0 already has support for it.

jcameron:
> I'm working on this too. Btrfs quotas are complex. I'm trying to make sure the upcoming Virtualmin 8.2.0 already has support for it.

Can we do this entitrely in the Webmin `quota` module, or do BTRFS quotas behave completely differently to EXT4 ?

iliaross:
> Can we do this entitrely in the Webmin `quota` module, or do BTRFS quotas behave completely differently to EXT4 ?

Yes, absolutely! In fact, I've been working all day on this and related bits in Virtualmin and Virtualmin Config. Submitting the PR for the Webmin Quotas module in a few minutes. Stay tuned please.

iliaross:
https://github.com/webmin/webmin/pull/2807

iliaross:
Now, it's been all fully implemented:
https://github.com/virtualmin/virtualmin-gpl/pull/1261
https://github.com/virtualmin/Virtualmin-Config/commit/fb069cbe49878ff3fa6f82e4f261410fba97a0ee

PR #1261, opened 2026-08-04
Add Btrfs subvolume and hierarchical quota support

iliaross:
Hi Jamie!

This PR adds native Btrfs quota support for virtual servers and mailbox users using subvolumes and hierarchical qgroups.

Domain and mailbox homes are created as Btrfs subvolumes. Level-0 qgroups enforce administrator and mailbox limits, while a level-1 qgroup provides the aggregate quota for the complete virtual-server tree, including sub-servers.

The implementation covers domain and mailbox creation, quota editing and reporting, cloning, backup restoration, deletion, accounting repair, and stale qgroup cleanup. Configuration checks reject incompatible simple quota mode and report existing homes that still require migration.

Requires Webmin 2.654 and its Btrfs quota API from webmin/webmin#2804.

Addresses https://github.com/virtualmin/virtualmin-gpl/issues/1259.


btrfs-lib.pl
```diff
@@ -0,0 +1,890 @@
...
+	next if (!$current->{'id'} || $seen{$current->{'id'}}++);
+	push(@rv, $current) if (!$current->{'alias'});
+	push(@pending, &get_domain_by("parent", $current->{'id'}));
+	}
+return @rv;
+}
+
+# btrfs_quota_api_error()
+# Returns an error if the required Webmin quota APIs are unavailable.
+sub btrfs_quota_api_error
```
jcameron: Why not just check the webmin version? In fact, aren't we already doing this in module.info ?

btrfs-lib.pl
```diff
@@ -0,0 +1,890 @@
...
+}
+
+# path_is_under_btrfs_mount(mount, path)
+# Returns 1 when path is the mount point itself or one of its descendants.
+sub path_is_under_btrfs_mount
+{
+my ($mount, $path) = @_;
+return 1 if ($mount eq "/");
+$mount =~ s/\/*$/\//;
+return $path eq substr($mount, 0, -1) || index($path, $mount) == 0;
```
jcameron: Better to use is_under_directory than substr

btrfs-lib.pl
```diff
@@ -0,0 +1,890 @@
...
+		$best_mount = $mount;
+		$best_root = $root;
+		}
+	}
+return defined($best_mount) ? ($best_mount, $best_root) : ( );
+}
+
+# btrfs_mountinfo(path)
+# Reads Linux mountinfo and returns the visible mount point and filesystem root.
+sub btrfs_mountinfo
```
jcameron: Shouldn't this API be in the quota module?

btrfs-lib.pl
```diff
@@ -0,0 +1,890 @@
...
+return wantarray ? ($text{'btrfs_enotopdomain'}, 0) :
+		   $text{'btrfs_enotopdomain'} if (!$d);
+return wantarray ? (undef, 0) : undef if ($user->{'home'} eq $d->{'home'});
+return &ensure_btrfs_subvolume($user->{'home'}, $d);
+}
+
+# convert_btrfs_restored_user_home(&user, &domain)
+# Converts a populated mailbox directory extracted by restore_dir into a
+# subvolume without exposing this disruptive behavior to normal user creation.
+sub convert_btrfs_restored_user_home
```
jcameron, line 336: Can you explain in detail what is going on here with subvolumes? Does each domain and mailbox need to have a subvolume?

btrfs-lib.pl
iliaross: It is a defensive check. It is also useful for tests, where the required quota functions are mocked without depending on a particular Webmin version.

btrfs-lib.pl
```diff
@@ -0,0 +1,890 @@
...
+sub delete_btrfs_subvolume
+{
+my ($path, $id) = @_;
+my $repair_path = $config{'home_quotas'};
+# When quota editing is disabled, use the still-existing parent directory.
+if (!$repair_path) {
+	$repair_path = $path;
+	$repair_path =~ s{/+\z}{};
+	$repair_path =~ s{/[^/]+\z}{};
+	$repair_path ||= "/";
```
jcameron, line 630: Is repairing the root directory safe?

virtual-server-lib-funcs.pl
```diff
@@ -1408,13 +1414,20 @@ sub list_all_groups_quotas
...
+	&error($err) if ($err);
+	}
 my $u;
 foreach $u (@groups) {
 	$u->{'module'} = $usermodule;
-	$u->{'softquota'} = $main::gsoft_home_quota{$u->{'group'}};
-	$u->{'hardquota'} = $main::ghard_home_quota{$u->{'group'}};
-	$u->{'uquota'} = $main::gused_home_quota{$u->{'group'}};
-	$u->{'ufquota'} = $main::gused_home_fquota{$u->{'group'}};
+	if (!$hasbtrfs) {
```
jcameron, line 1426: Why not put this `if` statement around the whole loop? That would be easier to read..

virtual-server-lib-funcs.pl
```diff
@@ -2590,12 +2603,18 @@ sub validate_user
 return undef;
 }
 
-# set_user_quotas(username, home-quota, mail-quota, [&domain])
+# set_user_quotas(username, home-quota, mail-quota, [&domain], [deleting])
```
jcameron: Can't we use a quota of 0 as a flag indicating that it is being deleted?

virtual-server-lib-funcs.pl
```diff
@@ -15378,6 +15408,9 @@ sub change_home_directory
 sub move_virtual_server
 {
 my ($d, $parent) = @_;
+# Moving a home across ownership trees requires atomically rebuilding its
```
jcameron, line 15434: What are ownership trees exactly?

jcameron:
Do we need any new functional tests for BRTFS? Or will the current quota tests cover it?

btrfs-lib.pl
iliaross: Agreed. Fixed in this patch https://github.com/virtualmin/virtualmin-gpl/pull/1261/commits/8bf6b239f750d457807db467b887806073da9691.

btrfs-lib.pl
iliaross: Yes, agreed and moved https://github.com/virtualmin/virtualmin-gpl/pull/1261/commits/e3c675108cb64146dbc53c10955f0135e1cb5c1e.

btrfs-lib.pl
iliaross: Btrfs `qgroup` limits apply to `subvolumes`, not ordinary directories or Unix users. This is why, every independently limited home must be a `subvolume`.

Each top-level domain home is a `subvolume` for the administrator’s quota, and each mailbox or FTP user with a separate home gets its own `subvolume` for an individual quota.

A higher-level `qgroup` combines them to enforce the domain’s total quota, including its sub-servers. Sub-servers and aliases do not need separate domaim-home `subvolumes`.

btrfs-lib.pl
iliaross: Yes, as it does not repair the directory itself, and the path only identifies the Btrfs filesystem. `/` is used when the deleted `subvolume` was directly under the filesystem root, with the rescan running only if quota record is inconsistent.

iliaross: ...to be clear, it can become incorrect after creating or deleting `subvolumes`, or changing `qgroup` membership, or interrupting a quota update with a reboot or crash, and  in that state, reported disk usage may be stale until rescanned.

virtual-server-lib-funcs.pl
iliaross: The loop also sets the group's module, which is still needed on Btrfs. Only the traditional quota fields should be skipped, so the current code is fine to me.

virtual-server-lib-funcs.pl
iliaross: No, not really, as quota `0` also means unlimited, so it does not necessarily mean the user is being deleted,  this is why the separate flag is needed here.

virtual-server-lib-funcs.pl
iliaross: Normal quotas use both user and group quotas. The group quota covers the top-level server and its sub-servers. The Btrfs `qgroup` hierarchy is designed to provide the same structure using `subvolumes` instead of Unix users and groups.

That said, an ownership "tree" is a top-level virtual server and all sub-servers underneath it, with Btrfs their home `subvolumes` sharing one `qgroup`.

iliaross:
> Do we need any new functional tests for BRTFS? Or will the current quota tests cover it?

Yes, added here https://github.com/virtualmin/virtualmin-gpl/pull/1261/commits/d80dc9b0ec91e596ac17fe336c4d980d438c3f28.

iliaross:
Jamie, does it all look good to you now?

btrfs-lib.pl
webmin: Are tests not run against or the in the context of a particular Webmin version though?

I get that this is defensive, but it's also confusing because someone reading this code is going to wonder "why would some of these functions exist but not others?"

btrfs-lib.pl
webmin: Thanks!

btrfs-lib.pl
jcameron: Nice thanks

btrfs-lib.pl
jcameron: Ok so doesn't that mean that any files the domain or mailbox user has outside their home directory won't be covered by BTRFS quotas? Like MySQL databases, or files in /tmp ?

btrfs-lib.pl
jcameron: So is this like the quotacheck that is sometimes necessary on EXT filesystems?

virtual-server-lib-funcs.pl
jcameron: In that case, the loop that sets the module should be entirely outside the if statement

virtual-server-lib-funcs.pl
jcameron: Ah do does that mean BTRFS has separate concepts of "no quota", "unlimited quota" and "finite quota" ?

virtual-server-lib-funcs.pl
jcameron: Ok got it , thanks!

btrfs-lib.pl
iliaross: Alright, I agree. I have removed https://github.com/virtualmin/virtualmin-gpl/pull/1261/commits/e5d4bb6af59e531ddd6cc719cb2d4c4e6bfbdb6d that unnecessary check.

btrfs-lib.pl
iliaross: Yes. Btrfs quotas track `subvolumes`, not file ownership.

We count the domain's home `subvolume` and its separate mailbox-user `subvolumes` by connecting them to one domain quota group.

Files outside those `subvolumes`, like MySQL data or files in `/tmp`, will not be counted.

So, another location could only be included by making it a dedicated `subvolume` and connecting it to the same quota group, which is quite complicated to implement for databases, and isn't practical for MySQL really, and is an obvious limitation of Btrfs `qgroups` for use in shared envs.

btrfs-lib.pl
iliaross: Yeah, it serves a similar purpose. For Btrfs, it is called a `quota rescan` and runs only when Btrfs reports inconsistent accounting.

virtual-server-lib-funcs.pl
iliaross: No, because that would require iterating over `@groups` twice! As you can see in the full code below, `$u->{'module'} = $usermodule;` is set unconditionally.

virtual-server-lib-funcs.pl
iliaross: No. Here, "no quota" and "unlimited quota" mean the same thing and are represented by `0`. Any positive value is a finite quota. The separate `deleting` flag only tells Virtualmin why `0` was passed, so it can clear the limit without creating or converting the user's home.

iliaross:
> Jamie, does it all look good to you now?

Jamie, do you have any other questions about this one?

btrfs-lib.pl
jcameron: Ok, so does this PR update `get_domain_user_quotas` to handle the case where MySQL databases are on the same filesystem as the domains' homes, but not under the home directory?

btrfs-lib.pl
jcameron: Thanks!

virtual-server-lib-funcs.pl
jcameron: Ok makes sense

btrfs-lib.pl
iliaross: Not `get_domain_user_quotas` directly. It's unchanged. The "already counted in home" decision is made in `mysql_size` via `has_mysql_quotas`, and this PR makes `has_mysql_quotas` return false when Btrfs quotas are active.

With `qgroups`, "same filesystem" no longer implies "counted in the domain's quota". Only files inside the domain's `subvolume` tree count, so MySQL data outside the home is treated like it lives on a separate filesystem, like reported separately and never subtracted from home usage.

And, Postgres already returned `0` for that portion.

All that said means MySQL usage is not enforced by the Btrfs quota at all in that scenario (only reported), which is the limitation already we discussed earlier in the thread.

iliaross:
Alright, I think we're totally good now.

btrfs-lib.pl
jcameron: Ok makes sense, thanks!

PR #1264, opened 2026-08-08
Add complete domain owners to use the remote API

iliaross:
Hey Jamie,

This PR completes the ability for domain owners to use the remote API what was previously started in https://github.com/virtualmin/virtualmin-gpl/pull/1258.

So, about 90 commands are covered — users, aliases, databases, DNS, SSL, PHP, scripts, backups and domain management.

Owners can create sub-servers but not top-level ones, restores are limited to dir/mysql/postgres like in the UI, and backups can only go to their own backup directory.

Anything above the domain level (plans, templates, resellers, limits) stays root-only.

Also tested with t/remote-api-access.t and against a real install as a domain owner.

Similar changes submitted directly without PRs:

https://github.com/virtualmin/virtualmin-htpasswd/commit/e7993fa1c0887e1694dc7f48d5733a3a488be930
https://github.com/virtualmin/virtualmin-pro/commit/c2a38845d1b39915d1fdf1b7cb2f3df7ab3a8e7e

### DIFF
--- a/list-certs-expiry.pl
+++ b/list-certs-expiry.pl
@@ -123,10 +123,10 @@ if (@domains) {
 		my $domain = $d->{'dom'};
 		my $cfile = $d->{'ssl_cert'};
 		if ($cfile && -r $cfile) {
-			my $expiration_date = &backquote_command(
-				"openssl x509 -enddate -noout -in ".quotemeta($cfile));
-			($expiration_date) = $expiration_date =~ /notAfter=(.*)\sGMT/;
+			my $info = &cert_file_info($cfile);
+			my $expiration_date = $info && $info->{'notafter'};
 			next if (!$expiration_date);
+			$expiration_date =~ s/\sGMT$//;
 			my $valid_until    = Time::Piece->strptime($expiration_date, $fpm_in);
 			my $valid_until_st = $valid_until->strftime("%s");
 			my $now_st         = $now->strftime("%s");
--- a/list-certs-expiry.pl
+++ b/list-certs-expiry.pl
@@ -31,9 +31,9 @@ if (!$module_name) {
 	else {
 		chop($pwd = `pwd`);
 		}
-	$0 = "$pwd/list-certs.pl";
+	$0 = "$pwd/list-certs-expiry.pl";
 	require './virtual-server-lib.pl';
-	$< == 0 || die "list-certs.pl must be run as root";
+	$< == 0 || die "list-certs-expiry.pl must be run as root";
 	}
 
 # Parse command-line args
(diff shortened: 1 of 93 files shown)


iliaross:
Jamie, alright, I think it’s all ready and can be merged.

Please check!

list-certs-expiry.pl
```diff
@@ -144,14 +163,13 @@ package virtual_server;
...
 			}
-		$i++;
 		}
 
 	# Sort results
 	my @rows = $sort_type eq 'desc' ? reverse sort keys %rows : sort keys %rows;
 	if ($multiline) {
 		foreach my $column (@rows) {
-			if ($all_doms || $rows{$column}[0] =~ /$domain/) {
+			if ($all_doms || $rows{$column}[0] =~ $domain_re) {
```
jcameron, line 172: Are you sure the `/ /` aren't needed here?

jcameron:
Mostly looks good, added one comment..

list-certs-expiry.pl
iliaross: Yes, no slashes are needed here, as `$domain_re` is already a compiled regex.

jcameron: Let's stick with `/$domain/` to be consistent with the rest of Virtualmin

list-certs-expiry.pl
```diff
@@ -93,21 +109,24 @@ package virtual_server;
...
-		if ($i % 2 == 1) {
-			my $domain          = $d;
-			my $cfile           = $data[$i];
-			my $expiration_date = `openssl x509 -enddate -noout -in "$cfile"`;
+	foreach my $d (@domains) {
+		my $domain = $d->{'dom'};
+		my $cfile = $d->{'ssl_cert'};
+		if ($cfile && -r $cfile) {
+			my $expiration_date = &backquote_command(
+				"openssl x509 -enddate -noout -in ".quotemeta($cfile));
```
jcameron: Can't the existing `cert_file_info` function be used to get the end date, rather than manually calling `openssl` ?

list-certs-expiry.pl
iliaross: Let's not, as `$domain_re` is compiled once so we can catch invalid regexes cleanly and reuse it.

`/$domain/` would work, but it would repeat the compilation and lose that validation.

And, Virtualmin already uses compiled regex variables elsewhere, so I'd keep it as-is.

list-certs-expiry.pl
iliaross: Ah, yes, this is an old pre-existing code before we reworked `cert_file_info` sub. Fixed https://github.com/virtualmin/virtualmin-gpl/pull/1264/commits/3ef5d1fb1d01653d7a4d0ca9ceab44310a9e3ba4 now.

jcameron: Thanks!

list-certs-expiry.pl
jcameron: How compatible is this with older versions of Perl?

iliaross: Like very, starting Perl 5.005 in 1998.

jcameron: Ok cool .. I guess it's more efficient since the regexp only needs to be compiled once.

PR #1270, opened 2026-08-23
Add option to include MySQL binary log coordinates in backups

iliaross:
Hi, Jamie!

This PR adds a Virtualmin Configuration option, under Backup and restore, to include MySQL/MariaDB binary log coordinates in database dumps, so a backup can later serve as the starting point for point-in-time recovery (PITR).

When enabled, the binary log file and position are written to each dump as a commented `CHANGE MASTER TO` / `CHANGE REPLICATION SOURCE TO` statement, so restoring the dump is unaffected. The option name is chosen by probing the configured dump client and using `--source-data=2` when it is supported, or `--master-data=2` otherwise, so MariaDB, older MySQL, and mixed client/server setups all keep working.

The global setting can be overridden per backup with the new `--enable-mysql-binlog-coordinates`, `--disable-mysql-binlog-coordinates` and, for schedules, `--default-config-mysql-binlog-coordinates` flags in `backup-domain`, `create-scheduled-backup` and `modify-scheduled-backup`.

It also fixes config-based scheduled backups reading and writing feature options under the wrong key, which caused those options to be silently dropped.

To be clear, this only records the coordinates. Binary logging must already be enabled on the server, and nothing is done about retaining or archiving binlogs.

Proper MySQL/MariaDB PITR support, including binary log management in the Webmin MySQL module and replaying binlogs on restore, is planned as a follow-up.


jcameron:
Any reason to not just always include these in the backup? Presumably it is harmless unless they are used at restore time?

Also, is there a user feature request or ticket this is in response to?

iliaross:
> Any reason to not just always include these in the backup? Presumably it is harmless unless they are used at restore time?

They are harmless when restoring, but generating them will fail if binary logging is disabled.

> Also, is there a user feature request or ticket this is in response to?

It was requested in the Virtualmin [forum discussion](https://forum.virtualmin.com/t/feature-request-allow-additional-flags-for-mysqldump-backups/137366/?u=ilia).

jcameron:
> They are harmless when restoring, but generating them will fail if binary logging is disabled.

Can we instead detect that, and include them always if binary logging is enabled?

iliaross:
> Can we instead detect that, and include them always if binary logging is enabled?

Gosh, of course! Sorry, I over-engineered this, because checking if binary logging is enabled is actually cheap. I fixed it in https://github.com/virtualmin/virtualmin-gpl/pull/1270/commits/dabb4d7ad292554cb0d04c1dbea8d7b4cddfe2e3 to do so automatically.

backups-lib.pl
```diff
@@ -36,7 +36,9 @@ sub list_scheduled_backups
 	my @bf;
 	foreach $f (&get_available_backup_features(), &list_backup_plugins()) {
 		push(@bf, $f) if ($config{'backup_feature_'.$f});
-		$backup{'opts_'.$f} = $config{'backup_opts_'.$f};
+		# Keep both keys because the restore form still uses opts_*
```
jcameron: Why not just change the restore form?

backups-lib.pl
```diff
@@ -138,7 +140,10 @@ sub save_scheduled_backup
 	my @bf = split(/\s+/, $backup->{'features'});
 	foreach $f (&get_available_backup_features(), &list_backup_plugins()) {
 		$config{'backup_feature_'.$f} = &indexof($f, @bf) >= 0 ? 1 : 0;
-		$config{'backup_opts_'.$f} = $backup->{'opts_'.$f};
+		# Accept either key when saving the legacy schedule
```
jcameron: Same here, seems like we should fix whever is causing the option to be inconsistent..

iliaross: Agreed. Fixed for both https://github.com/virtualmin/virtualmin-gpl/pull/1270/commits/557bf2babc32855936047f20f4627566f34cb83c.

jcameron:
Thanks for this feature!

iliaross:
> Thanks for this feature!

Jamie, my pleasure! And it wouldn't be complete without the binlogs replay feature. It was pretty hard to implement and required changes in both Webmin and Virtualmin, but I think I'm almost done testing it and will submit two separate PRs to the Webmin and Virtualmin GPL repos a bit later tonight.

Thanks!

PR #1281, opened 2026-09-06
Fix incorrect SSL cert paths left by older linkage-breaking code

iliaross:
Hi, Jamie!

This PR fixes incorrect SSL certificate paths saved by older releases when breaking certificate sharing.

Now, during package upgrades, it will recover the current incorrect certificate and missing CA chain from `ssl.combined`, correcting affected domain configs and preventing repeated bundle growth, even for domains that have renewed since the bug happened.


postinstall.pl
```diff
@@ -575,6 +575,42 @@ sub module_install
...
+# Repair cert paths saved by older releases when breaking SSL sharing. Finish
+# all repairs before syncing, as a shared domain can sync its owner.
+foreach (grep { $_->{'ssl_cert'} =~ /\/ssl\.combined$/ } &list_domains()) {
+	my $id = $_->{'id'};
+	&lock_domain($id);
+	my $d = &get_domain($id, undef, 1);
+	if ($d && $d->{'ssl_cert'} =~ /\/ssl\.combined$/ &&
+	    $d->{'ssl_cert'} eq $d->{'ssl_combined'}) {
+		my @certs = &unique(&cert_file_split($d->{'ssl_combined'}));
+		if (@certs && $certs[0] =~ /^-----BEGIN CERTIFICATE-----/) {
```
jcameron: Best to not assume that cert files start with "BEGIN CERTIFICATE", as some have a different format. And I think we have a function in Virtualmin already that can parse and validate a cert file, and handles this case.

iliaross: Thanks, good point. I completely forgot about that function. It’s fixed now: https://github.com/virtualmin/virtualmin-gpl/pull/1281/commits/8317df6812171d787db8641d3c767c875a84aa85

jcameron:
Thanks for the fix!

PR #1283, opened 2026-09-10
Fix ProFTPD SFTP and TLS log rotation

iliaross:
Hey Jamie,

This PR adds uncovered SFTP and TLS logs to the existing rotation block, preserving its settings and avoiding duplicate rules.

Verified on Debian, Ubuntu, AlmaLinux, and Rocky Linux VMs.

Fixes https://forum.virtualmin.com/t/possible-virtualmin-proftpd-sftp-log-rotation-oversight/137938.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 #### Version 8.2.1
 * Add Matomo web analytics installer to Virtualmin GPL, including detection of existing installs and display of the tracking code to embed in website pages
+* Fix missing ProFTPD SFTP and TLS log rotation while preserving existing rotation rules
 * Fix cloning Apache virtual servers overwriting newly issued SSL certificates with the source server's certificate [#1280](https://github.com/virtualmin/virtualmin-gpl/issues/1280)
 * Fix incorrect SSL certificate paths left by older releases when breaking certificate sharing, recovering the current certificate and any missing CA chain from the combined file
 * Fix batch-created virtual servers not applying the template's initial website content
--- a/postinstall.pl
+++ b/postinstall.pl
@@ -644,6 +644,9 @@ if ($config{'rs_endpoint'} eq 'https://lon.auth.api.rackspacecloud.com/v1.0') {
 # Unlock config now we're done with it
 &unlock_file($module_config_file);
 
+# Repair ProFTPd log rotation on upgrades after releasing the config lock.
+&setup_proftpd_logrotate();
+
 if (!defined($gconfig{'forgot_pass'})) {
 	# Enable Webmin forgotten password recovery
 	&lock_file("$config_directory/config");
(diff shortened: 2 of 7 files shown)


jcameron:
Is this `/var/log/proftpd/sftp.log` log file something Virtualmin sets up? It seems odd that we're updating the global logrotate config for it, since I would expect that to be the distro package's responsibility..

iliaross:
It is distro-specific. In particular, Rocky/Alma/RHEL have it, but Debian/Ubuntu do not. We are basically patching it for those distros that do not have it to help our users out.

jcameron:
Perhaps the call to `setup_proftpd_logrotate` would be better to have in `postinstall.pl` then, so it happens as soon as Virtualmin is upgraded?

iliaross:
> Perhaps the call to `setup_proftpd_logrotate` would be better to have in `postinstall.pl` then, so it happens as soon as Virtualmin is upgraded?

Yeah, and moreover, as I thought more about it, there is no reason to run it everytime in config check. In fact, it should only happen once. Fixed here https://github.com/virtualmin/virtualmin-gpl/pull/1283/commits/af94c5cc11d97bf6592cb13c691c3e6c4b26f475.

jcameron:
Excellent thanks!

One more question - is the directory `/var/log/proftpd` used on all Debian, Ubuntu, AlmaLinux, and Rocky Linux systems?

iliaross:
> is the directory `/var/log/proftpd` used on all Debian, Ubuntu, AlmaLinux, and Rocky Linux systems?

Yes, it's the same directory on all machines I ran tests on:

```
root@alma10-pro:~# ls -ld /var/log/proftpd
drwxr-x---. 2 root root 26 Aug 24 02:00 /var/log/proftpd

root@rocky9-pro:~# ls -ld /var/log/proftpd
drwxr-x---. 2 root root 112 Sep 10 02:57 /var/log/proftpd

root@debian12-pro:/etc# ls -ld /var/log/proftpd
drwxr-xr-x 2 root root 4096 Sep 11 00:00 /var/log/proftpd

root@debian13-pro:~# ls -ld /var/log/proftpd
drwxr-xr-x 2 root root 4096 Sep 10 15:36 /var/log/proftpd

root@ubuntu26-pro:~# ls -ld /var/log/proftpd
drwxr-xr-x 2 root root 4096 Sep 10 01:49 /var/log/proftpd
```

jcameron:
Wouldn't it make more sense to fetch these paths from the ProFTPd config rather than hard-coding them though?

iliaross:
> Wouldn't it make more sense to fetch these paths from the ProFTPd config rather than hard-coding them though?

Yeah, it would, but it's not that easy...alright, have a look at this https://github.com/virtualmin/virtualmin-gpl/pull/1283/commits/7056f476eb1a8777db7dc9b0e2866c8b618ad5fc patch. Like this looks good to you?

jcameron:
Nice thanks!

PR #1284, opened 2026-09-10
Fix DNS provider overrides during switch

iliaross:
Hi Jamie,

This PR prevent template and alias defaults from overriding the explicitly selected DNS provider during switch. This fixes moves to local DNS that could report success while leaving the domain configured for Cloudflare.

Preserve the original provider during rollback. Normal domain creation continues to use template defaults.

Fixes https://forum.virtualmin.com/t/the-cloud-dns-provider-with-the-cloudflare-identifier-is-either-not-supported-or-not-configured-on-this-system/137922

### DIFF
--- a/feature-dns.pl
+++ b/feature-dns.pl
@@ -47,11 +47,11 @@ if (!$d->{'dns'}) {
 return undef;
 }
 
-# setup_dns(&domain)
-# Set up a zone for a domain
+# setup_dns(&domain, [keep-provider])
+# Set up a zone for a domain, optionally keeping its selected DNS provider
 sub setup_dns
 {
-my ($d) = @_;
+my ($d, $keep_provider) = @_;
 &require_bind();
 my $tmpl = &get_template($d->{'template'});
 my $ip = $d->{'dns_ip'} || $d->{'ip'};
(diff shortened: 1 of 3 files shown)


feature-dns.pl
```diff
@@ -47,11 +47,11 @@ sub check_anti_depends_dns
...
 }
 
-# setup_dns(&domain)
-# Set up a zone for a domain
+# setup_dns(&domain, [keep-provider])
+# Set up a zone for a domain, optionally keeping its selected DNS provider
 sub setup_dns
 {
-my ($d) = @_;
+my ($d, $keep_provider) = @_;
```
jcameron: So is this needed because `set_provision_features` overrides domain-level fields like `provision_dns` ?

If so, it would be better to pass this in as a new field in `$d` rather than a param to `setup_dns`

iliaross: > If so, it would be better to pass this in as a new field in `$d` rather than a param to `setup_dns`

Agreed! Fixed https://github.com/virtualmin/virtualmin-gpl/pull/1284/changes/19a520488247e1cc29f4e64d160e313bf225ec21.

jcameron:
Thanks for this fix!!

PR #1285, opened 2026-09-11
Fix Postfix transport map backups

iliaross:
Hi Jamie,

This PR include transport-map source files in mail backups and restore them with rebuilt indexes.

Req.: https://github.com/webmin/webmin/commit/fc5794b04

Fixes: https://forum.virtualmin.com/t/reinstalling-virtualmin-restoring-backup-has-issues/137944/10


jcameron:
Code looks fine, but does Virtualmin modify the `transport_map` file? I think we only edit `sender_dependent_default_transport_maps` ..

iliaross:
> Code looks fine, but does Virtualmin modify the `transport_map` file? I think we only edit `sender_dependent_default_transport_maps` ..

Well yeah... however we already back up and restore `main.cf`, and manualy configured `transport_maps`. I think, it makes sense to include the files it references, too, so restoring doesn't leave Postfix pointing to missing maps. Also, our user stumbled against it, which is why I felt obliged to fix it.

iliaross:
Bump.

jcameron:
No I don't think this should go into Virtualmin, since we mainly do domain-level backups. And there are a vast number of different Postfix config files that could potentially be included.

Instead it would be better handled by the Webmin module Backup Config Files, which can already call he Postfix module to get all configs to include.

iliaross:
> Instead it would be better handled by the Webmin module Backup Config Files, which can already call he Postfix module to get all configs to include.

Alright, that means users can already handle Postfix module backups.

> No I don't think this should go into Virtualmin, since we mainly do domain-level backups. And there are a vast number of different Postfix config files that could potentially be included.

Alright, it's reasonable! I will close it then.

PR #1286, opened 2026-09-14
Fix WHOIS collection lock scope

iliaross:
Hey Jamie,

This PR fixes slow WHOIS calls holding domain locks.

WHOIS already sets `whois_next` 7–8 days ahead after every attempt, including failures.

The case I saw involved many domains without `whois_next`, so all their first checks ran during the same `collectinfo.pl` run.

And about your suggestion that this fix is too complex to address the problem... I don't think so.

Rare calls can still overlap an upgrade, and newly added domains can create a large first-run burst.

The **network call should run outside the lock**!


collect-lib.pl
```diff
@@ -218,30 +218,102 @@ sub collect_system_info
...
+		$current->{'whois_err'} = $error;
+		$current->{'whois_expiry'} = $expiry;
+		&save_domain($current);
+		}
+	}
+&unlock_domain($id);
+return 1;
+}
+
+# domain_whois_state(&domain)
```
jcameron, line 304: Can't we just use locking instead of this state tracking to prevent concurrent changes?

iliaross: No really. The domain lock must be released while WHOIS runs, or we recreate the original problem.

When we lock it again, we need to know whether another process changed the WHOIS data.

This state check prevents us from overwriting newer data without requiring another lock for every WHOIS path complexity...

jcameron: Do you mean like if multiple updates to the domain expire date happened at the same time? These are only done by collectinfo.pl right, which generally only executes once at a time. And even if somehow two updates happened close to each other, wouldn't they produce the same results anyway?

iliaross: No, `recollect_whois.cgi` can also refresh WHOIS data while `collectinfo.pl` is running.

We cannot keep the domain locked during the request because that caused the original problem.

Yet, I simplified the check in https://github.com/virtualmin/virtualmin-gpl/pull/1286/commits/4db92c077bdbec142eb23ecb32a2b88e2bf7c01a patch to compare only `whois_last` after locking again, and when it changed, another refresh finished first, so we discard the older result.

jcameron: Ok but even if that concurrent update does happen, wouldn't the results be the same? In this case it doesn't actually matter if an overwrite happens, the only difference would be the last check time, and even that would be minimal.

iliaross: As one WHOIS check can succeed while another fails. Without this check, the failed result could overwrite the good one, right?

jcameron: Can we just not update the expiry date in the failure case? Assuming that "failure" means a temporary failure, not a legitimate response that the domain doesn't exist.

iliaross: Only if we can tell a temporary failure from a domain that no longer exists. Currently both return an error with no expiry date. We're good _as is_!!

collect-lib.pl
```diff
@@ -218,30 +218,89 @@ sub collect_system_info
...
+my $domain = &get_domain($id, undef, 1);
+if (!$domain) {
+	&unlock_domain($id);
+	return 0;
+	}
+if ($domain->{'whois_next'} && $now < $domain->{'whois_next'}) {
+	&unlock_domain($id);
+	return 0;
+	}
+if ($domain->{'dom'} ne $name) {
```
jcameron, line 262: How could this ever happen? Didn't we get the name from the domain above?

collect-lib.pl
jcameron: Hmm .. I still think this code is just over-complication to deal with a rare edge case that we could just avoid by simply holding the domain lock for the duration of the whois lookup. But if you insist on keeping this complexity, fine.

I still think a better approach would be to add a timeout on the whois lookup of say 5 seconds, limit the number of whois refreshes done per collectinfo run, and randomize the order in which domains are refreshed.

collect-lib.pl
iliaross: In practice, it's only possible on an active multi-user system, as `$name` comes from the earlier domain list, the one before locking. So `$domain` is read again after locking, so the name can differ if the domain was renamed in between.

collect-lib.pl
iliaross: > But if you insist on keeping this complexity, fine.

I prefer to keep it because this complexity is justified.

> I still think a better approach would be to add a timeout on the whois lookup of say 5 seconds

It won't help much if I have 500 domains...

> limit the number of whois refreshes done per collectinfo run, and randomize the order in which domains are refreshed.

That's also complexity, though different type of.

collect-lib.pl
jcameron: Yeah ok I could see how that could happen

PR #1288, opened 2026-09-14
Fix background operations overwriting newer domain settings

iliaross:
Aloha Jamie,

This PR adds `save_domain_keys` and `save_domain_diff`, which lock the domain file, re-read it and merge only the requested keys, mirroring the module config helpers added in 8.2.0.

Use them in the login collector, scheduled disabling, dynamic and bulk IP updates, quota notifications and the certificate, password, limit and feature writers, so a domain disabled or edited while `collectinfo.pl` runs keeps its newer values.

Also make `save_domain` keep a lock the caller already holds instead of releasing it early, which the new helper and the scheduled-disable loop rely on, and recheck scheduled disabling under the domain lock so a schedule cancelled during collection is honoured.


jcameron:
This is the wrong approach - we should be making sure that locking is done before domains are read (or that they are re-read after taking a lock) rather than attempting to merge in updates from multiple writes. Because that can still causing corruption if two writes happen at the same time.

iliaross:
> We should be making sure that locking is done before domains are read (or that they are re-read after taking a lock) 

`save_domain_keys` already took the domain file lock, and forced re-read of the record. And, the nested lock change made sure that `save_domain` did not release that lock prematurely.

> This is the wrong approach

Anyway, I've reworked it, I think correctly this time. I will submit a new PR.

Thanks!

PR #1289, opened 2026-09-15
Fix domain updates to reread under lock

### LINKED ISSUE #1277: collectinfo.pl silently reverts domain metadata written while it runs (stale write in update_domains_last_login_times)
## Summary

`collectinfo.pl` reads every domain record at the start of its run and writes some of them back up to two minutes later, without re-reading them under the lock. Any change written to `/etc/webmin/virtual-server/domains/<id>` in that window is silently lost — including the `disabled` / `disabled_*` keys written by `disable_domain.cgi`.

The result is a virtual server that is really disabled (Apache, DNS, shell, MySQL all changed permanently) while its metadata says it is not, and the GUI reports *"Virtual server is not disabled"*.

`save_domain()` already detects this exact condition and prints `Domain <file> was modified since last read!` to stderr — and then overwrites the file anyway.

## Environment

- Virtualmin 8.1.0 GPL, Webmin 2.660, Authentic Theme
- RHEL 9.8, ~590 virtual servers
- `show_domains_lastlogin=1` (default)
- Reproduced against current `master`; the code below is unchanged there.

## Mechanism

`collectinfo.pl` runs from webmincron at `mins=3,8,13,18,23,28,33,38,43,48,53,58` and calls, among other things:

```perl
# collectinfo.pl
&update_domains_last_login_times()
      if ($config{'show_domains_lastlogin'});
```

`update_domains_last_login_times()` (`virtual-server-lib-funcs.pl`, ~line 22618 on master):

```perl
foreach my $d (&list_domains()) {              # whole list materialised at t0
      next if ($d->{'alias'});
      next if ($d->{'no_last_login'});
      my @logins;
      foreach my $user (&list_domain_users($d, 0, 1, 1, 1)) {   # slow: minute
              my $ll = &get_last_login_time($user->{'user'});
              ...
              }
      next if (!@logins);
      @logins = sort { $b <=> $a } @logins;
      &lock_domain($d);
      $d->{'last_login_timestamp'} = $logins[0];
      &save_domain($d);                      # writes the t0 snapshot
      &unlock_domain($d);
      }
```

Two things combine here:

1. `&list_domains()` materialises every domain object before the loop body runs. By the time domain *N* is saved, `$d` can be minutes old.
2. `&lock_domain($d)` is taken **after** the record was read, and nothing re-rlock correctly serialises writers, but it cannot help a writer that is holdingstale data — this is a plain lost update, not a torn write.

A naive fix of re-reading with `&get_domain($d->{'id'})` does not work either, because `get_domain()` returns `$main::get_domain_cache{$id}` and would hand back the same stale object. The
re-read has to pass the `$force` argument.

## Evidence that this is already detected

`save_domain()` stats the file and compares against `lastread_time`:

```perl
if (&read_file($file, $oldd)) {
      my @st = stat($file);
      if ($d->{'lastread_time'} && $st[9] > $d->{'lastread_time'}) {
              print STDERR "Domain $file was modified since last read!\n";
              &print_call_stack() if ($gconfig{'error_stack'});
              }
      }
...
&write_file($file, $d);     # overwrites regardless
```

So the condition is recognised, reported to a log nobody reads, and then ignor

```
# grep -c 'modified since last read' /var/webmin/miniserv.error
15329
```

They arrive in bursts of dozens of consecutive domains, which is the signature of the `collectinfo.pl` loop rather than of interactive use.

## Timing

Measured on this host by looking at `lastsave=` in the domain files whose last writer was webmincron:

- 587 domain files; 137 last written by `/webmincron/webmincron.pl`, of which 122 contain `last_login_timestamp` — i.e. this function is the writer.
- In the run starting at the `:43` cron minute, the 121 writes landed between

So each record is written back roughly 60–125 s after it was read, every 5 minhat lands in that window on an affected domain is discarded.

## Impact

The lost `disabled` flag is the small part. The damage compounds:

- **The GUI lies.** The domain is down, the metadata says it is up. There is nuse webmincron does not log file changes; the only fingerprint is`lastsave_script=/webmincron/webmincron.pl` in the domain file.
- **Clicking Disable a second time poisons the restore values.** The second rustate as the state to restore: `disabled_oldpass=*`, the lock hash in`disabled_oldmysql`, and no `disabled_shell` at all. A later Enable then does not unlock the account, and the original values have to be recovered from the `disable_domain.cgi` lines in
`/var/webmin/webmin.log`.
- **Duplicate `AliasMatch`.** Two Disables append `AliasMatch ^/.*$ /home/<user>/public_html/disabled_by_virtualmin.html` twice to every vhost. Enable removes one; the orphan keeps
serving the disabled page for a domain that is supposedly enabled. `virtualminentry once, so the duplicate is invisible from the GUI.
- **Half-disabled DNS zone.** The SOA gets the `.disabled.` suffix while the NS record keeps the old name, which surfaces as `out-of-zone data` / `has no NS records` / `bad zone` in the
dashboard validation.

Recovery means hand-editing the `disabled*` keys back into the domain file in , which is not something an admin should have to reason about.

## Steps to reproduce

Natural path, on a host with a few hundred domains and `show_domains_lastlogin from the GUI while the collectinfo webmincron job is in its write phase(roughly 1–2 minutes past the `:03`, `:08`, `:13`, … minute). Then look at `/etc/webmin/virtual-server/domains/<id>`: the `disabled` and `disabled_*` keys are gone, while the vhost, the
zone, the shell and the MySQL password stay changed.

Deterministic path:

1. `tail -f /var/webmin/miniserv.error`
2. Start `/usr/libexec/webmin/virtual-server/collectinfo.pl` by hand.
3. While it is in `update_domains_last_login_times()`, run `virtualmin disable (or click Disable in the GUI).
4. `Domain /etc/webmin/virtual-server/domains/<id> was modified since last read!` appears, and the `disabled` keys are gone from the file afterwards.

## Suggested fix

Re-read the record inside the lock, forcing past the cache:

```diff
      # Save most recent timestamp
      &lock_domain($d);
-     $d->{'last_login_timestamp'} = $logins[0];
-     &save_domain($d);
+     my $nd = &get_domain($d->{'id'}, undef, 1);
+     if ($nd) {
+             $nd->{'last_login_timestamp'} = $logins[0];
+             &save_domain($nd);
+             }
      &unlock_domain($d);
```

Beyond this one call site, two things seem worth considering:

- Any long-running loop that pairs an up-front `&list_domains()` with a late `exposure. Several other functions reached from `collectinfo.pl` iterate a listbuilt at the top of the run.
- Since `save_domain()` already knows the file changed under it, should it stireading and applying only the caller's changed keys, would turn a silent dataloss into something the caller can handle. Right now the detection exists but has no effect.

Happy to test a patch on this host — it reproduces reliably here.

iliaross:
This PR fixes other running processes overwriting newer virtual server settings.

It adds with_locked_domain to lock and reread the domain before making changes, keeping the lock through the final save.

It also makes save_domain preserve the caller's lock and saves backup metadata directly to the archive without overwriting live settings.

Fixes https://github.com/virtualmin/virtualmin-gpl/issues/1277.


jcameron:
Hang on, this seems like an awfully complex change for a simply lack of locking. I think we need to take a step back and first understand the exact trigger for the problem. 

The original poster said the trigger is : 

1. tail -f /var/webmin/miniserv.error
2. Start /usr/libexec/webmin/virtual-server/collectinfo.pl by hand.
3. While it is in update_domains_last_login_times(), run `virtualmin disable (or click Disable in the GUI).
4. Domain /etc/webmin/virtual-server/domains/<id> was modified since last read! appears, and the disabled keys are gone from the file afterwards.

The issue must be just somewhere in collectinfo.pl or update_domains_last_login_times

iliaross:
> I think we need to take a step back and first understand the exact trigger for the problem.

Fully agreed! 🙂 

And, I agree that rereading the domain after taking the lock would fix/work-around the reported login collector issue as well.

_However_, other code, including dynamic IP updaters, backups, quota notifications, manual WHOIS refreshes, and disable schedules, can and will sporadically also save an old copy over newer settings. Fixing only the collector leaves those other pretty plausible cases open!!

`with_locked_domain` gives callers a **shared way** to do this safely. To lock the domain, read its current settings, make changes, save, and unlock. The `save_domain` change only ensures it **does not release a lock the caller still needs**.

And, rereading inside `save_domain` would be too late, as the caller has already made its changes, and replacing its copy would lose them.

iliaross:
Jamie, please take a fresh deeper look into it.

jcameron:
Yeah what I am going to need to do is look through all the code that takes locks and makes sure it re-reads the domain afterwards, similar to what I did in https://github.com/virtualmin/virtualmin-gpl/commit/5e05724e5b8e10ea8d96292ab9b77ba40a6f8f2e . Probably the best way would be to add a new function named like `get_lock_domain` or something that gets a lock and re-reads the domain hash afterwards, and make sure that is called everywhere the matters.

PR #1290, opened 2026-09-18
Add comments to Apache directives in modify-web

iliaross:
Hi Jamie,

This PR adds comment support to modify-web `--add-directive` by passing each comment immediately before its directive using `--add-directive "# comment"`.

Comments are attached using Webmin's directive comment field. This depends on webmin/webmin#2847.


modify-web.pl
```diff
@@ -1050,6 +1068,18 @@ package virtual_server;
 				push(@old, $a->[1]);
 				&apache::save_directive(
```
jcameron: If there is a comment, can we not use save_directive_struct to add it with the comment? That would be cleaner than adding it, then finding and then adding a comment.

jcameron:
Nice thanks!!

modify-web.pl
iliaross: Could you explain it differently? I don't think I get it. Sorry.

jcameron: I meant why call save_directive here, and then save_directive_struct again below? Why not just call save_directive_struct once to add the value and it's comment?

iliaross: Got it now. See this change https://github.com/virtualmin/virtualmin-gpl/pull/1290/commits/b3939bac053d89b803503a2efbb117edc9eef0a7.

jcameron: Nice, but why not just always call save_directive_struct and not set the `comments` field if there is no comment?

iliaross: I see. What about this https://github.com/virtualmin/virtualmin-gpl/pull/1290/commits/22856803aab29c3ba82e62361c98f7b0a31472b8 patch?

jcameron: Cool that works

PR #1291, opened 2026-09-18
Add Apache comment preservation tests

iliaross:
Hey Jamie,

This PR adds functional tests for the new Apache comment preservation.

It uses modify-web `--add-directive` without changing Apache templates.

Related https://github.com/virtualmin/virtualmin-gpl/pull/1290 https://github.com/webmin/webmin/pull/2847


functional-test.pl
```diff
@@ -12855,6 +12855,26 @@ package virtual_server;
...
+	my (@comment_add_args, @comment_remove_args);
+	foreach my $directive (@comment_directives) {
+		push(@comment_add_args,
+			[ 'add-directive', $directive->[0] ],
+			[ 'add-directive', $directive->[1] ]);
+		push(@comment_remove_args,
+			[ 'remove-directive', $directive->[1] ]);
+		}
+	my $comment_search = 'grep -rhsF -- '.&quote_path($comment_marker).
+		' /etc/apache2 /etc/httpd 2>/dev/null | wc -l';
```
jcameron: What if the Apache config is in another directory? It's not save to hard-code paths like this in tests, and generally better to avoid using grep / awk at all.

A cleaner solution (even though it's more work) would be to create an API command like `list-apache-config` that takes a domain as parameter and dumps the `<Virtualhost>` . This could then be used in tests like this.

iliaross: Agreed about the hard-coded configs. I removed the `grep` and `awk` checks in https://github.com/virtualmin/virtualmin-gpl/pull/1291/commits/a9fca39db9dbba62cfefcad810f5d3366c2c23db patch.

The functional test now uses `modify-web`, `validate-domains`, and HTTP/HTTPS requests.

And I'm not adding `list-apache-config` because it isn't necessary for the functional tests, as Webmin's Perl `.t` tests already verify the actual comment preservation.

jcameron: Nice thanks!

PR #1292, opened 2026-09-20
Fix database user collisions in cPanel and Plesk migrations

iliaross:
Hello Jamie,

This PR fixes cPanel and Plesk migrations creating duplicate Unix accounts for imported database users.

It skips existing Unix usernames and correctly recognizes the shared parent owner during sub-server imports.

Plesk imports now merge database grants when the same user belongs to multiple databases.


jcameron:
Thanks for the fix!

